PR #11, opened 2022-03-10
Clarify that bundle types used for registers cannot contain analog, not just that a raw analog can't be used

darthscsi:


darthscsi:
Confirmed that this matches the current chisel behavior.

seldridge (approved):
LGTM!

(I pushed a commit to get this back to 80 columns instead of 72. 😉)

PR #13, opened 2022-03-11
Allow analog types in nodes

nandor:
Removed the restriction on assigning analog types to nodes, fixing a discrepancy between the definition of nodes and analog types.


darthscsi:
You might want to talk about this one with Adam.  The restriction is intentional to force only pass-though behavior and to disallow the type on any declaration.

PR #14, opened 2022-03-22
Correct issues with module description

dansvo:
This PR corrects two small issues I spotted while browsing this doc. See commit message bodies for more-detailed descriptions of the two changes.


seldridge (approved):
Thanks for taking the time to fix this!

Changes look good to me.

PR #15, opened 2022-04-07
Register reset values must be `equivalent`, not the same

youngar:
This changes the spec to allow implicit truncation and extension for
register reset values, which matches both the original FIRRTL compiler
and the CIRCT compiler's behaviour.


seldridge (approved):
Thanks for taking the time to update the spec!

PR #18, opened 2022-06-08
Adding `ExtCircuit`

debs-sifive:
Proposal for `ExtCircuit`s.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -251,8 +251,6 @@ A public module is accessible by `Circuit`s other than the `Circuit` it is
 defined in. Unless otherwise specified, all modules inside of a `Circuit` are
 private and cannot be accessed by any other `Circuit`.
 
-A public module has fixed ports.
-
 A module can be marked as public like so:
 
 ```firrtl
@@ -266,14 +264,23 @@ circuit FooCircuit :
 In the above example, the public module `Foo` gets instantiated in the
 `FooCircuit` the same way a private module would.
 
-A public module can be instantiated in an external Circuit like so:
+A public module must have a well-defined interface that can be captured in its
+declaration, so all ports should be fixed. The following example shows a
+declaration of a public module.
 
 ```firrtl
 extcircuit FooCircuit :
   extmodule Foo :
       input foo: UInt
       output bar: UInt
-      defname = VerilogName
+      ; ...
+```
+
+A public module can be instantiated in an external Circuit like so:
+
+```firrtl
+extcircuit FooCircuit :
+  extmodule Foo :
       ; ...
 
 circuit UsesFooCircuit :
--- a/spec.md
+++ b/spec.md
@@ -146,6 +146,7 @@ contributors is below:
 - [`@boqwxp`](https://github.com/boqwxp)
 - [`@chick`](https://github.com/chick)
 - [`@dansvo`](https://github.com/dansvo)
+- [`@debs-sifive`](https://github.com/debs-sifive)
 - [`@darthscsi`](https://github.com/darthscsi)
 - [`@donggyukim`](https://github.com/donggyukim)
 - [`@ekiwi`](https://github.com/ekiwi)
@@ -177,6 +178,22 @@ circuit MyTop :
       ; ...
 ```
 
+
+## Externally Defined Circuits
+
+An externally defined circuit is a circuit declaration whose implementation
+exists elsewhere. It should contain only module declarations
+([@sec:externally-defined-modules] or [@sec:public-modules]).
+
+```firrtl
+extcircuit MyExtCircuit :
+  extmodule FooExtModule :
+      ; ...
+  extmodule BarExtModule :
+      ; ...
+```
+
+
 ## Modules
 
 Each module has a given name, a list of ports, and a list of statements
@@ -199,6 +216,7 @@ Note that a module definition does *not* indicate that the module will be
 physically present in the final circuit. Refer to the description of the
 instance statement for details on how to instantiate a module
 ([@sec:instances]).
+
 
 ## Externally Defined Modules
 
@@ -229,6 +247,41 @@ ports.
 A common use of an externally defined module is to represent a Verilog module
 that will be written separately and provided together with FIRRTL-generated
 Verilog to downstream tools.
+
+
+## Public Modules
+
+A public module is accessible by `Circuit`s other than the `Circuit` it is
+defined in. Unless otherwise specified, all modules inside of a `Circuit` are
+private and cannot be accessed by any other `Circuit`.
+
+A public module has fixed ports.
+
+A module can be marked as public like so:
+
+```firrtl
+circuit FooCircuit :
+   module Foo is public :
+      input foo: UInt
+      output bar: UInt
+      ; ...
+```
+
+A public module can be instantiated like so:
+
+```firrtl
+extcircuit FooCircuit :
+  extmodule Foo :
+      input foo: UInt
+      output bar: UInt
+      defname = VerilogName
+      ; ...
+
+circuit UsesFooCircuit :
+   module FooUser :
+      inst foo of FooCircuit::Foo
+```
+
 
 # Types
 
--- a/spec.md
+++ b/spec.md
@@ -257,13 +257,16 @@ A module can be marked as public like so:
 
 ```firrtl
 circuit FooCircuit :
-   module Foo is public :
+   public module Foo :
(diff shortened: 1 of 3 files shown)


spec.md
```diff
@@ -230,6 +245,39 @@ A common use of an externally defined module is to represent a Verilog module
...
+
+```firrtl
+circuit FooCircuit :
+   module Foo is public :
+      input foo: UInt
+      output bar: UInt
+      ; ...
+```
+
+A public module can be instantiated like so:
```
debs-sifive, line 266: idk if I should move this down to the "instantiate" section

spec.md
```diff
@@ -230,6 +245,39 @@ A common use of an externally defined module is to represent a Verilog module
...
+defined in. Unless otherwise specified, all modules inside of a `Circuit` are
+private and cannot be accessed by any other `Circuit`.
+
+A public module has fixed ports.
+
+A module can be marked as public like so:
+
+```firrtl
+circuit FooCircuit :
+   module Foo is public :
```
mwachs5, line 260: Not sure `is public` is what we want here (that was just a sketch from me), probably just `public module` is more "what people are used to"?

spec.md
mwachs5, line 266:
```suggestion
A public module can be instantiated (in a different Circuit) like so:
```

Not sure if we want to mention that it will be instantiated in *this* Circuit as normal / in the same way it is instantiated if it were private?

spec.md
```diff
@@ -230,6 +245,39 @@ A common use of an externally defined module is to represent a Verilog module
...
+```
+
+A public module can be instantiated like so:
+
+```firrtl
+extcircuit FooCircuit :
+  extmodule Foo :
+      input foo: UInt
+      output bar: UInt
+      defname = VerilogName
```
mwachs5, line 273: again this is just a suggestion/placeholder for "other stuff we might eventually add that are not just ports", we should probably delete this `defname` thing for now.

debs-sifive: I added these in the style of other "placeholder" stuff in the `ExtModule` example... do you think it's better to leave the contents empty for now? Like just

```
extModule Foo :
  ; ...
```

spec.md
```diff
@@ -230,6 +245,39 @@ A common use of an externally defined module is to represent a Verilog module
 that will be written separately and provided together with FIRRTL-generated
 Verilog to downstream tools.
 
+## Public Modules
+
+A public module is accessible by `Circuit`s other than the `Circuit` it is
+defined in. Unless otherwise specified, all modules inside of a `Circuit` are
+private and cannot be accessed by any other `Circuit`.
+
+A public module has fixed ports.
```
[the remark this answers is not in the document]
mwachs5: I think we should make a higher level statement, that a public module has a well-defined interface that is capturable in an extcircuit.extmodule declaration (or something to that effect). Then we can explicitly enumerate what these are, currently that is any `port`.

spec.md
```diff
@@ -266,14 +264,23 @@ circuit FooCircuit :
 In the above example, the public module `Foo` gets instantiated in the
 `FooCircuit` the same way a private module would.
 
-A public module can be instantiated in an external Circuit like so:
+A public module must have a well-defined interface that can be captured in its
+declaration, so all ports should be fixed. The following example shows a
+declaration of a public module.
```
mwachs5, line 269: I'm not sure this sentence goes with the example below... this is basically how you are including a public module from Circuit into another circuitB

mwachs5, line 269:
```suggestion
A public module must have a well-defined interface that can be captured in its
declaration.
The well defined interface consists of:
 - All `ports`
 - The module's name
 
 The following example shows a declaration of a public module.
```

spec.md
```diff
@@ -177,6 +178,20 @@ circuit MyTop :
...
+exists elsewhere. It should contain only module declarations
+([@sec:externally-defined-modules] or [@sec:public-modules]).
+
+```firrtl
+extcircuit MyExtCircuit :
+  extmodule FooExtModule :
+      ; ...
+  extmodule BarExtModule :
+      ; ...
+```
```
seldridge, line 193: It would be great to also update the `include/firrtl.xml` grammar definition file so that syntax highlighting continues to work. This can be handled in a follow-up PR.

spec.md
seldridge: I think this should clearly define "well-defined interface":
1. the module name will not change
2. no ports will be added or deleted
3. aggregate type ports will be lowered to ground type ports using the algorithm in section ???

spec.md
```diff
@@ -230,6 +245,49 @@ A common use of an externally defined module is to represent a Verilog module
 that will be written separately and provided together with FIRRTL-generated
 Verilog to downstream tools.
 
+## Public Modules
+
+A public module is accessible by `Circuit`s other than the `Circuit` it is
+defined in. Unless otherwise specified, all modules inside of a `Circuit` are
+private and cannot be accessed by any other `Circuit`.
```
seldridge, line 252: A design point here is should the main module be default-public and should the main module be required to use the public keyword? I think it's reasonable to say that the main module is always public and does not use the public keyword. (This matches the current semantic for how the main module has some special handling around port removal right now, but this semantic could be changed.)

spec.md
```diff
@@ -230,6 +245,49 @@ A common use of an externally defined module is to represent a Verilog module
...
+
+A module can be marked as public like so:
+
+```firrtl
+circuit FooCircuit :
+   public module Foo :
+      input foo: UInt
+      output bar: UInt
+      ; ...
+```
```
seldridge, line 262: Bike-shedding ideas:

- `public module Foo:`
-  `module public Foo:`
- `module Foo public:`
-  `module[public] Foo:`
- `module Foo[public]:`
-  `module @public Foo:`
- `module Foo @public:`

spec.md
debs-sifive: I can try to do that in this PR! it seems straightforward

spec.md
debs-sifive: > aggregate type ports will be lowered to ground type ports using the algorithm in section ???

does such an algorithm exist yet, or does it need to be written?

spec.md
```diff
@@ -230,6 +245,49 @@ A common use of an externally defined module is to represent a Verilog module
...
+
+In the above example, the public module `Foo` gets instantiated in the
+`FooCircuit` the same way a private module would.
+
+A public module must have a well-defined interface that can be captured in its
+declaration, so all ports should be fixed. The following example shows a
```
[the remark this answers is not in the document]
darthscsi: "be fixed" seems sensible for external modules coming in not through firrtl, but why should firrtl modules coming in from other circuits be restricted to a subset of firrtl?

debs-sifive:
Pulled out `public` module stuff into a separate PR: https://github.com/chipsalliance/firrtl-spec/pull/20

spec.md
```diff
@@ -230,6 +245,49 @@ A common use of an externally defined module is to represent a Verilog module
 that will be written separately and provided together with FIRRTL-generated
 Verilog to downstream tools.
 
+## Public Modules
+
+A public module is accessible by `Circuit`s other than the `Circuit` it is
```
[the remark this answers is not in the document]
debs-sifive: > How does this mesh with the "circuit is a complete design".

Is a circuit a complete design? I don't think that's stated in the spec... Was that the intention in the past? It seems like that notion is no longer true.

> From later: "Modules have the property that instances can always be inlined into the parent module without affecting the semantics of the circuit." How is this valid for separately compiled modules?

What does "the semantics of a circuit" mean? If it means how the circuit is referenced I think that the statement still holds? If it means the behavior of the circuit, then the statement was never true.

> From later: "To disallow infinitely recursive hardware, modules cannot contain instances of itself, either directly, or indirectly through instances of other modules it instantiates." How is this valid for separately compiled modules?

I guess we can't guarantee anything about `extmodules` that are defined in Verilog, since MFC can't check those. I think for `public` modules, do you think it's possible to put in checks to prevent recursive `public` modules?

spec.md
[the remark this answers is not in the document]
debs-sifive: @seldridge's take on "fixed": https://github.com/chipsalliance/firrtl-spec/pull/18#discussion_r892771141

> I think this should clearly define "well-defined interface":
> - the module name will not change
> - no ports will be added or deleted
> - aggregate type ports will be lowered to ground type ports using the algorithm in section ???

PR #19, opened 2022-06-08
Makefile fix

debs-sifive:
Fixes issue of `make` not working if the `build` directory does not exist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,9 @@
 
 all: build/spec.pdf
 
-build/spec.pdf: spec.md include/spec-template.tex include/firrtl.xml include/ebnf.xml | build/
-	pandoc $< --template include/spec-template.tex --syntax-definition include/firrtl.xml --syntax-definition include/ebnf.xml -r markdown+table_captions+inline_code_attributes+gfm_auto_identifiers --filter pandoc-crossref -o $@
+build/spec.pdf: spec.md include/spec-template.tex include/firrtl.xml include/ebnf.xml
+	mkdir -p build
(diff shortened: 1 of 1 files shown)


seldridge (commented):
Can you clarify why this is needed or what error you are seeing?

This _shouldn't_ be needed because `build/` is specified as an `order-only-prerequisite` which is resolved by the `%/` rule at the bottom.

debs-sifive:
> by the %/ rule at the bottom.

I don't see this rule?

When I run `make` off master I see
```
make: *** No rule to make target `build', needed by `build/spec.pdf'.  Stop.
```

mwachs5:
> When I run make off master I see

Do you literally mean `master`, not `main`? Perhaps you are using some old version?

debs-sifive:
oh sorry, I do mean `main`. Are other people able to get `make` to work off of a clean `main`?

seldridge:
This works for me on on a clean clone of `main`. Did you accidentally change that line like the following:

```diff
diff --git a/Makefile b/Makefile
index d52634a..911bc9a 100644
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 
 all: build/spec.pdf
 
-build/spec.pdf: spec.md include/spec-template.tex include/firrtl.xml include/ebnf.xml | build/
+build/spec.pdf: spec.md include/spec-template.tex include/firrtl.xml include/ebnf.xml | build
 	pandoc $< --template include/spec-template.tex --syntax-definition include/firrtl.xml --syntax-definition include/ebnf.xml -r markdown+table_captions+inline_code_attributes+gfm_auto_identifiers --filter pandoc-crossref -o $@
 
 clean:
```

The trailing `/` is necessary to trip the directory building rule.

PR #23, opened 2022-06-09
Adding information about annotations

debs-sifive:
This is literally just a copy/paste of the first part of https://github.com/llvm/circt/blob/main/docs/FIRRTLAnnotations.md.

Is this the information we want to add to spec? Do we want to take some stuff out/change things?


spec.md
```diff
@@ -2369,6 +2369,85 @@ As an example `firetruck$y$z`{.firrtl} shares a prefix with
 `firetruck$y`{.firrtl} and `firetruck`{.firrtl}, but does not share a prefix
 with `fire`{.firrtl}.
 
+# Annotations
+
+Annotations encode arbitrary metadata and associate it with zero or more
+targets in a FIRRTL circuit.
```
mwachs5, line 2375: Targets are not yet defined, so I'm not sure if it makes sense to include this without the Targets info

spec.md
```diff
@@ -2369,6 +2369,85 @@ As an example `firetruck$y$z`{.firrtl} shares a prefix with
...
+
+Annotations are represented as a dictionary, with a "class" field which
+describes which annotation it is, and a "target" field which represents the IR
+object it is attached to. The annotation's class matches the name of a Java
+class in the Scala Chisel/FIRRTL code base. Annotations may have arbitrary
+additional fields attached. Some annotation classes extend other annotations,
+which effectively means that the subclass annotation implies to effect of the
+parent annotation.
+
+Annotations are serializable to JSON and either live in a separate file
```
mwachs5, line 2385: We should include the format from @jackkoenig / @seldridge for putting annotations within a .fir file

spec.md
mwachs5: oh you did define them. Should this be a Keyword, or should Targets come first?

spec.md
```diff
@@ -2369,6 +2369,85 @@ As an example `firetruck$y$z`{.firrtl} shares a prefix with
...
+circuit Foo:
+  module Foo:
+    inst a of Bar
+    inst b of Bar
+  module Bar:
+    inst c of Baz
+    inst d of Baz
+  module Baz:
+    skip
+```
```
mwachs5, line 2449: I think the examples are missing?

spec.md
```diff
@@ -2369,6 +2369,85 @@ As an example `firetruck$y$z`{.firrtl} shares a prefix with
...
+
```
mwachs5, line 2450: We should update the grammar at the end to encompass annotations once we have the in-.fir file representation

spec.md
debs-sifive: We were discussing on Slack... apparently this does not exist yet 😬

spec.md
```diff
@@ -2369,6 +2369,85 @@ As an example `firetruck$y$z`{.firrtl} shares a prefix with
...
+modules in a FIRRTL circuit.  A target consists of a circuit, a root module, an
+optional instance hierarchy, and an optional reference. A target can only
+identify hardware with a name, e.g., a circuit, module, instance, register,
+wire, or node. References may further refer to specific fields or subindices in
+aggregates. A target with no instance hierarchy is local. A target with an
+instance hierarchy is non-local.
+
+Targets use a shorthand syntax of the form:
+```
+target ::= “~” (circuit) (“|” (module) (“/” (instance) “:” (module) )* (“>” (ref) )?)?
```
mwachs5, line 2423: we should add this and `reference` to the grammar at the end

spec.md
seldridge: I think the plan was to land this change first, then add the syntax for the separate `%[<jsonArray>]` syntax. (That would be a trivial follow-up.)

Makefile
```diff
@@ -1,9 +1,12 @@
-
 all: build/spec.pdf
 
 build/spec.pdf: spec.md include/spec-template.tex include/firrtl.xml include/ebnf.xml | build/
 	pandoc $< --template include/spec-template.tex --syntax-definition include/firrtl.xml --syntax-definition include/ebnf.xml -r markdown+table_captions+inline_code_attributes+gfm_auto_identifiers --filter pandoc-crossref -o $@
 
+IMG_SRC=$(wildcard include/img_src/*.dot) 
+images: $(IMG_SRC) | build/ build/img/
+	$(foreach dotfile,$(IMG_SRC),dot -Tpng $(dotfile) -o build/img/$(patsubst %.dot,%.png,$(lastword $(subst /, ,$(dotfile))));)
+
```
debs-sifive, line 9: @seldridge this is the best I could do ;_;

I am not v good with makefiles, please lmk if there is a nicer way

spec.md
debs-sifive: Do Targets make sense in a non-Annotation context? Like do other things besides Annotations make use of Targets?

If not, I think it makes sense to keep targets as a subsection here. I can add a link here to the subsection.

spec.md
debs-sifive: @seldridge @jackkoenig is this syntax part of FIRRTL? I thought it was just some one-off representation for use in annotation JSONs.

If this representation is solely for targets/annotations, I feel like it shouldn't go in the FIRRTL syntax reference.

mwachs5: but the point of this PR is to get this into the FIRRTL syntax reference  and *make* it part of FIRRTL.

mwachs5: Sorry to be clear, I thought that this was how Targets are serializized in annotations, independently of the in-fir-file format we are saying is to be done in a follow-up.

Makefile
seldridge: You can change the `%/` rule to be `mkdir -p` and simplify the order-only prerequisite to just be `build/img/`.

A slightly more clean way to do this is to use a simpler rule that builds one PNG for one DOT file. I can never remember the wildcard syntax, but the actual files can be made explicit for now. Something like:

```make
IMG_PNG=include/img_src/foo.dot include/img_src/bar.dot
build/img/%.png: include/img_src/%.dot | build/img/
    dot -Tpng $< -o $@
```

You can then add a normal (not order-only) prerequisites to `build/spec.pdf` (and move `IMG_PNG` up to the top of the file:
```make
build/sped.pdf: spec.md include/spec-template.tex include/firrtl.xml include/ebnf.xml $(IMG_PNG)
```

spec.md
seldridge: I'm fine to just add this, but we can wait until we put the `%[<annotationArray>]` format in the spec. That will certainly motivate defining the exact syntax of a serialized annotation.

(What is annoying about this is that it is effectively enshrining the JSON serialization format with configuration options that we happen to be using for json4s. 😂)

spec.md
```diff
@@ -2369,6 +2369,108 @@ As an example `firetruck$y$z`{.firrtl} shares a prefix with
...
 
+# Annotations
+
+Annotations encode arbitrary metadata and associate it with zero or more
+targets in a FIRRTL circuit.
+
+Annotations are represented as a dictionary, with a "class" field which
+describes which annotation it is, and a "target" field which represents the IR
+object it is attached to. The annotation's class matches the name of a Java
+class in the Scala Chisel/FIRRTL code base. Annotations may have arbitrary
```
seldridge, line 2380: I'd remove language about Java and Scala Chisel/FIRRTL.

spec.md
```diff
@@ -2369,6 +2369,108 @@ As an example `firetruck$y$z`{.firrtl} shares a prefix with
...
+additional fields attached. Some annotation classes extend other annotations,
+which effectively means that the subclass annotation implies to effect of the
+parent annotation.
+
+Annotations are serializable to JSON and either live in a separate file
+(e.g., during the handoff between Chisel and the SFC) or are stored in-memory
+(e.g., during SFC-based compilation). The SFC pass API requires that passes
+describe which targets in the circuit they update. SFC infrastructure then
+automatically updates annotations so they are always synchronized with their
+corresponding FIRRTL IR.
```
seldridge, line 2390: Don't mention "SFC". Just use "a FIRRTL Compiler" instead. Actually, I'd just remove these sentences.

spec.md
```diff
@@ -2369,6 +2369,108 @@ As an example `firetruck$y$z`{.firrtl} shares a prefix with
...
+
+An example of an annotation is the `DontTouchAnnotation`, which can be used to
+indicate to the compiler that a wire "foo" should not be optimized away.
+
+```json
+{
+  "class":"firrtl.transforms.DontTouchAnnotation",
+  "target":"~MyCircuit|MyModule>foo"
+}
+```
```
seldridge, line 2400: Can you change this to not use `DontTouchAnnotation` and just use a generic annotation with a class called "Foo" or anything else. I'd prefer to not hint at optimization barrier semantics here.

spec.md
```diff
@@ -2369,6 +2369,108 @@ As an example `firetruck$y$z`{.firrtl} shares a prefix with
...
+
+Some annotations have more complex interactions with the IR. For example the
+[BoringUtils](https://www.chisel-lang.org/api/latest/chisel3/util/experimental/BoringUtils$.html)
+provides FIRRTL with annotations which can be used to wire together any two
+things across the module instance hierarchy.
```
seldridge, line 2405: This can be deleted.

seldridge (approved):
This looks great! I'm happy to finally be seeing this be documented in the spec.

My comments are mostly nits to remove stuff. Looks great otherwise. Thanks for fighting with Makefiles.

PR #25, opened 2022-06-13
Adding async reset info

debs-sifive:
Not really sure how to add `AsyncReset` into the https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#firrtl-language-definition section... should it go into `type_ground`?

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -534,10 +534,11 @@ type.
 Clock types are equivalent to clock types, and are not equivalent to any other
 type.
 
-An uninferred `Reset`{.firrtl} can be connected to either `UInt<1>`{.firrtl} or
-`AsyncReset`{.firrtl}, but not both.
+An uninferred `Reset`{.firrtl} can be connected to another `Reset`{.firrtl},
+`UInt`{.firrtl} of unknown width, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}.
+It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
 
-The `AsyncReset`{.firrtl} type is not equivalent to any other type.
+The `AsyncReset`{.firrtl} type is only equivalent to `AsyncReset`{.firrtl}.
 
 Two vector types are equivalent if they have the same length, and if their
 element types are equivalent.
--- a/spec.md
+++ b/spec.md
@@ -534,6 +534,11 @@ type.
 Clock types are equivalent to clock types, and are not equivalent to any other
 type.
 
+An uninferred `Reset`{.firrtl} can be connected to either `UInt<1>`{.firrtl} or
+`AsyncReset`{.firrtl}, but not both.
+
+The `AsyncReset`{.firrtl} type is not equivalent to any other type.
+
 Two vector types are equivalent if they have the same length, and if their
 element types are equivalent.
 
--- a/spec.md
+++ b/spec.md
@@ -2036,7 +2036,7 @@ Notationally, the width of an argument e is represented as w~e~.
 
 ## Add Operation
 
-| Name | Arguments | Parmaeters | Arg Types     | Result Type | Result Width                |
+| Name | Arguments | Parameters | Arg Types     | Result Type | Result Width                |
 |------|-----------|------------|---------------|-------------|-----------------------------|
 | add  | (e1,e2)   | ()         | (UInt,UInt)   | UInt        | max(w~e1~,w~e2~)+1          |
 |      |           |            | (SInt,SInt)   | SInt        | max(w~e1~,w~e2~)+1          |
@@ -2047,7 +2047,7 @@ The add operation result is the sum of e1 and e2 without loss of precision.
 ## Subtract Operation
 
 
-| Name | Arguments | Parmaeters | Arg Types     | Result Type | Result Width                |
+| Name | Arguments | Parameters | Arg Types     | Result Type | Result Width                |
 |------|-----------|------------|---------------|-------------|-----------------------------|
 | sub  | (e1,e2)   | ()         | (UInt,UInt)   | UInt        | max(w~e1~,w~e2~)+1          |
 |      |           |            | (SInt,SInt)   | SInt        | max(w~e1~,w~e2~)+1          |
@@ -2058,7 +2058,7 @@ precision.
 
 ## Multiply Operation
 
-| Name | Arguments | Parmaeters | Arg Types     | Result Type | Result Width                |
+| Name | Arguments | Parameters | Arg Types     | Result Type | Result Width                |
 |------|-----------|------------|---------------|-------------|-----------------------------|
 | mul  | (e1,e2)   | ()         | (UInt,UInt)   | UInt        | w~e1~+w~e2~                 |
 |      |           |            | (SInt,SInt)   | SInt        | w~e1~+w~e2~                 |
@@ -2070,7 +2070,7 @@ precision.
 ## Divide Operation
 
 
-| Name | Arguments | Parmaeters | Arg Types   | Result Type | Result Width |
+| Name | Arguments | Parameters | Arg Types   | Result Type | Result Width |
 |------|-----------|------------|-------------|-------------|--------------|
 | div  | (num,den) | ()         | (UInt,UInt) | UInt        | w~num~       |
 |      |           |            | (SInt,SInt) | SInt        | w~num~+1     |
@@ -2081,7 +2081,7 @@ of a division where den is zero is undefined.
 
 ## Modulus Operation
 
-| Name | Arguments | Parmaeters | Arg Types   | Result Type | Result Width       |
+| Name | Arguments | Parameters | Arg Types   | Result Type | Result Width       |
 |------|-----------|------------|-------------|-------------|--------------------|
 | rem  | (num,den) | ()         | (UInt,UInt) | UInt        | min(w~num~,w~den~) |
 |      |           |            | (SInt,SInt) | SInt        | min(w~num~,w~den~) |
@@ -2094,7 +2094,7 @@ satisfies the relationship below:
 
 ## Comparison Operations
 
-| Name   | Arguments | Parmaeters | Arg Types     | Result Type | Result Width |
+| Name   | Arguments | Parameters | Arg Types     | Result Type | Result Width |
 |--------|-----------|------------|---------------|-------------|--------------|
 | lt,leq |           |            | (UInt,UInt)   | UInt        | 1            |
 | gt,geq | (e1,e2)   | ()         | (SInt,SInt)   | UInt        | 1            |
@@ -2107,7 +2107,7 @@ returns a value of zero otherwise.
 
 ## Padding Operations
 
-| Name | Arguments | Parmaeters | Arg Types | Result Type | Result Width                |
+| Name | Arguments | Parameters | Arg Types | Result Type | Result Width                |
 |------|-----------|------------|-----------|-------------|-----------------------------|
 | pad  | \(e\)     | \(n\)      | (UInt)    | UInt        | max(w~e~,n)                 |
 |      |           |            | (SInt)    | SInt        | max(w~e~,n)                 |
@@ -2121,7 +2121,7 @@ padding.
 
 ## Interpret As UInt
 
-| Name   | Arguments | Parmaeters | Arg Types    | Result Type | Result Width |
+| Name   | Arguments | Parameters | Arg Types    | Result Type | Result Width |
 |--------|-----------|------------|--------------|-------------|--------------|
 | asUInt | \(e\)     | ()         | (UInt)       | UInt        | w~e~         |
 |        |           |            | (SInt)       | UInt        | w~e~         |
@@ -2133,7 +2133,7 @@ The interpret as UInt operation reinterprets e's bits as an unsigned integer.
 
 ## Interpret As SInt
 
-| Name   | Arguments | Parmaeters | Arg Types    | Result Type | Result Width |
+| Name   | Arguments | Parameters | Arg Types    | Result Type | Result Width |
 |--------|-----------|------------|--------------|-------------|--------------|
 | asSInt | \(e\)     | ()         | (UInt)       | SInt        | w~e~         |
 |        |           |            | (SInt)       | SInt        | w~e~         |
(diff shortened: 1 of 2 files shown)


mwachs5:
Should this PR include:
 - the notion of abstract resets of type `Reset`, sync resets of type `Bool`, async resets of type `AsyncReset`
 - rules for which types can drive which other types
 - any operators for converting between them?
 - Rules about what FIRRTL levels can contain uninferred resets (`Reset`) type
 - Rules about reset type on extmodules (can they have uninferred `Reset`, / AsyncReset types?)

seldridge:
I would suggest adding this into the Types section of the document as opposed to putting it in the documentation of Registers.  It needs to be called out that a register can take a reset of type reset, asyncreset, or uint<1>, but asyncreset is a fundamental type.

The main things that should be bundled with this is the reset inference algorithm. That is essentially three bullet points taken from here: https://github.com/chipsalliance/firrtl/blob/577327379a16e7329f22e971d8dd913c0bdbaaca/src/main/scala/firrtl/transforms/InferResets.scala#L89. There may also be useful information her: https://github.com/llvm/circt/blob/main/lib/Dialect/FIRRTL/Transforms/InferResets.cpp#L306. (`FullAsyncResetTransform` isn't part of the spec.)

> should it go into type_ground?

Yeah, that seems like the best place for it.

I may have erred towards mixing syntax with semantics there. Changing that to just use `type` instead of the `type_aggregate`/`type_ground` split in a follow-up PR seems logical to me.

debs-sifive:
@seldridge can you help me with some Latex stuff? I can't get the `####`-level headers + following paragraphs to display nicely without the weird indent :(

I tried `\noindent`, but that is not working (at least in my env).

Screen Shot 2022-06-29 at 11 26 33 AM

spec.md
```diff
@@ -2703,7 +2805,7 @@ info = "@" , "[" , { string , " " , linecol } , "]" ;
 (* Type definitions *)
 width = "<" , int , ">" ;
 binarypoint = "<<" , int , ">>" ;
-type_ground = "Clock"
+type_ground = "Clock" | "AsyncReset"
```
debs-sifive, line 2808: @seldridge I'm not sure how this is supposed to look 😬

spec.md
```diff
@@ -320,6 +320,68 @@ The clock type is specified as follows:
 Clock
 ```
 
+### Reset Type
+
+The abstract `Reset`{.firrtl} type is either inferred to `UInt<1>`{.firrtl}
+(synchronous reset) or `AsyncReset`{.firrtl} (asynchronous reset) after
+compilation.
```
seldridge, line 327: Nit:
```suggestion
The abstract `Reset`{.firrtl} type is either inferred to `UInt<1>`{.firrtl}
(synchronous reset) or `AsyncReset`{.firrtl} (asynchronous reset) after
compilation.
```
```suggestion
The abstract `Reset`{.firrtl} type is either inferred to `UInt<1>`{.firrtl}
(synchronous reset) or `AsyncReset`{.firrtl} (asynchronous reset) during
compilation.
```

spec.md
```diff
@@ -320,6 +320,68 @@ The clock type is specified as follows:
...
+1. An abstract reset driven by and/or driving only asynchronous resets will be
+inferred as asynchronous reset
+1. An abstract reset driven by and/or driving both asynchronous and synchronous
+resets will error
+1. Otherwise, the reset is inferred as synchronous (i.e. the abstract reset is
+only invalidated or is driven by or drives only synchronous resets)
+
+Note that an exception will be thrown if a `Reset`{.firrtl} is driven by
+both a synchronous and asynchronous type; last connect semantics will not
+determine the final type.
```
seldridge, line 368: Everything before the `;` is restating enumeration item (2). Consider removing.

seldridge (approved):
LGTM

spec.md
```diff
@@ -2703,7 +2782,7 @@ info = "@" , "[" , { string , " " , linecol } , "]" ;
 (* Type definitions *)
 width = "<" , int , ">" ;
 binarypoint = "<<" , int , ">>" ;
-type_ground = "Clock"
+type_ground = "Clock" | "AsyncReset"
```
mwachs5, line 2785: Where is Reset (uninferred) in this table?

spec.md
```diff
@@ -320,6 +320,67 @@ The clock type is specified as follows:
...
+### Reset Type
+
+The abstract `Reset`{.firrtl} type is either inferred to `UInt<1>`{.firrtl}
+(synchronous reset) or `AsyncReset`{.firrtl} (asynchronous reset) during
+compilation.
+
+Synchronous resets used in registers will be mapped to a hardware description
+language representation for synchronous resets.
+
+The following example demonstrates usage of a synchronous reset.
```
mwachs5, line 332:
```suggestion
The following example shows an uninferred reset which can/must/will be lowered to a `UInt<1>`{.firrtl} or `AsyncReset`{.firrtl}.
```

spec.md
```diff
@@ -320,6 +320,67 @@ The clock type is specified as follows:
...
+``` firrtl
+input clock : Clock
+input r : UInt<1>
+input x : UInt<8>
+wire reset: Reset
+reset <= r
+reg y : UInt<8>, clock with : (reset => (reset, UInt(123)))
+; ...
+```
+
```
mwachs5, line 343:
```suggestion
The following shows a synchronous reset...
```

show the thing with type `input reset : UInt<1>`

spec.md
```diff
@@ -320,6 +320,67 @@ The clock type is specified as follows:
...
+input clock : Clock
+input reset : AsyncReset
+input x : UInt<8>
+reg y : UInt<8>, clock with : (reset => (reset, UInt(123)))
+; ...
+```
+
+Inference rules are as follows:
+
+1. An abstract reset driven by and/or driving only asynchronous resets will be
```
mwachs5, line 359: let's pick a consistent term of either `abstract` or `uninferred`

spec.md
```diff
@@ -320,6 +320,67 @@ The clock type is specified as follows:
...
+inferred as asynchronous reset
+1. An abstract reset driven by and/or driving both asynchronous and synchronous
+resets will error
+1. Otherwise, the reset is inferred as synchronous (i.e. the abstract reset is
+only invalidated or is driven by or drives only synchronous resets)
+
+Note that an exception will be thrown if a `Reset`{.firrtl} is driven by
```
mwachs5, line 366: combine with line 361

spec.md
mwachs5, line 2785:
```suggestion
type_ground = "Clock" | "AsyncReset" | "Reset"
```

spec.md
```diff
@@ -2057,19 +2121,34 @@ type has binary point p.
...
+|         |           |            | (AsyncReset) | Clock       | n/a          |
 
 The result of the interpret as clock operation is the Clock typed signal
 obtained from interpreting a single bit integer as a clock signal.
 
+## Interpret as AsyncReset
+
+| Name         | Arguments | Parameters | Arg Types    | Result Type | Result Width |
+|--------------|-----------|------------|--------------|-------------|--------------|
+| asAsyncReset | \(e\)     | ()         | (AsyncReset) | AsyncReset  | n/a          |
```
mwachs5, line 2139:
```suggestion
| asAsyncReset | \(e\)     | ()         | (AsyncReset) | AsyncReset  | n/a          |
| asAsyncReset |             |          | (Reset) | AsyncReset  | n/a          |
```

add Reset everywhere

mwachs5:
Add something to this section: https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#lofirrtl for the rule about when all `Reset` must be either `AsyncReset` or `UInt<1>`

spec.md
mwachs5: Also we need to be careful if we just stick `Reset` in here, because this is used in extmodule.port.type ... is it legal to have a `Reset` here for extmodule? I'm not sure because Chisel doesn't even let you generate that.

debs-sifive:
```suggestion
type_ground = "Clock" | "AsyncReset"
type_abstract = "Reset"
...
// have a special extmodule port type
```

youngar:
Probably need to expand this section for rules on connecting Reset and AsyncReset to other types
https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#type-equivalence

youngar:
> Last-connect semantics for Reset type rules -- can something be driven as a Bool and then later an AsyncReset? What is legal?

It might be worth noting that Reset inference happens before last-connect semantics are resolved. Sorry if this is somewhere already and I didn't read closely enough. E.g. https://github.com/chipsalliance/firrtl/blob/master/src/test/scala/firrtlTests/InferResetsSpec.scala#L139

spec.md
```diff
@@ -2786,8 +2878,9 @@ module = "module" , id , ":" , [ info ] , newline , indent ,
            { port , newline } ,
            { statement , newline } ,
          dedent ;
+port_extmodule = ( "input" | "output" ) , id , ":": , type_ground | type_aggregate , [ info ] ;
 extmodule = "extmodule" , id , ":" , [ info ] , newline , indent ,
-              { port , newline } ,
+              { port_extmodule , newline } ,
```
debs-sifive, line 2883: @seldridge I think this is a bit messy, but you said you wanted to clean up the types later, so not sure if this flies. The notable thing that @mwachs5 and I discovered yesterday is that extmodules can't have uninferred `Reset`-type ports.

spec.md
```diff
@@ -320,6 +320,71 @@ The clock type is specified as follows:
...
+; ...
+```
+
+Inference rules are as follows:
+
+1. An uninferred reset driven by and/or driving only asynchronous resets will be
+inferred as asynchronous reset.
+1. An uninferred reset driven by and/or driving both asynchronous and synchronous
+resets will cause an exception to be thrown. This occurs because reset inference
+happens before last-connect semantics are resolved.
```
debs-sifive, line 369: @youngar ty for the info on when reset inference happens; I've added it here!

spec.md
```diff
@@ -469,6 +534,11 @@ type.
 Clock types are equivalent to clock types, and are not equivalent to any other
 type.
 
+An uninferred `Reset`{.firrtl} can be connected to either `UInt<1>`{.firrtl} or
+`AsyncReset`{.firrtl}, but not both.
+
+The `AsyncReset`{.firrtl} type is not equivalent to any other type.
+
```
debs-sifive, line 541: @youngar current attempt at talking about reset equivalence 😬

spec.md
```diff
@@ -2786,8 +2884,9 @@ module = "module" , id , ":" , [ info ] , newline , indent ,
            { port , newline } ,
            { statement , newline } ,
          dedent ;
+port_extmodule = ( "input" | "output" ) , id , ":": , type_ground | type_aggregate , [ info ] ;
```
seldridge, line 2887: This type of change blurs the line between syntax and semantics while also making the EBNF grammar more confusing. I.e., you likely do not have to make this differentiation because the EBNF grammar is only expressing legal syntax. If that fails type checking / type inference rules later, that's fine.

A good litmus test here is, "Should a FIRRTL _parser_ reject this code?" I would argue no.

I would not make the distinction between `port` and `port_extmodule`.

seldridge (approved):
LGTM

One comment about EBNF syntax vs. semantics.

spec.md
debs-sifive: kk, I will add `"Reset"` into `type_ground` then as well to simplify

spec.md
mwachs5: Ok, if we are not talking about this here we should mention it in the extmodule section I guess.

Overall I would desire that we could have inferred Resets (both inputs and outputs) on extmodule since it simplifies things  -- the other rules should kick in to interpret and enforce the inference rules. Such as if I have a Reset coming out of a black box and it drives an AsyncReset, the other rules already say we should know that extmodule's output is AsyncReset

mwachs5: ok great. Let's just put a bit in the `extmodule` section then about that?

seldridge: So, technically that test is wrong. SFC allows uninferred resets on external module inputs, but not outputs. The test should be split into one test of input and one test of output.

I am actually not sure that this restriction has to be made. I.e., the reset inference algorithm should always do the right thing here.

debs-sifive: hmmm does MFC support resets on extmodule ports? I guess we can leave this stuff out if so.

seldridge: For the following test:

debs-sifive: Cool, I'll just not say anything special about extmods then, since that makes the spec simpler, too!

seldridge (approved):

mwachs5 (changes requested):
Sorry one more thing, I realized this sentence needs to be expanded: 

https://github.com/chipsalliance/firrtl-spec/pull/25/files#diff-bc6661da34ecae62fbe724bb93fd69b91a7f81143f2683a81163231de7e3b545L815

"Note that the clock signal for a register must be of type `clock`{.firrtl}, the
reset signal must be a single bit `UInt`{.firrtl}, and the type of
initialization value must be equivalent to the declared type of the register
(see [@sec:type-equivalence] for details)."

Need to expand the "single bit `UInt`{.firrtl}`" to allow Reset or AsyncReset as well.

spec.md
```diff
@@ -800,9 +872,9 @@ reg myreg: SInt, myclock
 ```
 
-declared with a reset signal and value. In the following example,
-`myreg`{.firrtl} is assigned the value `myinit`{.firrtl} when the signal
-`myreset`{.firrtl} is high.
+declared with a reset signal (synchronous or asynchronous) and value. In the
```
mwachs5, line 875:
```suggestion
declared with a reset signal (uninferred, synchronous, or asynchronous) and value. In the
```

seldridge:
> Also, not sure what the actual firrtl rules are, but is there a firrtl spec requirement that AsyncReset type must have a literal value for the initializer (@jackkoenig , @seldridge ?) Or is that just a "good design practice" requirement

This is a check that is enforced by the SFC (and also enforced by the MFC due to the SFC doing this). This is a good practice as it guarantees that the reset initial value will not violate setup/hold time when the reset is asserted, but is not a one-size-fits-all solution. Legal counterexamples that violate this simple check.

1. A reset scheme where the asynchronous reset initial value is non-constant, but the reset is only ever applied while the reset is clock-gated.
2. A reset scheme where the asynchronous reset initial value is non-constant, but is synchronous with respect to the asynchronous reset signal.
3. A reset scheme where the asynchronous reset initial value is constant, but this is driven from a port that is outside the circuit and the two circuits are being compiled separately.

FIRRTL IR is arguably the wrong thing to even try to capture this and it requires higher-level IRs or is purely a front-end language concern.

There are two paths that make sense to me: (1) just be silent here, (2) provide detailed information about the complications that arise in circuits which have asynchronous reset, or (3) suggest that a FIRRTL compiler may implement a check to disallow non-constant reset initial value.

For the purposes of this PR, I think (1) is the best incremental approach.

mwachs5 (approved):
Thanks so much for doing this!

PR #29, opened 2022-08-05
Add an optional enable signal to registers.

darthscsi:
Representation of enable as a mux feeding a register is insufficient as specific tools care about the form of the output for reasoning about enable.  Specifically, they need to be output as if blocks, not as muxes.  However, if and mux have different x behavior in verilog, making it not safe to generally replace a mux with an if.  Further, there is a desire to distingish between a mux before a register and a register with an enable port, making automatic conversion not possible.  Explicitly encoding the enable signal solves both these problems and allows direct expression of intent.

This allows directly capturing the intent expressed by chisel here: https://github.com/chipsalliance/chisel3/blob/v3.5.3/src/main/scala/chisel3/util/Reg.scala#L7

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -901,10 +901,32 @@ reg myreg: SInt, myclock with: (reset => (myreset, myinit))
 ; ...
 ```
 
-Note that the clock signal for a register must be of type `clock`{.firrtl},
-the reset signal must be a `Reset`{.firrtl}, `UInt<1>`{.firrtl}, or
-`AsyncReset`{.firrtl}, and the type of initialization value must be equivalent
-to the declared type of the register (see [@sec:type-equivalence] for details).
+Optionally, a register can be
+declared with an enable signal. In the following example,
+`myreg`{.firrtl} is written the value `myval`{.firrtl} when the signal
+`myen`{.firrtl} is high.  Enable must
+be a value of type UInt<1>.
+
+``` firrtl
+wire myclock: Clock
+wire myen: UInt<1>
+wire myval: SInt
+reg myreg: SInt, myclock with: (enable => myen)
+myreg <= myval
+; ...
+```
+
+When both reset and enable are present, reset shall be listed first, separated
+by a comman from enable.  Reset of the register is unaffected by the enable 
+signal, depending only on reset signal.
+```
+reg myreg: SInt, myclock with: (reset => (myreset, myinit), enable => myen)
+```
+
+
+Note that the clock signal for a register must be of type `clock`{.firrtl}, the
+reset signal must be a single bit `UInt`{.firrtl}, and the type of
+initialization value must match the declared type of the register.
 
 ## Invalidates
 
--- a/spec.md
+++ b/spec.md
@@ -880,10 +880,12 @@ A register is a named stateful circuit component.  Reads from a register return
 the value of the element, writes are not visible until after the positive edges 
 of the register's clock port.
 
+The clock signal for a register must be of type `Clock`{.firrtl}.  The type of a 
+register must be a passive type (see [@sec:passive-types]).
+
 The following example demonstrates instantiating a register with the given name
 `myreg`{.firrtl}, type `SInt`{.firrtl}, and is driven by the clock signal
-`myclock`{.firrtl}.  The clock signal for a register must be of type 
-`Clock`{.firrtl}.
+`myclock`{.firrtl}.
 
 ``` firrtl
 wire myclock: Clock
@@ -899,9 +901,9 @@ initialization  value must be equivalent to the declared type of the register
 (see [@sec:type-equivalence] for details). The behavior of the register depends 
 on the type of the reset signal.  `AsyncReset`.{firrtl} will immediately change 
 the value of the register, while other types will not change the value of the 
-register until the next positive edge of the clock signal.  In the following 
-example, `myreg`{.firrtl} is assigned the value `myinit`{.firrtl} when the 
-signal `myreset`{.firrtl} is high.  
+register until the next positive edge of the clock signal (see [@sec:reset-type]).
+In the following example, `myreg`{.firrtl} is assigned the value 
+`myinit`{.firrtl} when the signal `myreset`{.firrtl} is high.  
 ``` firrtl
 wire myclock: Clock
 wire myreset: UInt<1>
@@ -2907,7 +2909,12 @@ memory = "mem" , id , ":" , [ info ] , newline , indent ,
 (* Statements *)
 statement = "wire" , id , ":" , type , [ info ]
           | "reg" , id , ":" , type , expr ,
-            [ "(with: {reset => (" , expr , "," , expr ")})" ] , [ info ]
+            [ "(with: {"
+              ( "reset => (" , expr , "," , expr ")"
+              | "enable => (" , expr , ")"
+              | "reset => (" , expr , "," , expr "), enable => (" , expr , ")"
+              ) , "})" ,
+            [ info ]
           | memory
           | "inst" , id , "of" , id , [ info ]
           | "node" , id , "=" , expr , [ info ]
--- a/spec.md
+++ b/spec.md
@@ -910,12 +910,13 @@ reg myreg: SInt, myclock with: (reset => (myreset, myinit))
 ; ...
 ```
 
-Optionally, a register can be declared with an enable signal. Enable gates 
-writes to a register, such that they only take effect if the enable signal is 
-high.  Functionally this is equivalent to a register written by a mux of the old 
-value and the new value, controlled by the enable signal.  In the following 
+Optionally, a register can be declared with an enable signal. The enable signal 
+gates writes to a register, such that they only take effect if the enable signal 
+is high.  Functionally this is equivalent to a register written by a mux of the 
+old value and the new value, controlled by the enable signal.  In the following 
 example, `myreg`{.firrtl} is written the value `myval`{.firrtl} when the signal 
-`myen`{.firrtl} is high.  Enable must be a value of type `UInt<1>`{.firrtl}.
+`myen`{.firrtl} is high.  The enable signal must be a value of type 
+`UInt<1>`{.firrtl}.
 
 ``` firrtl
 wire myclock: Clock
--- a/spec.md
+++ b/spec.md
@@ -880,7 +880,8 @@ A register is a named stateful circuit component.
 
 The following example demonstrates instantiating a register with the given name
 `myreg`{.firrtl}, type `SInt`{.firrtl}, and is driven by the clock signal
-`myclock`{.firrtl}.
+`myclock`{.firrtl}.  The clock signal for a register must be of type 
+`clock`{.firrtl}.
 
 ``` firrtl
 wire myclock: Clock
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -901,10 +905,25 @@ reg myreg: SInt, myclock with: (reset => (myreset, myinit))
...
+wire myclock: Clock
+wire myen: UInt<1>
+wire myval: SInt
+reg myreg: SInt, myclock with: (enable => myen)
+myreg <= myval
+; ...
+```
+
+When both reset and enable are present, reset shall be listed first, separated
+by a comman from enable.  Reset of the register is unaffected by the enable
```
mwachs5, line 922:
```suggestion
by a comma from enable.  Reset of the register is unaffected by the enable 
```

spec.md
```diff
@@ -901,10 +905,25 @@ reg myreg: SInt, myclock with: (reset => (myreset, myinit))
 ; ...
 ```
 
-Note that the clock signal for a register must be of type `clock`{.firrtl},
-the reset signal must be a `Reset`{.firrtl}, `UInt<1>`{.firrtl}, or
-`AsyncReset`{.firrtl}, and the type of initialization value must be equivalent
-to the declared type of the register (see [@sec:type-equivalence] for details).
+Optionally, a register can be declared with an enable signal. In the following
+example, `myreg`{.firrtl} is written the value `myval`{.firrtl} when the signal
+`myen`{.firrtl} is high.  Enable must be a value of type UInt<1>.
```
mwachs5, line 910: should we define what the enable signal means (value is kept unless enable is true) or is that outside the scope of this spec (which doesn't explain what a register is, it seems to assume the reader already knows this)?

darthscsi: we should, I had a couple things lost in a rebase, let me revise.

darthscsi: Also the relationship between reset and enable.

mwachs5, line 910:
```suggestion
`myen`{.firrtl} is high.  The enable signal must be a value of type `UInt<1>`{.firrtl}.
```

spec.md
```diff
@@ -891,7 +892,10 @@ reg myreg: SInt, myclock
 Optionally, for the purposes of circuit initialization, a register can be
 declared with a reset signal and value. In the following example,
 `myreg`{.firrtl} is assigned the value `myinit`{.firrtl} when the signal
-`myreset`{.firrtl} is high.
+`myreset`{.firrtl} is high.  The reset signal must be a `Reset`{.firrtl},
+`UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}, and the type of initialization 
+value must be equivalent to the declared type of the register 
+(see [@sec:type-equivalence] for details).
```
seldridge: SFC requires the reset to be `UInt<1>`, but MFC will accept an unknown width `UInt` that must infer to a `UInt<1>`. It seems fine to start with the weaker MFC interpretation here.

spec.md
seldridge: Agree. Define the behavior, i.e., "hold previous value".

seldridge: Should enable be a `UInt<1>` or an unknown width that infers to a `UInt<1>`? I prefer the latter interpretation.

spec.md
darthscsi: This specifies Uint<1>

seldridge:
@azidar wants to take a look at this before it gets merged.

aswaterman:
It's not clear to me this is the right way to go.  We want to be able to emit trees of if-else statements for a single register, so being express only one enable doesn't seem to capture the entirety of the pattern.

Certainly, we don't want to synthesize our own single enables from complex conditionals.  If users write `when (en1) { r := wdata1 }; when (en2) { r := wdata2 }`, we don't want to emit `if (en2 || en1) r <= en2 ? wdata2 : wdata1;`.  We want to emit `if (en2) r <= wdata2; else if (en1) r <= wdata1;`, for two reasons: the code is more readable, and experiments from several years ago favored the latter construct.  (I never discovered why; it's possible that the former interfered with DC's pattern-matching of mux trees.)

Finally, I disagree with the premise that we need to express _any_ enables in the IR to accomplish this goal.  There's an existence proof: we were able to emit if-else trees for registers just fine in SFC without a special Firrtl construct.

sequencer:
`if-else` or `C?T:F` will be both synthesized to Mux tree, but as @aswaterman said, designer may need use `if-else` pattern to ask synthesis tool to the control the shape of the tree, while `C?T:F` may not giving to much room for optimization, it sounds like overfitting to the commercial tools like DC/Genus. But in the ASIC world, we cannot avoid this. :(

As for the `X` signal issues, I don't think it should be handled by any special code paradigm, we should guarantee the generated ASIC cells won't have `X` signals, I believe that may related to `last-connect-semantic` but not the `enable` signal.

From an ASIC designer perspective, I have been regarding `enable` signal to be the signal to connect the `ICG` cell for register, although Chisel/FIRRTL doesn't provide this function, but I do think the semantic looks same.
So if we are going to add the `enable` signal to FIRRTL IR, although most of modern vendor tools can automatically insert `ICG`, I still think this can be a option for the fine tune clock-gating circuit, rather than designing for X-prop or using it to indicate DC to auto-generated a specific mux trees.

So I don't think adding `enable` signal or emit a special code paradigm can fundamentally resolve this issue.

But I still have something which I think register should have:
1. Clock but with the clock edge, I prototyped it a long time ago in chipsalliance/firrtl/#1125
2. Reset and Reset Type(Async, Sync), Reset Input Value(constraint to be constant)
3. Preset value(constraint to be constant), this is useful in the FPGA design since the register do provide the power on initialization and also support the `inital` block in verilog, and Chisel already support it via `PresetAnnotation`.
4. `ICG` enable signal to indicate how to clock gate a register.
5. `DFT` related metadata(but I don't have a good answer yet, since I don't have enough quota on Tessent LOL)

darthscsi:
> There's an existence proof: we were able to emit if-else trees for registers in SFC, including those with complex conditionals, without a special Firrtl construct

But in doing so, SFC changes semantics regarding x-prop.  Generally SFC plays fairly fast and loose with mux and if.  The conversion from when -> mux -> if is done in a context-dependent manner, happening for some muxes and not others.  You can easily break existing code by, say, converting all muxes to ifs.  So, what in the source code indicates when it is safe and when it isn't?  We can also change firrtl to make it explicit that it's always safe to do mux -> if, but then we also have to deal with the fallout when doing so breaks existing designs (and it does, we've tried).  The untenable position is what SFC does now which is "oh, it looks like this might be an enable, so I'll do a semantics changing transformation in this case".

darthscsi:
Consider:
```
import chisel3._


class Foo extends Module {
  val io = IO(new Bundle {
    val a = Input(Bool())
    val b = Input(UInt(4.W))
    val c = Input(UInt(4.W))
    val d = Output(UInt(4.W))
    val e = Output(UInt(4.W))
  })
  val r = Reg(UInt(4.W));
  val t = Mux(io.a, io.b, io.c)
  io.d := t
  r := t
  io.e := r
}
```
Here we have a single mux statement which SFC generates in two ways which have different verilog semantics.  The assignment in the generated verilog to r and to io_d are not equivalent.  Now, we can clarify that firrtl is strictly a two-state logic language and a compiler is allowed to generate code which is equivalent under that restriction, but in practice, that isn't true.  Verification code in Chisel explicitly tests for 'x', and assertions in real code bases break if you change x-prop behavior of the generated code.
```
  reg [3:0] r; // @[main.scala 14:14]
  assign io_d = io_a ? io_b : io_c; // @[main.scala 15:14]
  assign io_e = r; // @[main.scala 18:8]
  always @(posedge clock) begin
    if (io_a) begin // @[main.scala 15:14]
      r <= io_b;
    end else begin
      r <= io_c;
    end
  end
```

seldridge:
The reason why I want this is that it is currently ambiguous what a user means when they use `RegEnable`, `when`, or `mux` and this is working towards clarifying that. 

I think we all agree how to lower `val r = RegEnable(r_next, en)` from Chisel. However, what does a user mean when the `RegEnable` is nested under a `when`? E.g.:

```scala
when (en2) {
  val r = RegEnable(r_next, en)
}
```

Noting that the FIRRTL produced by Chisel is:

```
when en2:
  when en:
    reg r: UInt<1>, clock
    r <= r_next
```

Put differently, is the enable signal `en2`, `en`, or `en2 && en`? Or: which of the following did the user intend when they wrote the above Chisel code?

```verilog
always @(posedge clock) begin
  /* (1) Straight when lowering */
  if (en2) begin
    if (en) begin
      r <= r_next;
    end
  end

  /* (2) Use RegEnable enable */
  if (en) begin
    if (en2) begin
      r <= r_next;
    end 
  end 

  /* (3) And of the two enables */
  if (en2 && en) begin
    r <= r_next;
  end
end
```

Adding an explicit enable allows FIRRTL compilers to unambiguously know what the user intended. This is also a mechanism to guarantee that we don't remove an enable that the user requested which may, hypothetically, work against Chisel users trying to decrease dynamic power (which seems implied by the `RegEnable` API). Example (1) is an enable of `en2`, example (2) is an enable of `en`, and example (3) is an enable of `and(en2, en)`.

Notably, this proposal does not define how to lower `when` statements into reg enables or if that should ever be allowable. This proposal is more narrowly trying to close the imprecision that arises around `when`, `mux`, and expectations on explicit Verilog coding styles that describe a register with an enable.

mwachs5:
I *think* that @seldridge's example is weird code for a user to write and not the best example of the pattern we're trying to clarify.

```
when (en2) {
  val r = RegEnable(r_next, en)
}
```

I think a more likely Chisel user code is:

```
val foo = RegEnable(r_next, en)

when (something) {
  foo := somethingElse
}

when(anotherThing)
  foo := anotherElse
}.otherwise {
  when (oneLastThing)
     foo := oneLastElse
  }
}
```

I think that is a better example to explore "what did the user mean and what should we emit". 
Frankly I don't even know. I am sort of of the feeling that it's fair to tell people who want registers to have enables to use `RegEnable`, but am sympathetic to @aswaterman argument that is changing the rules under user's feet a bit.

darthscsi:
@seldridge's code does point out the problem with trying to infer intent through raising, you have to deal with all these weird, but possible cases (and declarations inside whens happen, so it's not that weird).  Note that not all the possible intended outputs would be fixed with explicit enable representation (and the multiple different enables is also an interesting case).

Dealing with defining how/if/when firrtl transforms can change x-behavior problem is important, this has blocked PPA and readability improvements, and it is why we can't just output if for enable-like signals.

aswaterman:
I'm sympathetic to the X-prop concerns.

But as a separate matter, I maintain that it wasn't ambiguous that users of Chisel expected enables to have been inferred from both `when` and `RegEnable` (on the basis that it worked that way for roughly a decade and Chisel users are used to that behavior).

darthscsi:
The core issue preventing this happening (and Circt use to do this) is the non-equivalence of if and mux.  Specifying firrtl as only preserving 2-state behavior, which seems the general intention of the spec, makes these transformations safe.

aswaterman:
Yeah, there has been a general philosophy that transformations might not be 4-state equivalent.  (Even some basic things like constant propagation aren't 4-state equivalent, e.g., folding `0 << foo` into `0` eliminates an X when `foo` is X... because Verilog.)

seldridge:
We've seen this crop up before with the infamous `div(a, a) -> 1` folder in SFC which we pulled into MFC. (Long, long discussion: https://github.com/llvm/circt/issues/363) This is illegal if done in SV and can cause formal mismatches if the expectation is that FIRRTL w/ optimizations compiled to Verilog is formally equivalent to FIRRTL w/o optimizations compiled to Verilog.

The thing @darthscsi is referring to is that we added a folder _in the SV Dialect_ that would transform mux to if. The idea was that this could be used to convert the synchronous enable to the proper format. However, this is clearly not legal in SV and we ran into simulation failures.

The behavior of SFC of digging backwards through all the muxes that drive a register and sinking them into the always block may be a sufficient mechanism for better Verilog emission (and we can try this). I had hoped that adding a register enable would provide a clear non-heuristic/pattern matching approach to define the enable for a register and enable a more obvious `RegEnable` API. This is complicated because `RegEnable` has been sugar for `reg + when` since, as you correctly state, [the literal beginning of time](https://github.com/ucb-bar/chisel2-deprecated/commit/3e6b6c140e7aa461f4b252d26635a7e46b2f415e) and well into FIRRTL pre-history.

Issue #27, opened 2022-08-03
RFC: Capture the version of the FIRRTL spec in a .fir file


mwachs5:
As the FIRRTL specification may evolve, it would be nice to know what version of the specification a given .fir file claims to comply with.

Considerations:

- Should this just be a single version, or a range / regex?
- What should be the change on the spec version due to the addition of this version (what are FIRRTL's own versioning rules)?

darthscsi:
A file should only claim one version.  If it uses a subset of the language compatible with multiple versions, it should just specify the least.  But generally, it would be annoying to write an emitter to try to track ranges of versions based on what gets emitted.  Parseres already have to deal with multiple versions, so I don't think we need producers to also.

mwachs5:
> What should be the change on the spec version due to the addition of this version (what are FIRRTL's own versioning rules)?

I documented this in 1.0.0, so presumably adding something as an optional with a default-if-not-specified-is-1.0.0 would bump the MINOR version

seldridge:
Fixed via #30.

PR #30, opened 2022-08-05
Add a version preamble to fir files.  

darthscsi:
Make the version include a unix-style  magic string for file identification.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -182,8 +182,9 @@ contributors is below:
 
 # File Preamble
 
-A firrtl file includes a magic string and version identifier indicating the 
-version of this standard the file conforms to.  This will not be present on 
+A firrtl file begins with a magic string and version identifier indicating the 
+version of this standard the file conforms to 
+(see [@sec:versioning-scheme-of-this-document]).  This will not be present on 
 files generated according to versions of this standard prior to the first 
 versioned release of this standard to include this preamble.
 
@@ -2919,8 +2920,11 @@ extmodule = "extmodule" , id , ":" , [ info ] , newline , indent ,
               { "parameter" , "=" , ( string | int ) , newline } ,
             dedent ;
 
+(* Wersion definition *)
+version = "FIRRTL" , "version" , {digit_dec} , "."  , {digit_dec} , "." , {digit_dec}
+
 (* Circuit definition *)
-circuit = "circuit" , id , ":" , [ info ] , newline , indent ,
+circuit = version , newline , "circuit" , id , ":" , [ info ] , newline , indent ,
             { module | extmodule } ,
           dedent ;
 ```
--- a/spec.md
+++ b/spec.md
@@ -189,7 +189,7 @@ files generated according to versions of this standard prior to the first
 versioned release of this standard to include this preamble.
 
 ``` firrtl
-firrtl version 1.1.0
+FIRRTL version 1.1.0
 circuit MyTop...
 ```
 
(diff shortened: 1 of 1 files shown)


darthscsi:
closes #27

spec.md
```diff
@@ -180,6 +180,18 @@ contributors is below:
 - [`@tdb-alcorn`](https://github.com/tdb-alcorn)
 - [`@youngar`](https://github.com/youngar)
 
+# File Preamble
+
+A firrtl file includes a magic string and version identifier indicating the 
+version of this standard the file conforms to.  This will not be present on 
+files generated according to versions of this standard prior to the first 
+versioned release of this standard to include this preamble.
+
```
mwachs5, line 189: does this have to be the first thing in the file, or can it occur anywhere (the term "preamble" implies that it should be the first thing)? Is the "magic string" `firrtl` or is it `firrtl version`? Should this get added to the language reference at the end of the doc?

mwachs5: also why do we need to call this a "magic string". What is magic about it, vs just saying the version is specified in this way

spec.md
```diff
@@ -180,6 +180,18 @@ contributors is below:
...
+``` firrtl
+firrtl version 1.1.0
```
seldridge, line 191: What about making this a magic comment string as opposed to a magic string?

```suggestion
; firrtl version 1.1.0
```

seldridge:
Logging verbal confirmation from @azdiar that he approves of this.

spec.md
darthscsi: magic string being an in-data method of determining file type from contents, not extension.  This is used by file on unix (source of the term magic, since that's the file containing the patterns) and for identifying files for mime types.  Since we are adding a version anyway, having a header which can be used for this purpose adds little costs, only 7 bytes, and makes the format a tiny bit more self-identifying.
https://linux.die.net/man/1/file
https://en.wikipedia.org/wiki/List_of_file_signatures

spec.md
darthscsi: If it's required, why make it a comment?

mwachs5: i agree making it a comment just seems to add complexity

spec.md
mwachs5: cool, thanks for the explanation

spec.md
```diff
@@ -2907,8 +2920,11 @@ extmodule = "extmodule" , id , ":" , [ info ] , newline , indent ,
               { "parameter" , "=" , ( string | int ) , newline } ,
             dedent ;
 
+(* Wersion definition *)
```
mwachs5, line 2923:
```suggestion
(* Version definition *)
```

spec.md
```diff
@@ -180,6 +180,19 @@ contributors is below:
...
+# File Preamble
+
+A firrtl file begins with a magic string and version identifier indicating the 
+version of this standard the file conforms to 
+(see [@sec:versioning-scheme-of-this-document]).  This will not be present on 
+files generated according to versions of this standard prior to the first 
+versioned release of this standard to include this preamble.
+
+``` firrtl
+firrtl version 1.1.0
```
mwachs5, line 192:
```suggestion
FIRRTL version 1.1.0
```

i think your language table below has this as all-caps?

darthscsi: yea, meant to check that.  thanks.

seldridge (approved):

mwachs5 (approved):
thanks!

PR #31, opened 2022-08-06
[major] Remove Fixed Point Types

seldridge:
Remove Fixed Point and any references to it from the FIRRTL
specification.  This (and intervals) was not a FIRRTL feature that
achieved significant traction nor was it ever load-bearing in the same
way that the rest of the specification has been for many years.

This is not in any way intended to be the end of Fixed Point
representations in FIRRTL (or Chisel)!  There are planned future
improvements to provide Fixed Point libraries either directly in Chisel
or via user-defined FIRRTL types once we sort through what that should
look like.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -252,7 +252,7 @@ composed of one or more aggregate or ground types.
 
 ## Ground Types
 
-There are five ground types in FIRRTL: an unsigned integer type, a signed
+There are four ground types in FIRRTL: an unsigned integer type, a signed
 integer type, a clock type, and an analog type.
 
 ### Integer Types
--- a/spec.md
+++ b/spec.md
@@ -253,7 +253,7 @@ composed of one or more aggregate or ground types.
 ## Ground Types
 
 There are five ground types in FIRRTL: an unsigned integer type, a signed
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -271,7 +271,7 @@ are built up from aggregating other types.
 ## Ground Types
 
 There are five ground types in FIRRTL: an unsigned integer type, a signed
```
mwachs5, line 273: four ground types?

seldridge: Good catch. Fixed in the first `fixup!` commit.

mwachs5 (commented):
nit: can you follow the CONTRIBUTING guidelines to indicate that this is a (MAJOR?) change

seldridge:
> nit: can you follow the CONTRIBUTING guidelines to indicate that this is a (MAJOR?) change

Not a nit. 😂 Something I need to actually do. (To be fair, this PR pre-dated that.) Updated with the tag and with inclusion of information in the `revision-history.yaml`.

seldridge:
Logging verbal approval from @azidar on this.

PR #34, opened 2022-08-10
Use git-tag to compute spec version

seldridge:
Change the way that the version is computed to be derived from the
latest git tag that looks like `v*.*.*`.  This _should_ compose
correctly with the GitHub Action to build/publish the spec once a new
tag is created.  This also has the benefit of never having to manually
update the spec version.  Instead, only a new tag has to be created.

Closes #33 (as this is an alternative that avoids ever having to bump the version again).

If you build off of a spec version that is some commits after the latest tag you get something like the following:
Screen Shot 2022-08-09 at 21 34 10

A clean tag (which is what the CI will do on a release) will give you something like:
Screen Shot 2022-08-09 at 21 35 29
For a git history of:

```
* d73da20 (HEAD -> dev/seldridge/use-git-tag-for-version, tag: v1.42.1337, origin/dev/seldridge/use-git-tag-for-version) Use git-tag to compute spec version
* 9aa2f29 (origin/main, origin/HEAD, main) Simplify Makefile image target computation, NFC
* e7199fc Cleanup EBNF Grammar, NFC
* 14660c3 Cleanup Makefile, NFC
* 298b5d7 Change Makefile to work with MacOS make, NFC
| * 4f0a54c (origin/dev/seldridge/rm-fixed-point, dev/seldridge/rm-fixed-point) fixup! Remove Fixed Point Types
| * 699c3bd Remove Fixed Point Types
|/  
* 151302c Whitespace cleanup, NFC
* b1ead76 Add a version preamble to fir files.   (#30)
* a62e593 (tag: v1.0.0) Whitespace changes in spec.md, NFC
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -3,8 +3,6 @@ author:
 - The FIRRTL Specification Contributors
 title: Specification for the FIRRTL Language
 date: \today
-# Custom options added to the customized template
-version: 1.0.0
 # Options passed to the document class
 classoption:
 - 12pt
--- a/spec.md
+++ b/spec.md
@@ -39,23 +39,36 @@ secPrefix:
   - Sections
 # This 'lastDelim' option does not work...
 lastDelim: ", and"
+# Information about revision history.  This is used by the LaTeX template.
+revisionHistory:
+  # Information about what was added in the current version.  This will be
+  # populated using the "version" that the Makefile grabs from git.  Notable
+  # additions to the specification should append entries here.
+  thisVersion:
+    - Add version information to FIRRTL files
+  # Information about the old versions.  This should be static.
+  oldVersions:
+    - version: 1.0.0
+      changes:
+        - Document the versioning scheme of this specification.
+    - version: 0.4.0
+      changes:
+        - >
+          Add documentation for undocumented features of the Scala-based FIRRTL
+          Compiler (SFC) that are de facto a part of the FIRRTL specification
+          due to their widespread use in Chisel and the SFC: Annotations,
+          Targets, Asynchronous Reset,  Abstract Reset
+        - Minor typo corrections and prose clarifications.
+    - version: 0.3.1
+      changes:
+        - Clarify analog usage in registers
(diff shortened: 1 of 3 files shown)


mwachs5:
I def think this is an improvement. What should we do about the "revision history" table? Should we just get rid of that? Or we could do something like what was done in the RISC-V Debug Spec that just dumps the github changelog into the PDF :https://github.com/riscv/riscv-debug-spec/blob/master/Makefile#L60. Though it looks like that (understandably) does not make it in to the PDF anymore, or we could do some hybrid thing of automatically grabbing a changes since the last release from git?

seldridge:
> What should we do about the "revision history" table?

Great thought!

I converted this over to be encoded in YAML and have the Pandoc LaTeX template pull this information out. The version will then be automatically added.

E.g., until a new version is tagged, the `Revision History` will look like:

Screen Shot 2022-08-10 at 14 26 27

Once a tag is created, it will look like:

Screen Shot 2022-08-10 at 14 26 27

The only manual thing now is to move the `thisVersion` information into an `oldVersion` object in a commit _after_ a tag is created.

spec.md
```diff
@@ -41,24 +39,37 @@ secPrefix:
   - Sections
 # This 'lastDelim' option does not work...
 lastDelim: ", and"
+# Information about revision history.  This is used by the LaTeX template.
+revisionHistory:
+  # Information about what was added in the current version.  This will be
+  # populated using the "version" that the Makefile grabs from git.  Notable
+  # additions to the specification should append entries here.
```
mwachs5, line 46: ok so the idea is that PR authors should add something to this section for not-NFC changes? SGTM

mwachs5 (approved):

spec.md
seldridge: Yep! We can also move this into a separate `revisionHistory.yaml` file to make it more obvious (and get better support for YAML highlighting/editor modes than the inline stuff supports).

PR #37, opened 2022-08-11
Specify FIRRTL is 2-state, and limit observable behavior to that

darthscsi:
Make explicit that FIRRTL is a 2-state language and only defined on 2-state values.  This allows a compiler to ignore the x-prop and z  behavior in verilog (and similarly in VHDL) when transforming circuits.

This matches the spec to the current behavior of compilers, which have so far presumed this without it being specified.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -46,6 +46,7 @@ revisionHistory:
   # additions to the specification should append entries here.
   thisVersion:
     - Add version information to FIRRTL files
+    - Specify "As-If" limited to boolean
   # Information about the old versions.  This should be static.
   oldVersions:
     - version: 1.0.0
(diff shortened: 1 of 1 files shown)


darthscsi:
closes #29

spec.md
```diff
@@ -2692,6 +2692,15 @@ module MyModule :
    out <= r$0
 ```
 
+# Semantics of Values
+
+FIRRTL is defined for 2-state logic.  The behavior of a generated circuit in a
+language, such as Verilog or VHDL, which have multi-state logic, is undefined
+in the presence of values which are not 2-state.  A FIRRTL compiler need only
+respect the 2-state behavior of a circuit.  This is a limitation on the scope of
+what behavior is observable (e.g. a relaxation of the
```
ekiwi, line 2701: Should this be `i.e.` instead of `e.g.`?

I wonder how much the last sentence adds in general. It would be good to have a section on what it means for a firrtl compiler to "preserve semantics", but I recognize that would be a lot of work.

ekiwi (approved):
Besides my small comment, this is great!

spec.md
```diff
@@ -2692,6 +2692,15 @@ module MyModule :
    out <= r$0
 ```
 
+# Semantics of Values
+
+FIRRTL is defined for 2-state logic.  The behavior of a generated circuit in a
```
ekiwi, line 2697: This should probably be `2-state (boolean) logic` or something like that. We need to define what these two states are. If you want to rely on a common understanding around Verilog, we would have to cite the Verilog spec.

mwachs5 (commented):
please add something to the Revision History yml section for this change:
https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md?plain=1#L47

spec.md
```diff
@@ -2692,6 +2693,15 @@ module MyModule :
    out <= r$0
 ```
 
+# Semantics of Values
+
+FIRRTL is defined for 2-state boolean logic.  The behavior of a generated 
+circuit in a language, such as Verilog or VHDL, which have multi-state logic, is 
+undefined in the presence of values which are not 2-state.  A FIRRTL compiler 
+need only respect the 2-state behavior of a circuit.  This is a limitation on 
+the scope of what behavior is observable (i.e. a relaxation of the
```
seldridge, line 2702: Extreme nit:
```suggestion
the scope of what behavior is observable (i.e., a relaxation of the
```

seldridge (approved):

PR #39, opened 2022-08-12
Expand register text with at least cursory explanation

darthscsi:
The text on registers was pretty sparse and didn't hint at their behavior.  Update this and mention the type-dependent reset behavior here.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -901,7 +901,12 @@ wire mywire: UInt
 
 ## Registers
 
-A register is a named stateful circuit component.
+A register is a named stateful circuit component.  Reads from a register return 
+the value of the element, writes are not visible until after the positive edges 
+of the register's clock port.
+
+The clock signal for a register must be of type `Clock`{.firrtl}.  The type of a 
+register must be a passive type (see [@sec:passive-types]).
 
 The following example demonstrates instantiating a register with the given name
 `myreg`{.firrtl}, type `SInt`{.firrtl}, and is driven by the clock signal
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -901,7 +901,12 @@ wire mywire: UInt
 
 ## Registers
 
-A register is a named stateful circuit component.
+A register is a named stateful circuit component.  Reads from a register return 
+the value of the element, writes are not visible until after the positive edges
```
ekiwi, line 905: I would emphasize that read return the **current** value of the element.

Writes are not visible until after **a positive edge on** the register's clock port.

spec.md
```diff
@@ -914,10 +919,16 @@ reg myreg: SInt, myclock
 ```
 
-declared with a reset signal and value. In the following example,
-`myreg`{.firrtl} is assigned the value `myinit`{.firrtl} when the signal
-`myreset`{.firrtl} is high.
-
+declared with a reset signal and value. The register's value is updated with the
```
seldridge, line 922: I'd also strike the "Optionally, for the purposes of circuit initialization" (since you're here).

Reading this paragraph in its entirety is somewhat confusing because it mentions both a possible user intended function, "circuit initialization", when introducing a semantic, "register reset". This is confusing because "register initialization" at power on (which could easily be confused with "circuit initialization") is a separate, critical concept that isn't addressed by the spec.

spec.md
```diff
@@ -913,10 +918,16 @@ reg myreg: SInt, myclock
 ; ...
 ```
 
-Optionally, for the purposes of circuit initialization, a register can be
-declared with a reset signal and value. In the following example,
-`myreg`{.firrtl} is assigned the value `myinit`{.firrtl} when the signal
-`myreset`{.firrtl} is high.
+A register may be declared with a reset signal and value. The register's value
```
seldridge, line 921: Mega nit: inconsistent use of single and double space after period in this paragraph.

spec.md
```diff
@@ -913,10 +918,16 @@ reg myreg: SInt, myclock
...
 ```
 
-Optionally, for the purposes of circuit initialization, a register can be
-declared with a reset signal and value. In the following example,
-`myreg`{.firrtl} is assigned the value `myinit`{.firrtl} when the signal
-`myreset`{.firrtl} is high.
+A register may be declared with a reset signal and value. The register's value 
+is updated with the reset value when the reset is asserted.  The reset signal 
+must be a `Reset`{.firrtl}, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}, and 
+the type of initialization  value must be equivalent to the declared type of the
```
seldridge, line 924: Mega nit: double space after initialization.

spec.md
```diff
@@ -913,10 +918,16 @@ reg myreg: SInt, myclock
...
-declared with a reset signal and value. In the following example,
-`myreg`{.firrtl} is assigned the value `myinit`{.firrtl} when the signal
-`myreset`{.firrtl} is high.
+A register may be declared with a reset signal and value. The register's value 
+is updated with the reset value when the reset is asserted.  The reset signal 
+must be a `Reset`{.firrtl}, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}, and 
+the type of initialization  value must be equivalent to the declared type of the 
+register (see [@sec:type-equivalence] for details). The behavior of the register 
+depends on the type of the reset signal.  `AsyncReset`.{firrtl} will immediately 
+change the value of the register, while other types will not change the value of
```
seldridge, line 927: This is only true for synchronous reset. Abstract reset may be either a sync or sync. You can't reason about it until reset inference runs.

seldridge (approved):
LGTM 

Some nits and one clarification.

PR #40, opened 2022-08-23
clarify semantics aroud zero bit width integers

ekiwi:
This clarifies that all zero-width integers are zero extended which matches the current behavior of the `firrtl` compile.
Furthermore it allows UInt<0>(0) and SInt<0>(0) zero-width integer literal values.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -273,6 +273,36 @@ UInt
 SInt
 ```
 
+#### Zero Bit Width Integers
+
+Integers of width zero are permissable. They are always zero extended.
+Thus, when used in an operation that extends to a positive bit width, they
+behave like a zero. While zero bit width integer carry no information, we
+allow 0-bit integer constant zeros for convenience:
+`UInt<0>(0)` and `SInt<0>(0)`.
+
+``` firrtl
+wire zero_u : UInt<0>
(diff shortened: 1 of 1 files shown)


jackkoenig:
I think this is reasonable. I will admit that I'm not certain that zero-extension is the best behavior for zero-width SInts, as @ekiwi has pointed out to me, this is the existing behavior (https://scastie.scala-lang.org/GC7q16cMSCaARuvfKbZqSQ) so I do think it's reasonable to specify it as it exists today.

darthscsi:
Seems reasonable.

jackkoenig (approved):

jackkoenig:
I verified the behavior in MFC 1.14.0:
```
circuit Example :
  module Example :
    output u1 : UInt<8>
    output s1 : SInt<8>

    wire u0 : UInt<0>
    u0 is invalid
    u1 <= u0

    wire s0 : SInt<0>
    s0 is invalid
    s1 <= s0
```
Generates
```verilog
// Generated by CIRCT sifive/1/14/0
module Example(
  output [7:0] u1,
               s1);

  assign u1 = 8'h0;
  assign s1 = 8'h0;
endmodule
```

PR #42, opened 2022-09-06
[patch] Specify the behavior of "indeterminate values"

darthscsi:
Indeterminate values are underspecified.  This expands the constructs with explict indeterminate value to include register initialization.  This further describes the implementation constraints on how indeterminate values behave.  This is intended to make the current behavior, which changes visible behavior of circuits with indeterminate values, consistent with the spec. 
 Namely, current implementations will replace a register with a constant if the constant is written at some point (say reset), if there are no other writes.  This changes the observable state of the register between power-on and first write.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2767,8 +2767,9 @@ defined with the following constraints.
 * Register initialization is done in a consistent way for all registers.  If 
 code is generated to randomly initialize some registers (or 0 fill them, etc), 
 it should be generated for all registers.
-* All observations of an expression with indeterminate value must see the same 
-value at runtime.  Multiple readers of a value will see the same runtime value.
+* All observations of a unique instance of an expression with indeterminate 
+value must see the same value at runtime.  Multiple readers of a value will see 
+the same runtime value.
 * Indeterminate values are not time-varying.  Time-aware constructs, such as 
 registers, which hold an indeterminate value will return the same runtime value 
 unless something changes the value in a normal way.  For example, an 
--- a/spec.md
+++ b/spec.md
@@ -2781,7 +2781,8 @@ unless something changes the value in a normal way.  For example, an
 uninitialized register will return the same value over multiple clock cycles 
 until it is written (or reset).
 - The value produced at runtime for an expression which produced an intermediate
-value shall only be a function of the inputs of the expression.  For example, an out-of-bounds vector access shall produce the same value for a 
+value shall only be a function of the inputs of the expression.  For example, an
+out-of-bounds vector access shall produce the same value for a 
 given out-of-bounds index and vector contents.
 - Two constructs with indeterminate values place no constraint on the identity 
 of their values.  For example, two uninitialized registers, which therefore 
--- a/spec.md
+++ b/spec.md
@@ -2780,8 +2780,8 @@ registers, which hold an indeterminate value will return the same runtime value
 unless something changes the value in a normal way.  For example, an 
 uninitialized register will return the same value over multiple clock cycles 
 until it is written (or reset).
-- An expression which produces an indeterminate value shall produce the same 
-value for the same input.  For example, an out-of-bounds array access shall 
+- The value produced at runtime for an expression which produced an intermediate
+value shall only be a function of the inputs of the expression.  For example, an out-of-bounds array access shall 
 produce the same value for a given out-of-bounds index and array contents.
 - Two constructs with indeterminate values place no constraint on the identity 
 of their values.  For example, two uninitialized registers, which therefore 
--- a/spec.md
+++ b/spec.md
@@ -984,9 +984,7 @@ module MyModule :
    w.b is invalid
 ```
 
-For the purposes of simulation, invalidated components are initialized to random
-values, and operations involving indeterminate values produce undefined
-behaviour. This is useful for early detection of errors in simulation.
+The handing of invalidated components is covered in [@sec:indeterminate-values].
 
 ### The Invalidate Algorithm
 
@@ -2768,7 +2766,7 @@ value at runtime.  Multiple readers of a value will see the same runtime value.
 registers, which hold an indeterminate value will return the same runtime value 
 unless something changes the value in a normal way.  For example, an 
 uninitialized register will return the same value over multiple clock cycles 
-until it is written.
+until it is written (or reset).
 * Two constructs with indeterminate values place no constraint on the identity 
 of their values.  For example, two uninitialized registers, which therefore 
 contain indeterminate values, do not need to be equal under comparison.
--- a/spec.md
+++ b/spec.md
@@ -899,7 +899,6 @@ The following example demonstrates instantiating a wire with the given name
 wire mywire: UInt
 ```
 
-
 ## Registers
 
 A register is a named stateful circuit component.  Reads from a register return 
@@ -2718,13 +2717,13 @@ the scope of what behavior is observable (i.e., a relaxation of the
 An indeterminate value represents a value which is unknown or unspecified.  
 Indeterminate values are generally implementation defined, with constraints 
 specified below.  An indeterminate value may be assumed to be any specific 
-value, at an implementation's descresion, if, in doing so, all observable 
+value, at an implementation's discresion, if, in doing so, all observable 
 behavior is as if the indeterminate value always took the specific value.
 
 This allows transformations such as the following, where when `a` has an 
 indeterminate value, the implementation chooses to consistently give it a value 
-of 4.  There is no visibility of a when it has an indeterminate value which does 
-not see it as 4.
+of 4.  There is no visibility of `a` when it has an indeterminate value which 
+does not see it as 4.
 ``` firrtl
 module IValue :
   output o : UInt<8>
@@ -2736,7 +2735,7 @@ module IValue :
     a <= UInt<3>("h4")
   o <= a
 ```
-is transfromed to:
+is transformed to:
 ``` firrtl
 module IValue :
   output o : UInt<8>
@@ -2765,7 +2764,7 @@ to randomly initialize some registers (or 0 fill them, etc), it should be
 generated for all registers.
 * All observations of an expression with indeterminate value must see the same 
 value at runtime.  Multiple readers of a value will see the same runtime value.
-* Indeterminate values are not time-varying.  Time-aware construct, such as 
+* Indeterminate values are not time-varying.  Time-aware constructs, such as 
 registers, which hold an indeterminate value will return the same runtime value 
 unless something changes the value in a normal way.  For example, an 
 uninitialized register will return the same value over multiple clock cycles 
--- a/spec.md
+++ b/spec.md
@@ -2718,22 +2718,26 @@ the scope of what behavior is observable (i.e., a relaxation of the
 An indeterminate value represents a value which is unknown or unspecified.  
 Indeterminate values are generally implementation defined, with constraints 
 specified below.  An indeterminate value may be assumed to be any specific 
-value, at an implementation's discretion, if, in doing so, all observable 
-behavior is as if the indeterminate value always took the specific value.
+value (not necessarily literal), at an implementation's discretion, if, in doing
+so, all observable behavior is as if the indeterminate value always took the 
+specific value.
 
 This allows transformations such as the following, where when `a` has an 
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -2709,6 +2713,67 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
...
+  o <= a
+```
+
+The behavior of constructs which cause indeterminate values is implementation 
+defined with the following constraints.  
+* Registers initialization is of a uniform code pattern.  If code is generated 
+to randomly initialize some registers (or 0 fill them, etc), it should be 
+generated for all registers.
+* All observations of an expression with indeterminate value must see the same 
+value at runtime.  Multiple readers of a value will see the same runtime value.
```
dtzSiFive, line 2767: Does the "all observations" need to hold true across transforms that may create a copy or replace the construct/wire/reg?  As written I would tentatively say it does since this is the source-level semantics and must be true regardless of how the construct gets lowered.  Primarily thinking of inlining as where this would come up, although elsewhere too.  Similarly, I think this may cause different behavior  in a circuit before/after dedup if different values were used in each or possibly before/after lowering an aggregate register (to registers for each ground type).

Mostly lets not cause ourselves the need to take measures like assign these values early and track them throughout + across optimizations.

Standardizing on '0' does avoid this issue :upside_down_face: .

spec.md
```diff
@@ -2709,6 +2713,67 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
...
+* Indeterminate values are not time-varying.  Time-aware construct, such as
```
dtzSiFive, line 2768:
```suggestion
* Indeterminate values are not time-varying.  Time-aware constructs, such as 
```

spec.md
```diff
@@ -2709,6 +2713,67 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
...
+* Registers initialization is of a uniform code pattern.  If code is generated 
+to randomly initialize some registers (or 0 fill them, etc), it should be 
+generated for all registers.
+* All observations of an expression with indeterminate value must see the same 
+value at runtime.  Multiple readers of a value will see the same runtime value.
+* Indeterminate values are not time-varying.  Time-aware construct, such as 
+registers, which hold an indeterminate value will return the same runtime value 
+unless something changes the value in a normal way.  For example, an 
+uninitialized register will return the same value over multiple clock cycles 
+until it is written.
```
dtzSiFive, line 2772: ("reset" is covered by "is written", yes?)

spec.md
```diff
@@ -2709,6 +2713,67 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
...
+* Two constructs with indeterminate values place no constraint on the identity 
+of their values.  For example, two uninitialized registers, which therefore 
+contain indeterminate values, do not need to be equal under comparison.
+
```
dtzSiFive, line 2776: Elsewhere there's language saying in simulation operations involving indeterminate values produce undefined behavior? While not incompatible with this text (described behavior here for non-simulation, undef behavior for simulation) if that's the intended semantics I think it would help to maybe unify or note the other behavior to avoid confusion?

spec.md
```diff
@@ -2709,6 +2713,67 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
...
+  when c :
+    a <= UInt<3>("h4")
+   else :
+     a <= UInt<3>("h42")
+  o <= a
+```
+
+The behavior of constructs which cause indeterminate values is implementation 
+defined with the following constraints.  
+* Registers initialization is of a uniform code pattern.  If code is generated
```
dtzSiFive, line 2763: Not-even-a-nit: "Registers initialization" seems little strange phrasing-wise.

Oh I see, uniform code (but not values) for "initialize a register"-- either none or the same code "pattern"-- all or none? Just curious, is having same "code pattern" significant?

spec.md
```diff
@@ -2709,6 +2713,67 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
...
+
+An indeterminate value represents a value which is unknown or unspecified.  
+Indeterminate values are generally implementation defined, with constraints 
+specified below.  An indeterminate value may be assumed to be any specific 
+value, at an implementation's descresion, if, in doing so, all observable 
+behavior is as if the indeterminate value always took the specific value.
+
+This allows transformations such as the following, where when `a` has an 
+indeterminate value, the implementation chooses to consistently give it a value 
+of 4.  There is no visibility of a when it has an indeterminate value which does
```
dtzSiFive, line 2726:
```suggestion
of 4.  There is no visibility of `a` when it has an indeterminate value which does 
```

spec.md
```diff
@@ -899,6 +899,7 @@ The following example demonstrates instantiating a wire with the given name
 wire mywire: UInt
 ```
 
+
```
dtzSiFive, line 902: Stray whitespace change (and not consistently using two spaces before these headers, see `## Invalidates` below).

dtzSiFive (commented):
> This is intended to make the current behavior, which changes visible behavior of circuits with indeterminate values

Mind elaborating just a touch? When is behavior visibly changed and perhaps in what way, or does this refer to this PR enshrining assigning specific values when convenient ("changing" visible behavior).

Added some comments.  Clarifying these things is great, thanks!

spec.md
darthscsi: Yes, meant to erase that as it is clearly bogus and unworkable (every reachable operation from a reset-less register would have undefined behavior, which is clearly not true).

spec.md
darthscsi: Trying to get out of the "do whatever for each" to "do the same (implementation-defined) thing for all"

spec.md
darthscsi: The intent is certainly observations of runtime instances not input syntactic instances.
Maybe saying observations of unique runtime instances would help.

spec.md
```diff
@@ -2709,6 +2713,67 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
 the scope of what behavior is observable (i.e., a relaxation of the
 ["as-if"](https://en.wikipedia.org/wiki/As-if_rule) rule).
 
+## Indeterminate Values
+
+An indeterminate value represents a value which is unknown or unspecified.  
+Indeterminate values are generally implementation defined, with constraints 
+specified below.  An indeterminate value may be assumed to be any specific 
+value, at an implementation's discresion, if, in doing so, all observable 
+behavior is as if the indeterminate value always took the specific value.
```
seldridge, line 2722: Just to clarify, you are stating that a "value" may not necessarily be a literal, yes?

It may be useful to change the example below to clearly indicate this. I assuming that we want to also allow optimization of:

```
wire a: UInt<8>
a is invalid
when c :
  a <= b
```

To:
```
a <= b
```

It would probably be better to change the example below to not use `h4` and instead use a non-literal value.

If we are choosing to NOT allow this, then the SFC (and MFC) are in violation of the spec.

spec.md
```diff
@@ -2709,6 +2713,67 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
 the scope of what behavior is observable (i.e., a relaxation of the
 ["as-if"](https://en.wikipedia.org/wiki/As-if_rule) rule).
 
+## Indeterminate Values
+
+An indeterminate value represents a value which is unknown or unspecified.  
+Indeterminate values are generally implementation defined, with constraints 
+specified below.  An indeterminate value may be assumed to be any specific 
+value, at an implementation's discresion, if, in doing so, all observable
```
dtzSiFive, line 2721: Nit:
```suggestion
value, at an implementation's discretion, if, in doing so, all observable 
```

spec.md
seldridge: Would it help to lean on a new IR operation, like `freeze`, that would help clarify at what point these guarantees kick in as opposed to the somewhat more amorphous "expression with indeterminate value".

E.g.:

```
wire a: UInt<1>
a is invalid

node b = freeze(a)
node c = freeze(a)

assert(b == b) ; this is always true
assert(b == c) ; this is always false?
```

mwachs5:
I think this writeup is helpful, thanks!

spec.md
```diff
@@ -2709,6 +2713,67 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
...
+   else :
+     a <= UInt<3>("h42")
+  o <= a
+```
+
+The behavior of constructs which cause indeterminate values is implementation 
+defined with the following constraints.  
+* Register initialization is done in a consistent way for all registers.  If 
+code is generated to randomly initialize some registers (or 0 fill them, etc), 
+it should be generated for all registers.
```
ekiwi, line 2765: Why is that a constraint? What if someone (for verification purposes) wants to randomize only some registers. Since their startup value is indeterminate, any concrete assignment should be valid.

spec.md
ekiwi: Shouldn't the second sentence be:
> Multiple readers of _the same expression_ will see the same runtime value.
?

I am also wondering if this would benefit from a clarification about the temporal aspect. I.e., the:
> Multiple readers of the same expression, _in the same time step_ will see the same runtime value.

mwachs5 (commented):
is "clarification" patch (I think so). Can you update the revision log?

spec.md
darthscsi: In this case, an implementation would provide a way to initialize some registers, but not others.  But that mechanism would be applied uniformly.

spec.md
```diff
@@ -2714,6 +2718,75 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
...
+value must see the same value at runtime.  Multiple readers of a value will see 
+the same runtime value.
+* Indeterminate values are not time-varying.  Time-aware constructs, such as 
+registers, which hold an indeterminate value will return the same runtime value 
+unless something changes the value in a normal way.  For example, an 
+uninitialized register will return the same value over multiple clock cycles 
+until it is written (or reset).
+* An expression which produces an indeterminate value shall produce the same 
+value for the same input.  For example, an out-of-bounds array access shall 
+produce the same value for a given out-of-bounds index and array contents.
```
seldridge, line 2785: The second sentence can be confusing. Can you change this to narrowly talk about out-of-bounds accesses?

"E.g., an out-of-bounds array access will return the same value for the same out-of-bounds index for all time."

I don't think it is useful to combine both out-of-bounds and in-bounds in the same example.

spec.md
```diff
@@ -2714,6 +2718,75 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
...
+  when c :
+    a <= v
+   else :
+     a <= UInt<3>("h42")
+  o <= a
+```
+
+The behavior of constructs which cause indeterminate values is implementation 
+defined with the following constraints.  
+* Register initialization is done in a consistent way for all registers.  If
```
seldridge, line 2772: Mega-nit: the rest of the doc uses `-`, not `*` for itemized lists, and indents the rest of the paragraph to match the start of the item.

seldridge (approved):
LGTM. This has been a huge missing piece of the spec for a long time. Thanks for tackling this thorny part of FIRRTL.

Nitty comments/clarification requests throughout.

spec.md
darthscsi: I don't know that that rewording is helpful.  The intention is that the produced behavior is deterministic for it's input.

ekiwi: By `array`, do you mean a `mem` or a `Vec` or both?

spec.md
```diff
@@ -2714,6 +2718,75 @@ need only respect the 2-state behavior of a circuit.  This is a limitation on
...
+value must see the same value at runtime.  Multiple readers of a value will see 
+the same runtime value.
+- Indeterminate values are not time-varying.  Time-aware constructs, such as 
+registers, which hold an indeterminate value will return the same runtime value 
+unless something changes the value in a normal way.  For example, an 
+uninitialized register will return the same value over multiple clock cycles 
+until it is written (or reset).
+- The value produced at runtime for an expression which produced an intermediate
+value shall only be a function of the inputs of the expression.  For example, an out-of-bounds vector access shall produce the same value for a 
+given out-of-bounds index and vector contents.
```
seldridge, line 2785: 80-char. 👊

seldridge (approved):
LGTM

PR #43, opened 2022-09-14
[major] Intrinsics and Functions

darthscsi:
Define functions, invocation, and intrinsics.  Intrinsics provide a way to add simple operations without disturbing the spec (for example, specific verilog cover forms or operations mapping to specific vendor primitives).  These also provide a basis for generalized vector operations in the future.

Still a work in progress.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -901,7 +901,12 @@ wire mywire: UInt
 
 ## Registers
 
-A register is a named stateful circuit component.
+A register is a named stateful circuit component.  Reads from a register return 
+the current value of the element, writes are not visible until after a positive 
+edges of the register's clock port.
+
+The clock signal for a register must be of type `Clock`{.firrtl}.  The type of a 
+register must be a passive type (see [@sec:passive-types]).
 
 The following example demonstrates instantiating a register with the given name
 `myreg`{.firrtl}, type `SInt`{.firrtl}, and is driven by the clock signal
@@ -913,10 +918,17 @@ reg myreg: SInt, myclock
 ; ...
 ```
 
-Optionally, for the purposes of circuit initialization, a register can be
-declared with a reset signal and value. In the following example,
-`myreg`{.firrtl} is assigned the value `myinit`{.firrtl} when the signal
-`myreset`{.firrtl} is high.
+A register may be declared with a reset signal and value.  The register's value 
+is updated with the reset value when the reset is asserted.  The reset signal 
+must be a `Reset`{.firrtl}, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}, and 
+the type of initialization value must be equivalent to the declared type of the 
+register (see [@sec:type-equivalence] for details).  The behavior of the 
+register depends on the type of the reset signal.  `AsyncReset`.{firrtl} will 
+immediately change the value of the register.  `UInt<1> will not change 
+the value of the register until the next positive edge of the clock signal (see 
+[@sec:reset-type]).  `Reset`.{firrtl} is an abstract reset whose behavior 
+depends on reset inference.  In the following example, `myreg`{.firrtl} is 
+assigned the value `myinit`{.firrtl} when the signal `myreset`{.firrtl} is high.  
 
 ``` firrtl
 wire myclock: Clock
@@ -925,11 +937,6 @@ wire myinit: SInt
 reg myreg: SInt, myclock with: (reset => (myreset, myinit))
 ; ...
 ```
-
-Note that the clock signal for a register must be of type `clock`{.firrtl},
-the reset signal must be a `Reset`{.firrtl}, `UInt<1>`{.firrtl}, or
-`AsyncReset`{.firrtl}, and the type of initialization value must be equivalent
-to the declared type of the register (see [@sec:type-equivalence] for details).
 
 ## Invalidates
 
--- a/spec.md
+++ b/spec.md
@@ -175,7 +175,7 @@ FIRRTL version 1.1.0
 circuit MyTop...
 ```
 
-# Circuits and Modules and Functions
+# Circuits, Modules, and Functions
 
 ## Circuits
 
--- a/spec.md
+++ b/spec.md
@@ -175,7 +175,7 @@ FIRRTL version 1.1.0
 circuit MyTop...
 ```
 
-# Circuits and Modules
+# Circuits and Modules and Functions
 
 ## Circuits
 
@@ -243,6 +243,58 @@ ports.
 A common use of an externally defined module is to represent a Verilog module
 that will be written separately and provided together with FIRRTL-generated
 Verilog to downstream tools.
+
+## Functions
+
+A function has a name, list of arguments, and a list of statements representing
+a data-flow graph of stateless circuit components.  Each argument is specified 
+by its direction, which may be input or output, a name, and the data type of the 
+argument.
+
+Functions participate in width-inference (see [@sec:width-inference]) at 
+invocation sites in the same way modules to at instantiation sites.
+
+Functions cannot contain memories, registers, or instances.  This makes them
+"stateless"
+
+Functions can optional specify that they are "pure" functions.  These do not
+have side-effecting operations (such as printf).
+
+
+``` firrtl
+function MyPureFunction pure :
+  input foo : UInt
+  output bar : UInt
+  bar <= foo
+
+function MyFunction :
+  input clk: Clock
+  input halt: UInt<1>
+  stop(clk, halt, 42) : optional_name
+```
+
+## Intrinsics
+
+An intrinsic is a function. Intrinsics have all the components of a function
+except  a list of statements.  An intrinsic is a firrtl implementation provided 
+function.  The name of intrinsics are implementation defined.  A firrtl 
+implementation shall reject IR with unknown intrinsics.  Being functions,
+intrinsics are stateless.
+
+The widths of all intrinsic ports must be specified.  Width inference, described 
+in [@sec:width-inference], is not supported.
+
+Intrinsics can be "pure".  These do not have side-effects.  Whether a particular 
+intrinsic is pure or not is implementation defined and implementations shall 
+check that intrinsics declarations have the expected arguments, types, and 
+purity.
+
+``` firrtl
+intrinsic synopsys.4mux.u8 pure :
+  input cond : UInt<2>
+  output bar : Vector<4, UInt<8>>
+```
+
 
 # Types
 
@@ -1530,6 +1582,22 @@ module without affecting the semantics of the circuit.
 To disallow infinitely recursive hardware, modules cannot contain instances of
 itself, either directly, or indirectly through instances of other modules it
 instantiates.
+
+## Invoke
+
+FIRRTL functions and intrinsics are invoked with the `invoke` statement.
+
+``` firrtl
+circuit Top :
+   function MyFunc : 
+     input a : UInt
+     output b : UInt
+   module MyModule :
+      input a: UInt
+      output b: UInt
+      b <= invoke MyFunc(a)
+```
+
 
 ## Stops
 
--- a/spec.md
+++ b/spec.md
@@ -246,10 +246,12 @@ Verilog to downstream tools.
 
 ## Functions
 
-A function has a name, a return type, a list of arguments, and a list of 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -175,7 +175,7 @@ FIRRTL version 1.1.0
 circuit MyTop...
 ```
 
-# Circuits and Modules
+# Circuits and Modules and Functions
```
mwachs5, line 178:
```suggestion
# Circuits, Modules, and Functions
```

spec.md
```diff
@@ -244,6 +244,58 @@ A common use of an externally defined module is to represent a Verilog module
...
+A function has a name, list of arguments, and a list of statements representing
+a data-flow graph of stateless circuit components.  Each argument is specified 
+by its direction, which may be input or output, a name, and the data type of the 
+argument.
+
+Functions participate in width-inference (see [@sec:width-inference]) at 
+invocation sites in the same way modules to at instantiation sites.
+
+Functions cannot contain memories, registers, or instances.  This makes them
+"stateless"
```
mwachs5, line 258:
```suggestion
"stateless".
```

spec.md
```diff
@@ -244,6 +244,58 @@ A common use of an externally defined module is to represent a Verilog module
...
+
+Functions can optional specify that they are "pure" functions.  These do not
```
mwachs5, line 260:
```suggestion
Functions can optionally specify that they are "pure" functions.  These do not
```

spec.md
```diff
@@ -244,6 +244,58 @@ A common use of an externally defined module is to represent a Verilog module
...
+function MyFunction :
+  input clk: Clock
+  input halt: UInt<1>
+  stop(clk, halt, 42) : optional_name
+```
+
+## Intrinsics
+
+An intrinsic is a function. Intrinsics have all the components of a function
+except  a list of statements.  An intrinsic is a firrtl implementation provided
```
mwachs5, line 279: what is "a firrtl implementation". You mean "a firrtl compiler implementation"? Is this a term that is defined already in this document?

spec.md
```diff
@@ -244,6 +244,58 @@ A common use of an externally defined module is to represent a Verilog module
...
+Intrinsics can be "pure".  These do not have side-effects.  Whether a particular 
+intrinsic is pure or not is implementation defined and implementations shall 
+check that intrinsics declarations have the expected arguments, types, and 
+purity.
+
+``` firrtl
+intrinsic synopsys.4mux.u8 pure :
+  input cond : UInt<2>
+  output bar : Vector<4, UInt<8>>
+```
```
fabianschuiki, line 296: Are there use cases for intrinsics with module semantics? Something like memories, certain forms of registers, etc.. It seems like the intrinsic above could cover that, but I'm not entirely sure. E.g., what happens if someone puts a `reg` intside a `function`?

spec.md
```diff
@@ -244,6 +244,58 @@ A common use of an externally defined module is to represent a Verilog module
...
+The widths of all intrinsic ports must be specified.  Width inference, described 
+in [@sec:width-inference], is not supported.
+
+Intrinsics can be "pure".  These do not have side-effects.  Whether a particular 
+intrinsic is pure or not is implementation defined and implementations shall 
+check that intrinsics declarations have the expected arguments, types, and 
+purity.
+
+``` firrtl
+intrinsic synopsys.4mux.u8 pure :
```
mwachs5, line 293: does a .fir file need to include this `intrinsic` in it if it plans to use it?

spec.md
```diff
@@ -1531,6 +1583,22 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
...
+
+``` firrtl
+circuit Top :
+   function MyFunc : 
+     input a : UInt
+     output b : UInt
+   module MyModule :
+      input a: UInt
+      output b: UInt
+      b <= invoke MyFunc(a)
```
mwachs5, line 1598: what is the order of inputs to `MyFunc`? Should these be passed/connected-to by name?

spec.md
```diff
@@ -244,6 +244,58 @@ A common use of an externally defined module is to represent a Verilog module
 that will be written separately and provided together with FIRRTL-generated
 Verilog to downstream tools.
 
+## Functions
+
+A function has a name, list of arguments, and a list of statements representing
+a data-flow graph of stateless circuit components.  Each argument is specified 
+by its direction, which may be input or output, a name, and the data type of the 
+argument.
```
mwachs5, line 252: can you have multiple outputs?

spec.md
fabianschuiki: I think that's the idea. This way the FIR file declares what it expects the intrinsic to be in terms of type signature and name, so the FIR file remains self-contained. And the compiler would just verify that the expected type signature matches what the actual internal intrinsic wants.

spec.md
mwachs5: and if you have multiple outputs what does that look like on the LHS?

spec.md
```diff
@@ -1531,6 +1583,22 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
...
 
+## Invoke
+
+FIRRTL functions and intrinsics are invoked with the `invoke` statement.
+
+``` firrtl
+circuit Top :
+   function MyFunc : 
+     input a : UInt
+     output b : UInt
```
mwachs5, line 1594: are you allowed to have a `function` without assigning all the `output`s?

spec.md
mwachs5: in that sense is an `intrinsic` the same as an `external function`? Should we use the word `extfunction` to parallel the term `extmodule`?

spec.md
mwachs5: Isn't an intrinsic with module semantics an `extmodule`

spec.md
fabianschuiki: I'd expect there to be a difference. The `intrinsic` is something the compiler understands and will process/expand during compilation, so it doesn't make its way to the output. An `extfunction` I'd expect to be transported all the way through and to be emitted for example as a DPI call in the SV output.

spec.md
fabianschuiki: Yeah I guess you could do the whole thing as `extmodule`/`extfunction`. That's what LLVM does. Intrinsics are just external function declarations with a specific name. We could do `extmodule circt.intrinsic.synopsys.4mux.u8` instead. The good thing about an actual intrinsic construct is that you convey to the compiler that this thing _has_ to resolve to something the compiler internally knows, and the intrinsic doesn't accidentally show up in the SV output because the compiler didn't realize you wanted an intrinsic.

spec.md
mwachs5: why do we need a special word for that, and only for functions? Maybe we want the same thing for `extmodule`s as well.

I guess I'm generally confused why we are treating "function" which is basically "module with no sequential logic in it" so differently/non-parallel.

spec.md
mwachs5: sure... i think i get the idea/value of intrinsic, but maybe you could have intrinsic as a modifier of `extfunction` or `extmodule` to be more parallel

mwachs5: > and the intrinsic doesn't accidentally show up in the SV output

I think we *do* want the intrinsic in the SV output, don't we? Otherwise we really are just re-defining `extmodule`.

dtzSiFive: I'll let @darthscsi answer authoritatively but I'd expect this to be like LLVM intrinsics: https://llvm.org/docs/ExtendingLLVM.html#intrinsic-function .

Which is to say:
* intrinsics do not participate in linking and cannot be resolved to definitions in other files.  At FIRRTL level they're only ever declarations.  May be convenient to reject re-using their names elsewhere, or when linking (?).
* intrinsics defined entirely by the compiler (possibly emitting references to definitions linked in from support library (compiler-rt/libgcc, etc.) but doubt that's where we're going AFAIK)

fabianschuiki: > I think we do want the intrinsic in the SV output, don't we?

We want the compiler to expand the intrinsic and do whatever it wants with it, and then the result of that should show up. But the intrinsic wont show up in SV as a module/function to represent the intrinsic. For example, that `synopsys.4mux.u8` you'd expect to expand to the actual multiplexers and necessary vendor pragmas to ensure they get inferred as 4-way muxes.

spec.md
dtzSiFive: I vote that these are different.  External functions might be useful, `extfunction` makes sense to me for that, but I think should be separate from intrinsics (IMHO).  Functions are written by users, intrinsics are by compiler.

Are functions kinda like macros now? And/or do they lower to something suitable in verilog?

Presently we're able to compile a FIRRTL file w/`extmodule`'s to verilog in a reasonable way, not sure what we'd do for `extfunction`s (but maybe this is straightforward?).   Bringing it up so we're don't accidentally cross over the line into "you have to link this before you can produce verilog".

ekiwi:
I am curious: what is the rational for user defined functions and intrinsics? To me it seems like all examples presented so far could be handled with a `module` or  `extmodule`.
Is there a Chisel integration that I could look at to better understand the use cases?

spec.md
uenoku, line 293: If we allow vendor specific intrinsic, what happens if we do target other vendors? Do we want to introduce target-triple for FIRRTL and CIRCT land?

spec.md
darthscsi: The reason to make a distinction between modules and functions are that functions are stateless, which provides significant analysis power when dealing with them.  They are guaranteed to be data-flow like and behave as operations, not modules (and as a bonus, don't have to have connect semantics).

There is a use for module-like intrinsics.  Arguably, memories are already module-like intrinsics. Adding generic support for this is outside the scope of this change, but you could imagine an intmodule parallel to intrinsic.

module = connect semantics, instance semantics, stateful
extmodule = module with missing body (and some other restrictions)
intmodule = compiler/implementation provided module (to be added when needed)
Parallels
function = dataflow semantics, op-like semantics, stateless (but may be side-effecting)
extfunction = function with missing body
intrinsic = function/implemenation provided function

spec.md
darthscsi: extfunction is user-supplied, intrinsics are implementation provided and known.  They are just extfunctions, but they are ones the compiler directly knows.

darthscsi: What happens if you use an arm neon intrinsic when compiler C to amd64?

fabianschuiki: We should probably also provide a default lowering to standard Comb/Seq ops for the intrinsics where that makes sense. For example, a dedicated mux4 intrinsic should provide a lowering to a `comb.mux` for all those tools and passes that don't need the added detail of the intrinsic. But agreed, some will be vendor/target-specific and require the frontend to just not emit them.

mwachs5: this was answered in this comment: https://github.com/chipsalliance/firrtl-spec/pull/43#discussion_r975619940

spec.md
```diff
@@ -244,6 +244,80 @@ A common use of an externally defined module is to represent a Verilog module
...
+
+A function has a name, a return type, a list of arguments, and a list of 
+statements representing a data-flow graph of stateless circuit components.  Each
+argument specifies its name and  data type.  Functions have a single return 
+value.  The name of the function serves as the identifier for the return value.
+
+Functions participate in width-inference (see [@sec:width-inference]) at 
+invocation sites in the same way modules do at instantiation sites.
+
+Functions cannot contain memories, registers, non-local anotations, or
```
mwachs5, line 257:
```suggestion
Functions cannot contain memories, registers, non-local annotations, or 
```

spec.md
```diff
@@ -1538,6 +1612,22 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
 itself, either directly, or indirectly through instances of other modules it
 instantiates.
 
+## Invoke
+
+FIRRTL functions and intrinsics are invoked with the `invoke` statement.
```
mwachs5, line 1617:
```suggestion
FIRRTL functions, extfunctions and intrinsics are invoked with the `invoke` statement.
```

not sure if you want to call that out explicitly, since both extfunctions and functions are clearly stated to be functions

mwachs5, line 1617:
```suggestion
FIRRTL functions and intrinsics are invoked with the `invoke` statement.

The invoke statement assigns the arguments in order that they are defined in the function's `input` list.
```

darthscsi:
Closing as intrinsics are in and functions can wait.

PR #44, opened 2022-09-19
[patch] Add Explicit Info About Aggregate Types

seldridge:
Add clarifications to the spec that FIRRTL has both ground types and aggregate types.  The latter term "aggregate types" is used, but not defined.  Also, restructure the spec to include an explicit "aggregate types" section that is at the same level as "ground types"  Move vector and bundle types under this new "aggregate types" section.

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -5,6 +5,8 @@ revisionHistory:
   # additions to the specification should append entries here.
   thisVersion:
     - Specify behavior of zero bit width integers, add zero-width literals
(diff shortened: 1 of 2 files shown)


mwachs5:
I think this is very helpful. 

nit: please update the Revision History section before merging

seldridge:
> nit: please update the Revision History section before merging

🤦 Thanks for the reminder!

mwachs5 (approved):

PR #46, opened 2022-09-19
[major] Move Extmodule Params to Instantiation

seldridge:
This changes the format for FIRRTL external modules to move parameter values to the site of instantiation as opposed to being declared along with the extmodule definition.  This has the added benefit of removing the "defname" as this is now unified with the name of the external module.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -217,12 +217,9 @@ instance statement for details on how to instantiate a module
 ## Externally Defined Modules
 
 Externally defined modules are modules whose implementation is not provided in
-the current circuit.  Only the ports and name of the externally defined module
-are specified in the circuit.  An externally defined module may include, in
-order, an optional _defname_ which sets the name of the external module in the
-resulting Verilog and zero or more name--value _parameter_ statements.  Each
-name--value parameter statement will result in a value being passed to the named
-parameter in the resulting Verilog.
+the current circuit.  Only the ports of the externally defined module are
+specified in the circuit.  No other statements may exist in the external module
+body.
 
 An example of an externally defined module is:
 
@@ -231,14 +228,30 @@ extmodule MyExternalModule :
   input foo: UInt<2>
   output bar: UInt<4>
   output baz: SInt<8>
-  defname = VerilogName
-  parameter x = "hello"
-  parameter y = 42
 ```
 
 The widths of all externally defined module ports must be specified.  Width
 inference, described in [@sec:width-inference], is not supported for module
 ports.
+
+Externally defined modules may have zero or more parameters.  Parameters may be
+of known-width `UInt`{.firrtl} or `String`{.firrtl}.  The value of a parameter
+is set at each instantiation of an external module using a literal value.
+
+An example of a parametric externally defined module and its instantiation is:
+
+``` firrtl
+extmodule MyExternalModule2<parameter x: String, parameter y: UInt<8>> :
+  ; ...
+module Top:
+  inst foo of MyExternalModule2<"hello", UInt<8>(42)>
+  inst bar of MyExternalModule2<"world", UInt(0)>
+```
+
+As shown above, it is allowable to use a smaller, unknown width `UInt`{.firrtl}
+literal to set a parameter value so long as the width of the underlying
+parameter value is large enough to store the literal type.  A literal that is
+too large for a given parameter type is illegal.
 
 A common use of an externally defined module is to represent a Verilog module
 that will be written separately and provided together with FIRRTL-generated


seldridge:
This is only a sliver of where I and others (@darthscsi 👀 ) want to take parameters. However, this is the bare minimum cleanup that fixes a lot of weirdness around external modules. This also matches where I want to take the internal MFC representation of external modules.

spec.md
```diff
@@ -231,15 +228,36 @@ extmodule MyExternalModule :
...
-  parameter y = 42
 ```
 
 The widths of all externally defined module ports must be specified.  Width
 inference, described in [@sec:width-inference], is not supported for module
 ports.
 
+Externally defined modules may have zero or more parameters.  Parameters may be
+of known-width `UInt`{.firrtl} or `SInt`{.firrtl} types or `String`{.firrtl}
+type.  The value of a parameter is set at each instantiation of an external
```
mwachs5, line 239: Should we mention that they must be passed in declaration order (there is no name-specified parameter passing syntax)

spec.md
```diff
@@ -231,15 +228,36 @@ extmodule MyExternalModule :
...
+
+An example of a parametric externally defined module and its instantiation is:
+
+``` firrtl
+extmodule MyExternalModule2<
+  parameter x: String,
+  parameter y: UInt<8>,
+  parameter z: SInt<4>
+> :
+  ; ...
```
mwachs5, line 250: Not sure what this semicolon is doing here? It's a little confusing. Can we just show the full module with some ports?

Also is the expectation that those parameters could define the port width (this seems like a common use case), can you show that usage?

mwachs5 (commented):
It seems like we need to support / demonstrate parameters in UInt widths at least to make this change?

darthscsi:
> It seems like we need to support / demonstrate parameters in UInt widths at least to make this change?

Yes, the extmodules are currently expressing the result of their parameterization, which needs to be fixed before we can move the parameters to the instances.

seldridge:
> It seems like we need to support / demonstrate parameters in UInt widths at least to make this change?

Yes. I totally missed this. Currently you _can_ do this with multiple extmodules with the same defname, different parameters, and different widths. This is going to bloat this PR a bit as it gets into parameter expression operations. More to follow...

The most basic example of this is we need to deal with something like:

```
extmodule Bar1:
  input a: UInt<4>
  defname Bar
  param width = 4
extmodule Bar2:
  input a: UInt<8>
  defname Bar
  param width = 8
```

This needs to be represented as something like:

```
extmodule Bar<parameter width: UInt<32>>:
  input a: UInt<width>
```

seldridge:
I updated this with a couple of changes:

1. Parameters are now boxed with a `Param<...>` indicator. This is done to sequester parameters into their own area.
2. To solve the issue highlighted in the above comment (https://github.com/chipsalliance/firrtl-spec/pull/46#issuecomment-1252677177) as brought up by Megan and Lenharth, I've added an incomplete section on parameter primitive operations. This only shows `paramAdd` and would need to include more operations to be actually usable. FIRRTL does not normally create new ops for new types. It would be more consistent with the existing FIRRTL spec to define these on existing primitive operations.

There issues that need to be ironed out related to parameter directionality. E.g., are there input and output parameters?

As it came up in offline discussions, this change has impacts on how deduplication works. However, I don't expect it to change any deduplication behavior. Examples follow:

### Example 1

In the following circuit, `Bar` and `Baz` will not deduplicate:

```
circuit Top1:
  extmodule Foo<parameter x: UInt<8>> :
    output a: UInt<1>

  module Bar:
    output a: UInt<1>

    inst foo of Foo<4>
    a <= foo.a

  module Baz:
    output a: UInt<1>

    inst foo of Foo<8>
    a <= foo.a

  module Top1:
    output a: UInt<1>

    inst bar of Bar
    inst baz of Baz
    a <= xor(bar.a, baz.a)
```

This is equivalent to the current situation now. The difference is that `inst foo of Foo_4`/`inst foo of Foo_8` are replaced with `inst foo of Foo<4>`/`inst foo of Foo<8>`:

```
circuit Top1:
  extmodule Foo_4:
    output a: UInt<1>
    defname = Foo
    parameter x = 4

  extmodule Foo_8:
    output a: UInt<1>
    defname = Foo
    parameter x = 8

  module Bar:
    output a: UInt<1>

    inst foo of Foo_4
    a <= foo.a

  module Baz:
    output a: UInt<1>

    inst foo of Foo_8
    a <= foo.a

  module Top1:
    output a: UInt<1>

    inst bar of Bar
    inst baz of Baz
    a <= xor(bar.a, baz.a)
```

### Example 2

If, on the other hand, the parameters passed to the external module instance are the same, then `Bar` and `Baz` will deduplicate:

```
circuit Top2:
  extmodule Foo<parameter x: UInt<8>> :
    output a: UInt<1>

  module Bar:
    output a: UInt<1>

    inst foo of Foo<4>
    a <= foo.a

  module Baz:
    output a: UInt<1>

    inst foo of Foo<4>
    a <= foo.a

  module Top2:
    output a: UInt<1>

    inst bar of Bar
    inst baz of Baz
    a <= xor(bar.a, baz.a)
```

This is equivalent to the current situation where an external module is instantiated twice:

```
circuit Top2:
  extmodule Foo_4:
    output a: UInt<1>
    defname = Foo
    parameter x = 4

  module Bar:
    output a: UInt<1>

    inst foo of Foo_4
    a <= foo.a

  module Baz:
    output a: UInt<1>

    inst foo of Foo_4
    a <= foo.a

  module Top2:
    output a: UInt<1>

    inst bar of Bar
    inst baz of Baz
    a <= xor(bar.a, baz.a)
```

Or where the external module is duplicated and should deduplicate:

```
circuit Top2:
  extmodule Foo_4_0:
    output a: UInt<1>
    defname = Foo
    parameter x = 4

  extmodule Foo_4_1:
    output a: UInt<1>
    defname = Foo
    parameter x = 4

  module Bar:
    output a: UInt<1>

    inst foo of Foo_4_0
    a <= foo.a

  module Baz:
    output a: UInt<1>

    inst foo of Foo_4_1
    a <= foo.a

  module Top2:
    output a: UInt<1>

    inst bar of Bar
    inst baz of Baz
    a <= xor(bar.a, baz.a)
```

### Example 3

There will need to be canonicalization/folding of parameters in order to cause situations like the following to properly recognize that two instances are the same. In the worst case, this requires walking backwards across expressions until you hit a root `Param` literal (or a root `Param` in a world where modules are also parameterized). In practice, this is handled automatically with canonicalization.

```
circuit Top3:
  extmodule Foo<parameter x: UInt<8>> :
    input a: UInt<1>

  module Bar:

    node a = Param(UInt<7>(2))
    node b = Param(UInt<7>(2))
    inst foo of Foo<paramAdd(a, b)>

  module Baz:

    node a = Param(UInt<7>(1))
    node b = Param(UInt<7>(3))
    inst foo of Foo<paramAdd(a, b)>

  module Top3:

    inst bar of Bar
    inst baz of Baz
```

spec.md
```diff
@@ -563,6 +585,37 @@ flipped orientations. Thus all ground types are passive types. Vector types are
 passive if their element type is passive. And bundle types are passive if no
 fields are flipped and if all field types are passive.
 
+## Parameter Types
+
+A `UInt`{.firrtl}, `SInt`{.firrtl}, or `String`{.firrtl} type may be boxed with
+a `Param`{.firrtl}` type.  This indicates that this is a type that is used to
+express parameterization.  Parameter types may be used to express
+parameterization on externally defined modules.  Parameter types may be used as
+nodes.  Parameter types may be used in passive aggregates.  All other uses of
```
mwachs5, line 594: can you show an example of "Parameter types may be used in passive aggregates"? what does that mean?

spec.md
```diff
@@ -2910,6 +2976,17 @@ reference = id
           | reference , "[" , int , "]"
           | reference , "[" , expr , "]" ;
 
+(* Parameter primitive operations *)
+primop_param_2expr_keyword = "add" ;
```
mwachs5, line 2980:
```suggestion
primop_param_2expr_keyword = "paramAdd" ;
```

if I am reading this right?

PR #47, opened 2022-09-21
Clarify time-invariant indeterminate value language.

darthscsi:


seldridge (approved):
LGTM

This is clarifying the register initialization behavior. It is likely useful to explicitly talk about register initialization behavior somewhere else, too.

Please add the appropriate tag before merging. I think this is `[patch]`, but could also be `[major]`. 🤔  (It doesn't honestly matter as this will get wrapped up into a new major release whenever we cut it.)

PR #48, opened 2022-10-12
Allow direct construction of aggregates

darthscsi:
This allows expression of aggregate literals.  Aggregate construction is an expression, so it is not usable on the lhs of a connect.


darthscsi:
I don't think it is necessary to have passive types.  An aggregate construction should work like a wire with all fields connected.  Obviously for a literal aggregate, you would need a passive type for flow to work (a constant can't be a sink).

PR #49, opened 2022-11-28
[patch] Move head and tail to primop_1expr1int_keyword (minor grammar change)

keszocze:
The `head` and `tail`  operations have one argument and one parameter (see Sections 9.24 and 9.25 of the spec v1.1.0) but are listed in the `primop_1expr_keyword` rule of the grammar. This pull request moves these operations to the `primop_1expr1int_keyword` rule.

I also manually added a label to Section 9 "Primitive Operations" (i.e. `{#sec:primitive-operations}`). This section was referenced by its name in other parts of the document. As Section 8.11 has the same name, it was referenced instead of the intended Section 9.

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -9,6 +9,8 @@ revisionHistory:
     - Add an explicit section about "Aggregate Types" and move "Vector Type" and
       "Bundle Type" under it.
     - Remove Fixed Point Types.
+    - Move "head" and "tail" from primop_1expr_keyword to primop_1expr1int_keyword
+      in the "FIRRTL Language Definition".
   # Information about the old versions.  This should be static.
   oldVersions:
     - version: 1.1.0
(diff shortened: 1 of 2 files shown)


revision-history.yaml
```diff
@@ -9,6 +9,8 @@ revisionHistory:
     - Add an explicit section about "Aggregate Types" and move "Vector Type" and
       "Bundle Type" under it.
     - Remove Fixed Point Types.
+    - Move "head" and "tail" from primop_1expr_keyword to primop_1expr1int_keyword
+      in the "FIRRTL Language Definition".
```
seldridge, line 13: Nit: 80-column:

```suggestion
    - Move "head" and "tail" from primop_1expr_keyword to
      primop_1expr1int_keyword in the "FIRRTL Language Definition".
```

seldridge (approved):
Nice! Thanks for taking the time to fix this.

One nit about column length.

keszocze:
Btw, I found some things in the grammar, that I am not sure about. Does it make sense to open an issue for that where I can ask for clarification/suggest changes? It is about the formal grammar description only.

seldridge:
Issue works fine.

PR #50, opened 2022-12-02
[Clarification] Allow sub-access to convert to sub-index.

darthscsi:
Strictly speaking, sub-index and sub-access have different out-of-bounds behaviors.  Since a sub-access returnes an indeterminate value and sub-index has a compile-time error, one cannot convert sub-access with constant index to a sub-index if the index is out-of-bounds.  This explicitly allows that conversion.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1809,7 +1809,9 @@ The sub-access expression dynamically refers to a sub-element of a vector-typed
 expression using a calculated index. The index must be an expression with an
 unsigned integer type.  An access to an out-of-bounds element results in an 
 indeterminate value (see [@sec:indeterminate-values]).  Each out-of-bounds 
-element is a different indeterminate value.
+element is a different indeterminate value.  Sub-access operations with constant
+index may be convereted to sub-index operations even though it converts
+indeterminate-value-on-out-of-bounds behavior to a compile-time error.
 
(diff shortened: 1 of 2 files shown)


seldridge (approved):

jackkoenig:
Should this have a note in https://github.com/chipsalliance/firrtl-spec/blob/main/revision-history.yaml?

spec.md
```diff
@@ -1809,7 +1809,9 @@ The sub-access expression dynamically refers to a sub-element of a vector-typed
 expression using a calculated index. The index must be an expression with an
 unsigned integer type.  An access to an out-of-bounds element results in an 
 indeterminate value (see [@sec:indeterminate-values]).  Each out-of-bounds 
-element is a different indeterminate value.
+element is a different indeterminate value.  Sub-access operations with constant
+index may be convereted to sub-index operations even though it converts
+indeterminate-value-on-out-of-bounds behavior to a compile-time error.
```
jackkoenig, line 1814: I think this is a reasonable (and good) change but I will note that it does not match the behavior of SFC. This is obviously fine from a specification perspective but as we've struggled with improving handling of undefined behavior, it's probably best to at least call this out in the revision history.

PR #51, opened 2022-12-21
[Minor] Define Combinational loops

prithayan:
Specify that combinational loops are not allowed in FIRRTL

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2345,28 +2345,20 @@ to show that the loop does not exist under actual mux select values.
 Combinational loops are not allowed and it shouldn't depend on any FIRRTL
 transformation to remove or break such combinational loops.
 
-The following example is not legal, since a combinational loop exists, 
-`w -> b -> w`, even it can be proved that the loop does not exist after
-some simplification.
-
+The module `Foo` has a combinational loop and is not legal,
+even though the loop will be removed by last connect semantics.
 ``` firrtl
-wire a: UInt
-wire b: UInt
-wire c: UInt<1>
-wire notC: UInt<1>
-wire w: UInt
-w <= a
-when c :
-  w <= b
-notC <= not(c)
-when notC :
-  b <= w
-```
+  module Foo:
+    input a: UInt<1>
+    output b: UInt<1>
+    b <= b
+    b <= a
+ ```   
 
-The  following circuit has a combinational loop, even if it can be proved
+The following module `Foo2` has a combinational loop, even if it can be proved
 that `n1` and `n2` never overlap.
 ``` firrtl
-module MyModule :  
+module Foo2 :  
   input n1: UInt<2>
   input n2: UInt<2>
   wire tmp: UInt<1>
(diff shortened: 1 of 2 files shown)


ekiwi:
This looks great. Could you also elaborate on the fact that word level loops are forbidden even if they do not exist on the bit-level? You can have a look at this PR for some examples of word-level, but not bit-level combinatorial loops: https://github.com/chipsalliance/firrtl/pull/2481

spec.md
```diff
@@ -2334,6 +2334,46 @@ The width of each primitive operation is detailed in [@sec:primitive-operations]
...
+wire c: UInt<1>
+wire notC: UInt<1>
+wire w: UInt
+w <= a
+when c :
+  w <= b
+notC <= not(c)
+when notC :
+  b <= w
+```
```
seldridge, line 2363: I think this can be more plainly stated that combinational loops that would be removed by last connect semantics, conditional last connect semantics, or optimizations are still flagged as loops. The most basic example which shows a loop that would be removed by last connect semantics is:

```
circuit Foo:
  module Foo:
    input a: UInt<1>
    output b: UInt<1>
    b <= b
    b <= a
```

prithayan: Thanks for the suggestion. Updated the example.

prithayan:
> This looks great. Could you also elaborate on the fact that word level loops are forbidden even if they do not exist on the bit-level? You can have a look at this PR for some examples of word-level, but not bit-level combinatorial loops: [chipsalliance/firrtl#2481](https://github.com/chipsalliance/firrtl/pull/2481)

Thanks @ekiwi , added an example, from my understanding. Please review it.

spec.md
```diff
@@ -2334,6 +2334,52 @@ The width of each primitive operation is detailed in [@sec:primitive-operations]
...
 sections.
 
+# Combinational Loops
+
+Combinational logic is a section of logic with no registers between gates.
+A combinational loop exists when the output of some combinational logic 
+is fed back into the input of that combinational logic with no intervening
+register. FIRRTL does not support combinational loops even if it is possible
+to show that the loop does not exist under actual mux select values.
+Combinational loops are not allowed and it shouldn't depend on any FIRRTL
```
mwachs5, line 2344:
```suggestion
Combinational loops are not allowed and designs should not depend on any FIRRTL
```

mwachs5 (commented):
please add something to the version history for this change

spec.md
```diff
@@ -2334,6 +2334,52 @@ The width of each primitive operation is detailed in [@sec:primitive-operations]
 The width of the integer literal expressions is detailed in their respective
 sections.
 
+# Combinational Loops
+
+Combinational logic is a section of logic with no registers between gates.
+A combinational loop exists when the output of some combinational logic 
+is fed back into the input of that combinational logic with no intervening
+register. FIRRTL does not support combinational loops even if it is possible
```
darthscsi, line 2342: /support/allow/

spec.md
```diff
@@ -2334,6 +2334,52 @@ The width of each primitive operation is detailed in [@sec:primitive-operations]
...
+to show that the loop does not exist under actual mux select values.
```
darthscsi, line 2343: Perhaps, more directly, "FIRRTL does not allow combinatorial loops.  The existence of a combinatorial loop is a static, structural property of the code, not dependent on runtime-values."

darthscsi (approved):

PR #55, opened 2023-01-10
[minor] Add inline circuit annotations

seldridge:
Add inline JSON annotations for FIRRTL Circuits to enable the annotation file to be provided in the same file as the circuit.  This is a format that has been used for a while by CIRCT and makes it easier to package the annotation file along with the FIRRTL file.


jackkoenig (commented):
LGTM, but this needs mentioning in the revision-history.

seldridge:
> but this needs mentioning in the revision-history.

😅 Added in the fixup commit.

PR #56, opened 2023-01-10
[major] Remove Discussion of FIRRTL Forms

seldridge:
Remove all discussion of FIRRTL forms and lowering, i.e., high/mid/low FIRRTL.  Add language to indicate that high-level constructs may be preserved or eliminated by a FIRRTL compiler implementation.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -99,21 +99,16 @@ mechanism for error checking.
 ## Design Philosophy
 
 FIRRTL represents the standardized elaborated circuit that the Chisel HDL
-produces. FIRRTL represents the circuit immediately after Chisel's elaboration
-but before any circuit simplification. It is designed to resemble the Chisel HDL
-after all meta-programming has executed. Thus, a user program that makes little
-use of meta-programming facilities should look almost identical to the generated
+produces. FIRRTL represents the circuit immediately after Chisel's
+elaboration. It is designed to resemble the Chisel HDL after all
+meta-programming has executed. Thus, a user program that makes little use of
+meta-programming facilities should look almost identical to the generated
 FIRRTL.
 
 For this reason, FIRRTL has first-class support for high-level constructs such
-as vector types, bundle types, conditional statements, and modules. These
-high-level constructs are then gradually removed by a sequence of *lowering*
-transformations. During each lowering transformation, the circuit is rewritten
-into an equivalent circuit using simpler, lower-level constructs. Eventually the
-circuit is simplified to its most restricted form, resembling a structured
-netlist, which allows for easy translation to an output language
-(e.g. Verilog). This form is given the name *lowered FIRRTL* (LoFIRRTL) and is a
-strict subset of the full FIRRTL language.
+as vector types, bundle types, conditional statements, and modules. A FIRRTL
+compiler may choose to convert high-level constructs into low-level constructs
+before generating Verilog.
 
 Because the host language is now used solely for its meta-programming
 facilities, the frontend can be very light-weight, and additional HDLs written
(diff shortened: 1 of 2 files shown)


mwachs5:
I like the change. Does it need to be coordinated with any others order-wise?

ekiwi:
I don't think MidFirrtl matters, but it might be nice to keep the LoFirrtl definition around since some tools might want to consume that. Two that come to mind are treadle and ESSENT.

spec.md
```diff
@@ -2524,94 +2484,6 @@ circuit Foo: %[[
 Any legal JSON is allowed, meaning that the above JSON may be stored "minimized"
 all on one line.
 
-# The Lowered FIRRTL Forms
```
ekiwi, line 2527: Maybe we could change this section to just define LoFirrtl without any requirement that compilers must implement it. Such a definition would be convenient as some tools like simulators or formal backends might want to refer to LoFirrtl. One example might be that CIRCT could be used to convert to LoFirrtl, desugaring any new frontend features, and then that LoFirrtl could be fed into the existing SMT backend which does not care about high level features. Of course, there should not be any requirement that a compliant compiler always goes through LoFirrtl.

seldridge:
I don't really think that it's useful to define Low FIRRTL explicitly because this is really a pass ordering concern. Put differently, I don't want Low FIRRTL to be an interchange format.

My reasoning is that things like Treadle (for the wider audience: a FIRRTL interpreter), ESSENT (FIRRTL optimization passes and a FIRRTL to C++ converter), or the SMT backend (a FIRRTL to SMT converter) are all just passes that need to be run at specific points in a FIRRTL compilation pipeline. The fact that these intentionally don't support specific constructs is really stating that they need to run after certain passes common or available to almost any FIRRTL compiler. There's nothing special about Low FIRRTL that should be spec'd out.

That said... what is really missing (and I intend to add following this) is a section on `Compiler Implementation Details` which will enumerate: (1) mandatory passes and (2) optional passes. This will include and document all of "When to Mux Conversion and Last Connect Resolution" (`ExpandWhens`), "Lower Aggregate Types" (`LowerTypes`). "Width Inference" (`InferWidths`), and "Abstract Reset Inference" (`InferResets`). This section should then add a note indicating that an expected useful place to inject custom user passes is after running all of these passes.

WDYT?

(I also note that `LowerTypes` additionally needs to be documented correctly as the top-level module and any external modules must have their types lowered for correctness.)

ekiwi:
> Put differently, I don't want Low FIRRTL to be an interchange format.

I personally think it is useful as an interchange format, i.e., a structural description of the circuit, but I think it is very reasonable to disagree. Feel free to remove it.

> That said... what is really missing (and I intend to add following this) is a section on `Compiler Implementation Details` which will enumerate:

I don't think the spec should prescribe a compiler implementation. It should only specify the semantics of firrtl + any API/ABI type things that are necessary to interface with the outside world.



> (I also note that `LowerTypes` additionally needs to be documented correctly as the top-level module and any external modules must have their types lowered for correctness.)

Do I understand this correctly in that you are concerned about how "public" I/O gets lowered? If so, I do think that it would be great to define a "ABI" type interface that makes guarantees about how top-level / extmodule I/O will be represented in generated Verilog or VHDL.

seldridge:
> I don't think the spec should prescribe a compiler implementation. It should only specify the semantics of firrtl + any API/ABI type things that are necessary to interface with the outside world.

That's not quite what I'm saying. I'm saying that it would be useful for the FIRRTL spec document to include additional information about how you build a compiler. It's not part of the _specification_, but additional material that is useful (and borderline mandatory knowledge). There are already hints of this with the sections on `Width Inference`, `Reset Inference`, etc.

> Do I understand this correctly in that you are concerned about how "public" I/O gets lowered? If so, I do think that it would be great to define a "ABI" type interface that makes guarantees about how top-level / extmodule I/O will be represented in generated Verilog or VHDL. This "ABI" spec should not mention the specific compiler pass which is an implementation detail.

Yes, though the notion of "public" is not yet defined and needs to be defined.

I think it is fine to include references to the documentation about type lowering which should be included in a compiler implementation details section.

seldridge:
I added a new section called `FIRRTL Compiler Implementation Details` with a subsection on `Aggregate Type Lowering (Lower Types)`.

This section includes information that the top-level module and external modules must have their ports lowered. I originally thought that this should be included in the section on module and external module. However, I concluded that this is really a detail of the compiler implementation which isn't part of the spec-proper.

@ekiwi: I agree that the RISC-V call-outs are great. I would like to explore adding that type of stuff in the future.

azidar:
Cases where `validif` was used (a mux where one branch was invalid) is probably needed in the IR spec if you want to represent that concept directly in LowFirrtl as an interchange format. Otherwise, you'd need either to use a `when`, or just use a mux and have the invalid branch be a reference to a wire who is invalidated.

azidar (approved):

PR #57, opened 2023-01-11
[major] Remove validIf

seldridge:
Remove the validIf expression from the spec.  This expression is an artifact of the Scala-based FIRRTL Compiler's internal representation and how it lowers conditional expressions.  It is not necessary that this is part of the public-facing FIRRTL spec.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1213,11 +1213,10 @@ wire w: UInt
 w <= mux(c, b, a)
 ```
 
-In the case where an invalid statement is followed by a conditional statement
-containing a connect to the invalidated component, the resulting connection to
-the component can be expressed using a conditionally valid expression. See
-[@sec:conditionally-valids] for more details about the conditionally valid
-expression.
+Because invalid statements assign indeterminate values to components, a FIRRTL
+Compiler is free to choose any specific value for an indeterminate value when
+resolving last connect semantics.  E.g., in the following circuit `w`{.firrtl}
+has an indeterminate value when `c`{.firrtl} is false.
 
 ``` firrtl
 wire a: UInt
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -1213,11 +1213,10 @@ wire w: UInt
 w <= mux(c, b, a)
 ```
 
-In the case where an invalid statement is followed by a conditional statement
-containing a connect to the invalidated component, the resulting connection to
-the component can be expressed using a conditionally valid expression. See
-[@sec:conditionally-valids] for more details about the conditionally valid
-expression.
+Because invalid statements assign indeterminate values to components, a FIRRTL
+Compiler is free to choose any value for an indeterminate value when resolving
```
ekiwi, line 1217: "any **concrete** value" ?

spec.md
```diff
@@ -1228,15 +1227,17 @@ when c :
   w <= a
 ```
 
-can be rewritten equivalently as follows:
+A FIRRTL compiler is free to optimize this to the following circuit.
```
ekiwi, line 1230: "by assuming that `w` takes on the value of `a` when `not(c)`"

ekiwi: Not sure about the exact wording, but expanding on how we reason about this optimization would be helpful, especially in the context of the previous definition which states that "a FIRRTL Compiler is free to choose any value for an indeterminate value when resolving last connect semantics"

spec.md
seldridge: Good point, are you fine with use of _specific_ value (which is the language in the indeterminate value section)?

ekiwi (approved):
LGTM

PR #59, opened 2023-01-11
[major] Remove Partial Connect (<-)

seldridge:
Remove the partial connect operation ("<-").  This commit should be merged once it is no longer possible for Chisel to emit partial connects.

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -17,6 +17,7 @@ revisionHistory:
     - Change connect to truncate widths to align with all existing FIRRTL
       Compiler implementations
     - Fix spelling/grammar issues
+    - Remove partial connect ("<-")
   # Information about the old versions.  This should be static.
   oldVersions:
     - version: 1.1.0
--- a/spec.md
+++ b/spec.md
@@ -106,14 +106,14 @@ use of meta-programming facilities should look almost identical to the generated
 FIRRTL.
 
 For this reason, FIRRTL has first-class support for high-level constructs such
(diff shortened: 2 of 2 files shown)


azidar:
Chisel's 5.0.0 release will remove the partial connection emission, as the compatibility mode will have been deleted.

azidar (approved):

seldridge:
Thinking about this more, I think it's fine to merge this now given how quickly Chisel 5 is coming as I think partial connect is only accessible through compile options right now anyway, yes?

Any objection to merging now?

jackkoenig:
I think it would be good to crack a version with https://github.com/chipsalliance/firrtl-spec/pull/58 before we move into the brave new world of what should probably be 2.0.0.

seldridge:
> I think it would be good to crack a version with https://github.com/chipsalliance/firrtl-spec/pull/58 before we move into the brave new world of what should probably be 2.0.0.

There's already a lot of accumulated major changes that would force a 2.0.0 version if I created a tag. I can always back-tag things if we'd like finer grained versioning (unless I cut a tag that lumps multiple major/minor changes). I can cut a 2.0.0 release with #58 and then merge (or back tag later if we choose).

jackkoenig:
Yeah fair point. I think we should consider back-tagging a 1.2.0 for SFC just to make it easy for people to see what is supported by what, but point taken, no reason to block on that.

seldridge:
I'm going to merge all this stuff onto `main` then create a synthetic branch for a 1.2 version with accumulated changes.

PR #60, opened 2023-01-12
[patch] Fix minor spelling and grammar issues

tymcauley:
It looked like we were using "behavior" (10 instances) more than "behaviour" (5 instances), so I went with the more common one.

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -16,6 +16,7 @@ revisionHistory:
     - Remove conditionally valid expression (`validif`)
     - Change connect to truncate widths to align with all existing FIRRTL
       Compiler implementations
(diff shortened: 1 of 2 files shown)


seldridge (approved):
Thanks for taking the time to fix this!

Can you modify the PR title (or commit message) to include `[patch]` in it and to add a line in the `revision-history.yaml` about "spelling/grammar fixes"?

tymcauley:
All set! Should I also re-run `scripts/get-contributors.sh`?

seldridge:
> All set! Should I also re-run scripts/get-contributors.sh?

I can run this later. This is querying GitHub information which would need to happen after this gets committed.

PR #63, opened 2023-01-19
[patch] Fix various typos.

dtzSiFive:


spec.md
```diff
@@ -1402,7 +1402,7 @@ wire b: UInt
 printf(clk, cond, "a in hex: %x, b in decimal:%d.\n", a, b) : optional_name
 ```
 
-On each positive clock edge, when the condition signal is high, the printf
+On each positive clock edge, when the condition signal is high, the `printf`{.firrtl}
```
seldridge, line 1405: Nit: 80-column. Likely other places, too.

seldridge (approved):
LGTM

80-column nits and use the `[patch] label`. This doesn't need to update the revision history as there's already a generic spelling/grammar issues line item.

spec.md
dtzSiFive: Sure thing, done!

dtzSiFive:
> 80-column nits and use the `[patch] label`. This doesn't need to update the revision history as there's already a generic spelling/grammar issues line item.

Done, thanks for the guidance.

azidar (approved):

PR #64, opened 2023-01-19
[patch] Clarify cover string argument as comment not message.

dtzSiFive:
Tweak the existing field to match its use in practice (SFC), and planned to soon behave similarly in MFC: https://github.com/llvm/circt/pull/4502 .

See #53 for discussion of this field, for now let's keep existing code working and update spec to reflect current behavior (and provide users a means to attach string information to the output that's not limited like `optional_name`).


seldridge (approved):
Good to define this existing behavior. 👍  LGTM

PR #65, opened 2023-01-20
Minor grammar changes

keszocze:
I am currently writing a FIRRTL lexer for [Pygments](https://pygments.org/) and stumbled upon some minor issues in the formal specification that I fixed. Furthermore I found some curiosities in the specs as well as a mismatch between what Chisel 3.5.5. produces and what is described in the spec. 

### Fixes
* Removal of unnecessary parentheses and a spurious colon
* Allowing to create decimal integer expressions (the compilers, of course, always allowed for that)
* Enforcing an `at least once` constraint in many places. The grammar uses `{ rule }` quite often when `rule , { rule }` should have been used. I do not know what formalism exactly is used in the "FIRRTL Language Definition", but as it looks XBNF enough for me, I opted to use the `{ rule }+` notation to indicate the "at least once" constraint.

#### Names 
The rule `id = ("_" | letter) , {"_" | letter | digit_dec)` allows for a name to be a single underscore only. The compiler   happily accepts this:
```Scala
val emptyNameExample =
    """
circuit emptyName :
  module emptyName :
    output _ : UInt<4>		
    out <= UInt(0)
"""
```

   ```
[info] running mhbs.FIRRTLParse 
FIRRTL version 1.1.0
circuit emptyName :
  module emptyName :
    output _ : UInt<4>

    out <= UInt<1>("h0")
```

   Is this intended behavior?

#### Comments
The rule `info = "@" , "[" , { string , " " , linecol } , "]" ;` allows for empty comments. The compiler happily accepts this:
```Scala
val emptyCommentExample =
    """
circuit emptyComment :
  module emptyComment :
    output out : SInt<4> @[]		
    out <= SInt(0)
"""
```

```
[info] running mhbs.FIRRTLParse 
FIRRTL version 1.1.0
circuit emptyComment :
  module emptyComment :
    output out : SInt<4>

    out <= SInt<1>("h0")
```

Is this intended behavior?

### Mismatches for registers

When using Chisel 3.5.5 registers result in the following 
```Scala
class RegisterExample extends Module{
    val out = IO(Output(UInt(4.W)))
    val r = RegInit(0.U(4.W))
    out := r
}
```
```
    reg r : UInt<4>, clock with :
      reset => (reset, UInt<4>("h0")) @[RegisterExample.scala 8:20]
```

while the spec would lead to
```
reg r: UInt<4> clock (with: {reset => (UInt<4>("h0"), r)})
```

Is the spec just newer than the Chisel version?

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1309,7 +1309,7 @@ analog types within the component (as they cannot be connected to).
 
 ### Alternate Syntax
 
-`is invalid`.{.firrtl} may also be specified by keyword.
+`is invalid`{.firrtl} may also be specified by keyword.
 
 ``` firrtl
 module MyModule :
--- a/spec.md
+++ b/spec.md
@@ -3672,11 +3672,11 @@ statement = "wire" , id , ":" , type , [ info ]
           | reference , "<=" , expr , [ info ]
           | reference , "is invalid" , [ info ]
           | "attach(" , reference , { "," ,  reference } , ")" , [ info ]
-          | "when" , expr , ":" [ info ] , newline , indent ,
-              statement, { statement } ,
-            dedent , [ "else" , ":" , indent , statement, { statement } , dedent ]
-          | "stop(" , expr , "," , expr , "," , int_any , ")" , [ info ]
-          | "printf(" , expr , "," , expr , "," , string_dq ,
+          | "when" , expr , ":" [ info ] , newline ,
+            indent , statement, { statement } , dedent ,
+            [ "else" , ":" , indent , statement, { statement } , dedent ]
+          | "stop(" , expr , "," , expr , "," , int , ")" , [ info ]
+          | "printf(" , expr , "," , expr , "," , string ,
             { expr } , ")" , [ ":" , id ] , [ info ]
           | "skip" , [ info ]
           | "define" , static_reference , "=" , ref_expr , [ info ]
--- a/spec.md
+++ b/spec.md
@@ -3588,7 +3588,7 @@ letter = "A" | "B" | "C" | "D" | "E" | "F" | "G"
 id = ( "_" | letter ) , { "_" | letter | digit_dec } ;
 
 (* Fileinfo communicates Chisel source file and line/column info *)
-linecol = num_dec , ":" , num_dec ;
+linecol = digit_dec , { digit_dec } , ":" , digit_dec , { digit_dec } ;
 lineinfo = string, " ", linecol
 info = "@" , "[" , lineinfo, { ",", lineinfo }, "]" ;
 
@@ -3677,7 +3677,7 @@ statement = "wire" , id , ":" , type , [ info ]
             indent , statement, { statement } , dedent ,
             [ "else" , ":" , indent , statement, { statement } , dedent ]
           | "stop(" , expr , "," , expr , "," , int , ")" , [ info ]
-          | "printf(" , expr , "," , expr , "," , string ,
+          | "printf(" , expr , "," , expr , "," , string_dq ,
             { expr } , ")" , [ ":" , id ] , [ info ]
           | "skip" , [ info ]
           | "define" , static_reference , "=" , ref_expr , [ info ]
@@ -3709,7 +3709,7 @@ intmodule = "intmodule" , id , ":" , [ info ] , newline , indent ,
 annotations = "%" , "[" , json_array , "]" ;
 
 (* Version definition *)
-sem_ver = num_dec , "."  , num_dec , "." , num_dec
+sem_ver = int , "."  , int , "." , int
 version = "FIRRTL" , "version" , sem_ver ;
 
 (* Circuit definition *)
--- a/spec.md
+++ b/spec.md
@@ -1072,7 +1072,7 @@ field.
 
 ### Alternate Syntax
 
-Connects may also be specified by keyword.  This form is identical to the `<=` 
+Connects may also be specified by keyword.  This form is identical to the `<=`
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -2622,10 +2622,10 @@ digit_hex = digit_dec
...
 
 (* An integer *)
-int = '"' , "b" , [ "-" ] , { digit_bin } , '"'
-    | '"' , "o" , [ "-" ] , { digit_oct } , '"'
-    | '"' , "h" , [ "-" ] , { digit_hex } , '"'
-    |             [ "-" ] , { digit_bin } ;
+int = '"' , "b" , [ "-" ] , { digit_bin }+ , '"'
+    | '"' , "o" , [ "-" ] , { digit_oct }+ , '"'
+    | '"' , "h" , [ "-" ] , { digit_hex }+ , '"'
+    |             [ "-" ] , { digit_dec }+ ;
```
seldridge, line 2628: I'd prefer to use the pedantic `digit_dec , { digit_dec }` formulation as opposed to using non-standard, though obvious to anyone who's written a regex, `+`. This is only to avoid future confusion around what this means.

If there are problems fitting this on a page, feel free to use more preliminary definitions, e.g.,:

```
num_dec = digit_dec , { digit_dec } ;
```

I have been annoyed about this with EBNF that it doesn't have one-or-more. 🥲 

Ditto for other corrections to use `+`.

seldridge: Nice catch about `digit_dec` vs. `digit_bin`. 👍

spec.md
```diff
@@ -2708,17 +2708,17 @@ statement = "wire" , id , ":" , type , [ info ]
           | "node" , id , "=" , expr , [ info ]
           | reference , "<=" , expr , [ info ]
           | reference , "is invalid" , [ info ]
-          | "attach(" , { reference } , ")" , [ info ]
+          | "attach(" , reference , { "," ,  reference } , ")" , [ info ]
```
seldridge, line 2711: Nice. Empty attaches don't make sense.

seldridge (commented):
Thanks for giving this a close inspection!

Changes look to correct a number of mistakes.

Generally, I'd prefer to not invent new syntax `+` only to avoid confusion. Is there a way to rework this to use more preliminary definitions, even if that increases the line count?

seldridge:
### Underscore Names

```
circuit emptyName :
  module emptyName :
    output _ : UInt<4>		
    out <= UInt(0)
```
☝️ This is illegal, but due to the undefined id `out`. I think you meant to have `_ <= UInt(0)` ? 

The legality of `_` as an identifier is intentional. This stems from this being legal Verilog. E.g., Verilator has no problem with this (and it's usually really good about flagging these problems). There should be a reference in the Verilog spec that `_` is okay. I haven't checked VHDL:

```
# cat Baz.fir && firtool Baz.fir | tee /dev/stderr | verilator -lint-only /dev/stdin
circuit emptyName :
  module emptyName :
    output _ : UInt<4>
    _ <= UInt(0)
// Generated by CIRCT unknown git version
module emptyName(
  output [3:0] _);

  assign _ = 4'h0;
endmodule
```

### Empty file info blocks

I think we should reject these. (Spec is wrong and should require one-or-more.)

### Register Format

I may have just screwed up the EBNF. Either of the following are allowable formats:

```
circuit emptyComment :
  module emptyComment :
    input clock: Clock

    reg r: UInt<4> clock with: (reset => (UInt<4>("h0"), r))

    reg s: UInt<4> clock with:
      (reset => (UInt<4>("h0"), s))
```

spec.md
keszocze: Okay, I'll explicitly list it then.

keszocze:
> ### Underscore Names
> point_up This is illegal, but due to the undefined id `out`. I think you meant to have `_ <= UInt(0)` ?

Yes, my bad. It shortened the example without checking again. If that *really* is a legal identifier, it shouldn't be removed from the spec.

> ### Empty file info blocks
> 
> I think we should reject these. (Spec is wrong and should require one-or-more.)

I pushed the following proposed change:
```
(* Fileinfo communicates Chisel source file and line/column info *)
linecol = num_dec , ":" , num_dec ;
lineinfo = string, " ", linecol
info = "@" , "[" , lineinfo, { ",", lineinfo }, "]" ;
```

As this is a "visual" change, I labeled the corresponding commit `[patch]`. 

> ### Register Format
> 
> I may have just screwed up the EBNF. Either of the following are allowable formats:
> 
> ```
> circuit emptyComment :
>   module emptyComment :
>     input clock: Clock
> 
>     reg r: UInt<4> clock with: (reset => (UInt<4>("h0"), r))
> 
>     reg s: UInt<4> clock with:
>       (reset => (UInt<4>("h0"), s))
> ```

So the `{`'s in the spec should be removed. There are other differences still. My locally installed Chisel (version 3.5.5) generates
```
    reg r : UInt<4>, clock with :
      reset => (reset, UInt<4>("h0")) @[RegisterExample.scala 8:20]
```

The differences are:
* Chisel puts the opening parenthesis *after* the `=>`, not in front of the `reset`
* Chisel repeats the signal that is used to reset the register, you repeated the register name

keszocze:
I also added some minor edits that
* Add (or fix) `{.firrtl}` annotations where necessary
* Remove references to the now-gone partial connect algorithm

I don't think that opening a separate PR for this would be necessary as the changes should be uncontroversial.

darthscsi (commented):
Generally looks good to me now.

PR #67, opened 2023-01-23
[Major] Explicit compile time constant expressions.

darthscsi:
Add a constant type modifier for compile-time evaluation (under whole-world assumptions).  Constant types in external ports implies that uses of constant types are not guaranteed to be know at compile time (this is not a mechanism to guarantee constant propagation).

Asynchronous resets are updated to take constant reset values.  This moves ad-hoc checks which exist in implementations now to type checks.  This is not sufficient for some tools, but is a strict improvement on the current implementations.

The change of asyncreset value is what makes this a major change.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -503,17 +503,18 @@ flows into the module, and the `d`{.firrtl} sub-field contained in the
 
 ### Constant Type
 
-A constant type is a type whose value is guaranteed to be a compile-time 
-constant.  Constant types may be used in ports, wire, nodes, and generally 
-anywhere a non-constant type is usable.  Operations on constant type are well 
-defined.  As a general rule (with any exception listed in the definition for 
-such operations as have exceptions), an operation whose arguments are constant
-produces a constant.  An operations with some non-constant arguments produce a
-non-constant.  Constants can be used in any context with a source flow which 
-allows a non-constant.  Constants may be used as the target of a connect so long 
-as the source of the connect is itself constant.  These rules ensure all 
-constants are derived from literals or from constant-typed input ports of the 
-top-level module.
+A constant type is a type whose value is guaranteed to be unchanging at circuit 
+execution time.  Constant is a constraint on the mutability of the value, it 
+does not imply a literal value at a point in the emitted design.  Constant types 
+may be used in ports, wire, nodes, and generally anywhere a non-constant type is 
+usable.  Operations on constant type are well defined.  As a general rule (with 
+any exception listed in the definition for such operations as have exceptions), 
+an operation whose arguments are constant produces a constant.  An operations 
+with some non-constant arguments produce a non-constant.  Constants can be used 
+in any context with a source flow which allows a non-constant.  Constants may be 
+used as the target of a connect so long as the source of the connect is itself 
+constant.  These rules ensure all constants are derived from literals or from 
+constant-typed input ports of the top-level module.
 
 ``` firrtl
 const UInt<3>
@@ -528,8 +529,8 @@ means if a constant is being assigned to in a `when` block, the `when`'s
 condition must be a constant.
 
 Output ports of external modules and input ports to the top-level module may be
-constant.  In such case, the value of the port is not known, but that it is time
-invariant at runtime is known.
+constant.  In such case, the value of the port is not known, but that it is 
+non-mutating at runtime is known.
 
 The indexing of a constant aggregate produces a constant of the appropriate type 
 for the element.
@@ -771,7 +772,8 @@ the current value of the element, writes are not visible until after a positive
 edges of the register's clock port.
 
 The clock signal for a register must be of type `Clock`{.firrtl}.  The type of a
-register must be a passive type (see [@sec:passive-types]) and may not be const.
+register must be a passive type (see [@sec:passive-types]) and may not be 
+`const`.{.firrtl}.
 
 The following example demonstrates instantiating a register with the given name
 `myreg`{.firrtl}, type `SInt`{.firrtl}, and is driven by the clock signal
@@ -1692,7 +1694,8 @@ will be rewritten as "the port `in`{.firrtl} is connected to the port
 
 The sub-field expression refers to a sub-element of an expression with a bundle
 type.  If the expression is of a constant bundle type, the sub-element shall be 
-of a constant type.
+of a constant type (`const`.{firrtl} propagates from the bundle to the element 
+on indexing).
 
 The following example connects the `in`{.firrtl} port to the `a`{.firrtl}
 sub-element of the `out`{.firrtl} port.
@@ -1703,6 +1706,26 @@ module MyModule :
   output out: {a: UInt, b: UInt}
   out.a <= in
 ```
+
+The following example is the same as above, but with a constant output bundle.
+
+``` firrtl
+module MyModule :
+  input in: const UInt
+  output out: const {a: UInt, b: UInt}
+  out.a <= in // out.a is of type const UInt
+```
+
+The following example is the same as above, but with a bundle with a constant 
+field.
+
+``` firrtl
+module MyModule :
+  input in: const UInt
+  output out: {a: const UInt, b: UInt}
+  out.a <= in // out.a is of type const UInt
+```
+
 
 ## Sub-indices
 
@@ -1721,6 +1744,16 @@ module MyModule :
   output out: UInt[10]
   out[4] <= in
 ```
+
+The following example is the same as above, but with a constant vector.
+
+``` firrtl
+module MyModule :
+  input in: const UInt
+  output out: const UInt[10]
+  out[4] <= in // out[4] has a type of const UInt
+```
+
 
 ## Sub-accesses
 
--- a/spec.md
+++ b/spec.md
@@ -1371,6 +1371,12 @@ data that is read out of the read port is undefined.
 In all cases, if a memory location is written to by more than one port on the
 same cycle, the stored value is undefined.
 
+### Constant memory type
+
+A memory with a constant data-type represents a ROM and may not have 
+write-ports.  It is beyond the scope of this specification how ROMs are 
+initialized.
+
 ## Instances
 
 FIRRTL modules are instantiated with the instance statement. The following
--- a/spec.md
+++ b/spec.md
@@ -2711,7 +2711,8 @@ type_ground = "Clock" | "Reset" | "AsyncReset"
 type_aggregate = "{" , field , { field } , "}"
                | type , "[" , int , "]" ;
 field = [ "flip" ] , id , ":" , type ;
-type = type_ground | type_aggregate ;
+consttype = "const", type_ground | type_aggregate
+type = type_ground | type_aggregate | consttype ;
 
 (* Primitive operations *)
 primop_2expr_keyword =
--- a/spec.md
+++ b/spec.md
@@ -540,13 +540,13 @@ for the element.
 Constant types are a restriction on FIRRTL types.  Therefore, FIRRTL structures 
 which would be expected to produce certain Verilog structures will produce the
 same structure if instantiated with a constant type.  For example, an input port
-of type `const UInt` will result in a port in the verilog, if under the same
+of type `const UInt` will result in a port in the Verilog, if under the same
 conditions an input port of type `UInt` would have.
 
 It is not intended that constants are a replacement for parameterization.  
 Constant typed values have no particular meta-programming capability.  It is,
 for example, expected that a module with a constant input port be fully 
-compilable to non-parameterized verilog.
+compilable to non-parameterized Verilog.
 
 
 ## Passive Types
--- a/spec.md
+++ b/spec.md
@@ -499,6 +499,54 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
 flows into the module, and the `d`{.firrtl} sub-field contained in the
 `b`{.firrtl} sub-field flows out of the module.
 
+## Type Modifiers
+
+### Constant Type
+
+A constant type is a type whose value is guaranteed to be a compile-time 
+constant.  Constant types may be used in ports, wire, nodes, and generally 
+anywhere a non-constant type is usable.  Operations on constant type are well 
+defined.  As a general rule (with any exception listed in the definition for 
+such operations as have exceptions), an operation whose arguments are constant
+produces a constant.  Operations, except where noted, do not take a mixture of 
+constant and non-constant arguments.  Constants can be used in any context with 
+a source flow which allows a non-constant.  Constants may be used as the target 
+of a connect so long as the source of the connect is itself constant.  These 
+rules ensure all constants are derived from literals or from input ports to the 
+top-level module which are constant.
+
+``` firrtl
+const UInt<3>
+const SInt
+const {real: UInt<32>, imag : UInt<32>}
+const 
+
+Last-connect semantics of constant typed values is well defined, so long as any 
+control flow is conditioned on an expression which has a constant type.  This 
+means if a constant is being assigned to in a `when` block, the `when`'s 
+condition must be a constant.
+
+Output ports of external modules and input ports to the top-level module may be
+constant.  In such case, the value of the port is not known, but that it is time
+invariant at runtime is known.
+
+The indexing of a constant aggregate produces a constant of the appropriate type 
+for the element.
+
+#### A note on implementation
+
+Constant types are a restriction on firrtl types.  Therefore, firrtl structures 
+which would be expected to produce certain verilog structures will produce the
+same structure if instantiated with a constant type.  For example, an input port
+of type `const UInt` will result in a port in the verilog, if under the same
+conditions an input port of type `UInt` would have.
+
+It is not intended that constants are a replacement for parameterization.  
+Constant typed values have no particular meta-programming capability.  It is,
+for example, expected that a module with a constant input port be fully 
+compilable to non-parameterized verilog.
+
+
 ## Passive Types
 
 It is inappropriate for some circuit components to be declared with a type that
--- a/spec.md
+++ b/spec.md
@@ -508,12 +508,12 @@ constant.  Constant types may be used in ports, wire, nodes, and generally
 anywhere a non-constant type is usable.  Operations on constant type are well 
 defined.  As a general rule (with any exception listed in the definition for 
 such operations as have exceptions), an operation whose arguments are constant
-produces a constant.  Operations, except where noted, do not take a mixture of 
-constant and non-constant arguments.  Constants can be used in any context with 
-a source flow which allows a non-constant.  Constants may be used as the target 
-of a connect so long as the source of the connect is itself constant.  These 
-rules ensure all constants are derived from literals or from input ports to the 
-top-level module which are constant.
+produces a constant.  An operations with some non-constant arguments produce a
+non-constant.  Constants can be used in any context with a source flow which 
+allows a non-constant.  Constants may be used as the target of a connect so long 
+as the source of the connect is itself constant.  These rules ensure all 
+constants are derived from literals or from constant-typed input ports of the 
+top-level module.
 
 ``` firrtl
 const UInt<3>
@@ -770,7 +770,7 @@ the current value of the element, writes are not visible until after a positive
 edges of the register's clock port.
 
 The clock signal for a register must be of type `Clock`{.firrtl}.  The type of a
-register must be a passive type (see [@sec:passive-types]).
+register must be a passive type (see [@sec:passive-types]) and may not be const.
 
 The following example demonstrates instantiating a register with the given name
 `myreg`{.firrtl}, type `SInt`{.firrtl}, and is driven by the clock signal
@@ -1562,8 +1562,8 @@ for performing primitive operations.
 ## Unsigned Integers
 
 A literal unsigned integer can be created given a non-negative integer value and
-an optional positive bit width. The following example creates a 10-bit unsigned
-integer representing the number 42.
+an optional positive bit width. Integer literals are of constant type.  The 
+following example creates a 10-bit unsigned integer representing the number 42.
 
 ``` firrtl
 UInt<10>(42)
@@ -1580,7 +1580,8 @@ UInt(42)
 ## Unsigned Integers from Literal Bits
 
 A literal unsigned integer can alternatively be created given a string
-representing its bit representation and an optional bit width.
+representing its bit representation and an optional bit width.  Like with the
+integer representation, the expression has a constant type.
 
 The following radices are supported:
 
@@ -1617,8 +1618,9 @@ UInt<7>("hD")
 ## Signed Integers
 
 Similar to unsigned integers, a literal signed integer can be created given an
-integer value and an optional positive bit width. The following example creates
-a 10-bit unsigned integer representing the number -42.
+integer value and an optional positive bit width. Integer literals are of 
+constant type.  The following example creates a 10-bit unsigned integer 
+representing the number -42.
 
 ``` firrtl
 SInt<10>(-42)
@@ -1637,7 +1639,8 @@ SInt(-42)
 
 Similar to unsigned integers, a literal signed integer can alternatively be
 created given a string representing its bit representation and an optional bit
-width.
+width.  Like with the integer representation, the expression has a constant 
+type.
 
 The bit representation contains a binary, octal or hex indicator, followed by an
 optional sign, followed by the value.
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -499,6 +499,54 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
...
+Constant types are a restriction on firrtl types.  Therefore, firrtl structures 
+which would be expected to produce certain verilog structures will produce the
+same structure if instantiated with a constant type.  For example, an input port
+of type `const UInt` will result in a port in the verilog, if under the same
+conditions an input port of type `UInt` would have.
+
+It is not intended that constants are a replacement for parameterization.  
+Constant typed values have no particular meta-programming capability.  It is,
+for example, expected that a module with a constant input port be fully 
+compilable to non-parameterized verilog.
```
ekiwi, line 547: Maybe it would help to clarify that the values are _firrtl compile time_ constants, not _Verilog compile time_ constants (as would be parameters).

spec.md
```diff
@@ -499,6 +499,54 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
...
+constant.  Constant types may be used in ports, wire, nodes, and generally 
+anywhere a non-constant type is usable.  Operations on constant type are well 
+defined.  As a general rule (with any exception listed in the definition for 
+such operations as have exceptions), an operation whose arguments are constant
+produces a constant.  An operations with some non-constant arguments produce a
+non-constant.  Constants can be used in any context with a source flow which 
+allows a non-constant.  Constants may be used as the target of a connect so long 
+as the source of the connect is itself constant.  These rules ensure all 
+constants are derived from literals or from constant-typed input ports of the 
+top-level module.
```
ekiwi, line 516: Shouldn't the toplevel be forbidden from featuring constant type input ports? Since then we would not be able to tell the value at compile time?

spec.md
```diff
@@ -499,6 +499,54 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
...
+const 
+
+Last-connect semantics of constant typed values is well defined, so long as any 
+control flow is conditioned on an expression which has a constant type.  This 
+means if a constant is being assigned to in a `when` block, the `when`'s 
+condition must be a constant.
+
+Output ports of external modules and input ports to the top-level module may be
+constant.  In such case, the value of the port is not known, but that it is time
+invariant at runtime is known.
```
ekiwi, line 531: If the value of the port is not know, it isn't a compile time constant, right?
Are you trying to emulate C++'s `const` instead of `constexpr`?

spec.md
```diff
@@ -499,6 +499,54 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
...
+such operations as have exceptions), an operation whose arguments are constant
+produces a constant.  An operations with some non-constant arguments produce a
+non-constant.  Constants can be used in any context with a source flow which 
+allows a non-constant.  Constants may be used as the target of a connect so long 
+as the source of the connect is itself constant.  These rules ensure all 
+constants are derived from literals or from constant-typed input ports of the 
+top-level module.
+
+``` firrtl
+const UInt<3>
```
mwachs5, line 519: shouldn't this result in something in the parser grammar at the end?

ekiwi (changes requested):
I like the idea of formalizing legal reset values. However, I am quite confused on whether you want `const` to be a compile time constant, which would mean that the toplevel cannot have `const` inputs (and extmodule no `const` outputs) or if you want it to just be a value that does not change. Personally I feel like the more strict focus on compile time constants, i.e., expressions that the compiler can compute the value of, would be beneficial since that would also allow us to use the `const` attribute for things like MSB/LSB in bit-extract, SubIndex, and even `UInt` widths.

I would also be very curious how this is supposed to be implemented in Chisel. Do you have a PR for that ready? Is Chisel going to annotate `const` wherever possible? Is the user going to explicitly annotate `const`? Or are you going to run an inference pass in the firrtl compiler that greedily annotates `const`?

spec.md
darthscsi: thanks.

spec.md
```diff
@@ -722,7 +770,7 @@ the current value of the element, writes are not visible until after a positive
 edges of the register's clock port.
 
 The clock signal for a register must be of type `Clock`{.firrtl}.  The type of a
-register must be a passive type (see [@sec:passive-types]).
+register must be a passive type (see [@sec:passive-types]) and may not be const.
```
mwachs5, line 773:
```suggestion
register must be a passive type (see [@sec:passive-types]) and may not be `const`{.firrtl}.
```

Maybe? I'm not sure the rules when we use these modifiers or not

spec.md
```diff
@@ -1517,8 +1566,8 @@ for performing primitive operations.
 ## Unsigned Integers
 
 A literal unsigned integer can be created given a non-negative integer value and
-an optional positive bit width. The following example creates a 10-bit unsigned
-integer representing the number 42.
+an optional positive bit width. Integer literals are of constant type.  The 
+following example creates a 10-bit unsigned integer representing the number 42.
 
 ``` firrtl
```
mwachs5, line 1572: Is this equivalent to `const UInt<10>(42)`? Or is it illegal to put the const in front (part of the type only not of the value)?

seldridge: You wouldn't be able to do this. It's saying that `Uint<10>(42)` has type `const Uint<10>`.

Using a Scala-like syntax for type annotations. Before the type was:

```
node a: Uint<1> = Uint<1>(42)
```

Now the type is:

```
node a: const Uint<1> = Uint<1>(42)
```

spec.md
seldridge: This is intentionally allowed. 

The language here is intending that a value is known to be a constant value to the compiler, not that the compiler knows the value.

This is intended to support a use case of two separately compiled circuits that are then linked at FIRRTL or Verilog levels. Without further FIRRTL spec modifications, this has to be represented as a boundary where an extmodule has a const port and a top-level module also has a const port.

ekiwi: Then I don't think that it is correct to call these `compile time` constants. They are more `runtime constants`, i.e. the value does not change while the circuit is executing, but the compiler does not actually know the value.

ekiwi: To make the comparison to C++ again: You can have a `extern const int a;` (which is resolved at link time, just like your example) but not a `extern constexpr int a;` (which would be a compile time constant expression).

spec.md
```diff
@@ -737,15 +785,16 @@ reg myreg: SInt, myclock
...
-signal (see [@sec:reset-type]).  `Reset`{.firrtl} is an abstract reset whose
-behavior depends on reset inference.  In the following example, `myreg`{.firrtl}
-is assigned the value `myinit`{.firrtl} when the signal `myreset`{.firrtl} is
-high.
+the type of reset value must be equivalent to the declared type of the register 
+(see [@sec:type-equivalence] for details).  If the reset signal is an 
+`AsyncReset`{.firrtl}, then the reset value must be a constant type.  The 
+behavior of the register depends on the type of the reset signal.  
+`AsyncReset`.{firrtl} will immediately change the value of the register.  
+`UInt<1> will not change the value of the register until the next positive edge
```
dtzSiFive, line 793:
```suggestion
`UInt<1>`{.firrtl} will not change the value of the register until the next positive edge 
```
(fix run-away, primarily)

spec.md
```diff
@@ -499,6 +499,54 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
...
+allows a non-constant.  Constants may be used as the target of a connect so long 
+as the source of the connect is itself constant.  These rules ensure all 
+constants are derived from literals or from constant-typed input ports of the 
+top-level module.
+
+``` firrtl
+const UInt<3>
+const SInt
+const {real: UInt<32>, imag : UInt<32>}
+const
```
azidar, line 522: Looks like you are missing a trailing ```

spec.md
azidar: Yeah I think I agree with @ekiwi , they are more runtime constants. Maybe the wording could be clarified?

spec.md
dtzSiFive: Me either, but for lack of a better idea I think they should be used on all FIRRTL language snippets (including variable references), but certainly involving keywords.

I don't think there's consistent use, although I added it to a few places recently.   Any consistent rule is probably an improvement.

spec.md
```diff
@@ -722,7 +770,7 @@ the current value of the element, writes are not visible until after a positive
 edges of the register's clock port.
 
 The clock signal for a register must be of type `Clock`{.firrtl}.  The type of a
-register must be a passive type (see [@sec:passive-types]).
```
dtzSiFive, line 725: Probably don't want to allow const memories either?

spec.md
```diff
@@ -1638,7 +1690,8 @@ will be rewritten as "the port `in`{.firrtl} is connected to the port
 ## Sub-fields
 
 The sub-field expression refers to a sub-element of an expression with a bundle
-type.
+type.  If the expression is of a constant bundle type, the sub-element shall be 
+of a constant type.
```
rwy7, line 1694: I think there is a bug in the new wording. If I understand correctly, a bundle is const iff all fields are const. So if we take a subfield of a "mixed const" bundle:

```
wire w: {a: const UInt<8>, b: UInt<8>}
node a = w.a 
```

By the wording, wire w is non-const, and thus node a is also non-const.  I believe w.a should be const.

I'm not sure we needed to update the wording here at all. The type of a sub-field expression is the type of the field that was accessed, same as before.

spec.md
azidar: Maybe it's clearer to use the term "literal", rather than "const" or "constant" ?

spec.md
darthscsi: No, a bundle can be const regardless of its fields.  The wording on indexing operations says that indexing a const bundle will return a const type.

darthscsi: In your example, a would be const since the type of the rhs is const.

spec.md
darthscsi: That is a very interesting question I don't know the answer to.  If firrtl supported initialization, then yes, constant memories would make sense.

spec.md
darthscsi: Yes.  I was a bit undecided what to do here.  I decided that the best interpretation is that `UInt<10>(42)` is a primitive operator with the same name as the type it produces, currently, so it could have it's type changed without change in syntax.  Since const decay to non-const, this doesn't change existing usage.

spec.md
darthscsi: It depends on what "compile time" means.  I'll think about the wording.  But to answer your question, I'm trying to emulate neither per-se.

youngar:
Might have been asked already, but do we need something in type equivalence to say that `const {a : UInt<1>}` is equivalent to `{const a : UInt<1>}`, so that they may be connected? https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#type-equivalence

darthscsi:
> Might have been asked already, but do we need something in type equivalence to say that `const {a : UInt<1>}` is equivalent to `{const a : UInt<1>}`, so that they may be connected? https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#type-equivalence

If you decompose it element-wise, it type checks.  Will have to think about the equivalence rules.

youngar:
Yeah, the type equivalence rules have this line "Two bundle types are equivalent if they have the same number of fields, and both the bundles' i'th fields have matching names and orientations, as well as equivalent types.". Maybe we just need to say that type equivalence ignores const-ness, and to the [connect algorithm](https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#connects) section a fourth bullet that says "The left hand side expression may only be const if the right hand side expression is const"

mwachs5 (approved):
This makes sense to me now, though would liek clarification on @youngar point above

spec.md
```diff
@@ -499,6 +499,55 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
 flows into the module, and the `d`{.firrtl} sub-field contained in the
 `b`{.firrtl} sub-field flows out of the module.
 
+## Type Modifiers
+
+### Constant Type
+
+A constant type is a type whose value is guaranteed to be a compile-time 
+constant.  Constant types may be used in ports, wire, nodes, and generally
```
ekiwi, line 507: "whose value is guaranteed to never change during the execution of the circuit."

or

"whose value is guaranteed to be constant in the final physical circuit"

Since the value can actually change _after_ the firrtl compilation, when you link together multiple Verilog files.

dtzSiFive:
Who is doing the guarantee?  What are the requirements/preconditions for this to be guaranteed?
Right now I think it's something like "we expect this to be const, but it's less a guarantee as much as a user assertion who is also responsible to arrange that to be so in the final circuit/execution environment, it is undefined behavior if that is not the case"?

azidar:
It's buried in a comment, but I feel that "const" and "constant" aren't really the right words, in that they have many different connotations/definitions which are confusing. I much prefer "literal" as the term (and is used as such in the FIRRTL spec) to indicate this "const" property. Can we re-name it to "literal types" ?

azidar (changes requested):
I feel like there are still outstanding questions/comments which need addressing, prior to a major specification change.

seldridge:
> Can we re-name it to "literal types" ?

I don't think "literal" is the right word. Existing literals have constant type (or they may be used in places that require constants), but constant types are not literals (in [the PL sense](https://en.wikipedia.org/wiki/Literal_(computer_programming)) of a fixed value in the language) nor are they the same thing s literals in the existing FIRRTL spec.

Consider the strap pin situation where you need a constant on the port before reset. However, this may not be a literal, but a choice of any literal.

azidar:
@seldridge I disagree. `UInt(0)` is a literal **value**. We can define a literal **type**.

azidar:
To clarify further, it seems like the property we are guaranteeing is in effect "this signal must only be driven by combinational logic whose inputs are always sourced with literal values".

darthscsi:
> Who is doing the guarantee? What are the requirements/preconditions for this to be guaranteed? Right now I think it's something like "we expect this to be const, but it's less a guarantee as much as a user assertion who is also responsible to arrange that to be so in the final circuit/execution environment, it is undefined behavior if that is not the case"?

We can guarantee it within the bounds of the language (firrtl).  We cannot necessarily guarantee the language the compiled circuit  is in has a type system which can both match the constraints from firrtl and produce the desired output.  Which means we cannot guarantee a user will respect the constraint on the port (or whatever).  But this is really just standard compiling problems (imagine a stray library/asm file which felt the need to change the permissions on the bss segment and start rewriting values).  There is only so much you can do to ensure the rest of the system doesn't break the language-imposed constraints.

ekiwi:
> There is only so much you can do to ensure the rest of the system doesn't break the language-imposed constraints.

You can guarantee it if you do not allow inputs to be marked `const`. But I realize that it is desirable for these `const` values to be only known at link time. So how about changing the language to clarify that the value isn't a constant at compile time, but only at link time, since the value might be different every time the module is linked, it just isn't allowed to change kin the actual circuit that is produced once everything is linked together?

dtzWill:
Yeah, you're right there's only so much to be done within a language.  Should've given more context, meant to this morning to be honest.  This thinking came from similar thoughts for, e.g., ref types, where it's very likely more important.  What can we, do we want to, expect from other circuit components we're combined with?  Introducing a larger scope of the build flow (e.g., linker) would definitely help, especially if we optionally want more ambitious/creative codegen approaches there-- effectively some sort of ABI, const is similar re:imposing more restrictions than verilog interface conveys, but is much more easily understood and handled by a user, "have value available before any portion of any FIRRTL-based circuit and it can't change" or whatever is right (I am unsure).

But since mentioning a (FIRRTL/Verilog) linker seemed a little out of scope for this PR, I'm suggesting we for now write down and be clear about what we expect from the user/circuit/etc. that we rely on but don't handle or check.  And be clear what is and is not guaranteed, otherwise it does seem we claim to guarantee properties that we do not vs places users can assert they've handled the prerequisites for.

(As you know) Other languages usually provide for much of this as part of their implementation strategy if not in the language spec although it's usually documented somewhere.  The bss example seems more like "what if someone hooks up probes to what we do guarantee as const but changes it (or disconnects wires, etc.)".  To me this isn't the same as, "What needs to be true for the guaranteed language semantics to hold" (that we expect the user to be aware of and handle, but could easily be missed or handled incorrectly, and aren't true by default if the user does not take action) but you're right there's a line to draw on this.

I think if we clarified what we do/don't guarantee re:input ports that'd go a long way, especially if additionally we do clarify what precisely we mean by const so someone interested in their circuit working knows what is needed.  I think we could say we guarantee const across component boundaries if both-sides/source-side is built from FIRRTL?

spec.md
darthscsi: Examples added.

spec.md
darthscsi: Wording hopefully clarified as to what is meant by "constant".

spec.md
darthscsi: Hopefully clarified.

spec.md
darthscsi: Went ahead and defined const memories as ROMs and no write-ports allowed.  Punted on how you get the data into the ROM.

spec.md
```diff
@@ -722,7 +772,8 @@ the current value of the element, writes are not visible until after a positive
 edges of the register's clock port.
 
 The clock signal for a register must be of type `Clock`{.firrtl}.  The type of a
-register must be a passive type (see [@sec:passive-types]).
+register must be a passive type (see [@sec:passive-types]) and may not be 
+`const`.{.firrtl}.
```
mwachs5, line 776:
```suggestion
`const`{.firrtl}.
```

spec.md
```diff
@@ -499,6 +499,56 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
...
+The indexing of a constant aggregate produces a constant of the appropriate type 
+for the element.
+
+#### A note on implementation
+
+Constant types are a restriction on firrtl types.  Therefore, firrtl structures 
+which would be expected to produce certain verilog structures will produce the
+same structure if instantiated with a constant type.  For example, an input port
+of type `const UInt` will result in a port in the verilog, if under the same
+conditions an input port of type `UInt` would have.
```
dtzSiFive, line 544: Hmm, how do we ensure this? I know it's an implementation detail but is it now an assertion/guarantee that no optimization or composition of them cannot discover a port to be dead due to const-ness that would not otherwise?

Don't have a complete example, but seems possible to have something like: a register initialized+assigned from a constant input that we could optimize to that input value when we know it's constant but cannot if the value may change (where we need to dutifully store the latest value at clock edge or so).  Or maybe just the "has this changed since initialization" sort of module where the output is a comparison of the register to the input signal it's initialized with?

dtzSiFive (commented):
Some nits and a question about avoiding const-ness resulting in discovering a port is dead when it wouldn't otherwise.

spec.md
```diff
@@ -499,6 +499,56 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
...
+
+A constant type is a type whose value is guaranteed to be unchanging at circuit 
+execution time.  Constant is a constraint on the mutability of the value, it 
+does not imply a literal value at a point in the emitted design.  Constant types 
+may be used in ports, wire, nodes, and generally anywhere a non-constant type is 
+usable.  Operations on constant type are well defined.  As a general rule (with 
+any exception listed in the definition for such operations as have exceptions), 
+an operation whose arguments are constant produces a constant.  An operations 
+with some non-constant arguments produce a non-constant.  Constants can be used 
+in any context with a source flow which allows a non-constant.  Constants may be
```
trilorez, line 514: I think this sentence covers it, but is it worth elaborating on how `const` composes with `flip`? 
For example, I believe this should work since flow is from `const` to non-`const`:
```firrtl
module FlippedConst:
    input in: {flip x: UInt<1>}
    output out: const {flip x: UInt<1>}
    out <= in
```

darthscsi: It should.

spec.md
```diff
@@ -499,6 +499,56 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
...
+in any context with a source flow which allows a non-constant.  Constants may be 
+used as the target of a connect so long as the source of the connect is itself 
+constant.  These rules ensure all constants are derived from literals or from 
+constant-typed input ports of the top-level module.
+
+``` firrtl
+const UInt<3>
+const SInt
+const {real: UInt<32>, imag : UInt<32>, other : const SInt}
+const
```
seldridge, line 523: What is a bare `const`?

spec.md
```diff
@@ -499,6 +499,56 @@ the module.  The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field
...
+constant-typed input ports of the top-level module.
+
+``` firrtl
+const UInt<3>
+const SInt
+const {real: UInt<32>, imag : UInt<32>, other : const SInt}
+const 
+```
+
+Last-connect semantics of constant typed values is well defined, so long as any
```
seldridge, line 526: Nit:
```suggestion
Last-connect semantics of constant typed values are well defined, so long as any 
```

spec.md
```diff
@@ -2649,7 +2747,8 @@ type_ground = "Clock" | "Reset" | "AsyncReset"
 type_aggregate = "{" , field , { field } , "}"
                | type , "[" , int , "]" ;
 field = [ "flip" ] , id , ":" , type ;
-type = type_ground | type_aggregate ;
+consttype = "const", type_ground | type_aggregate
+type = type_ground | type_aggregate | consttype ;
```
seldridge, line 2751: Nit: This can be simplified to the following if `consttype` is not intended to be used elsewhere (and I don't think it makes sense to try to put the `AsyncReset` reg _semantic_ into EBNF which is the only place I can think of where `consttype` could be used):

```suggestion
type = [ "const" ] , ( type_ground | type_aggregate );
```

seldridge (approved):
Happy to see this come together!

A few comments throughout. This LGTM.

spec.md
```diff
@@ -1318,6 +1369,12 @@ data that is read out of the read port is undefined.
 In all cases, if a memory location is written to by more than one port on the
 same cycle, the stored value is undefined.
 
+### Constant memory type
+
+A memory with a constant data-type represents a ROM and may not have 
+write-ports.  It is beyond the scope of this specification how ROMs are 
+initialized.
```
azidar, line 1376: Is it worth outlawing constant memories until we have a mechanism to initialize them?

spec.md
```diff
@@ -1665,15 +1749,26 @@ module MyModule :
...
-unsigned integer type.  An access to an out-of-bounds element results in an 
-indeterminate value (see [@sec:indeterminate-values]).  Each out-of-bounds 
-element is a different indeterminate value.  Sub-access operations with constant
-index may be convereted to sub-index operations even though it converts
-indeterminate-value-on-out-of-bounds behavior to a compile-time error.
+unsigned integer type.  If the expression is of a constant vector type, the 
+sub-element shall be of a constant type.  An access to an out-of-bounds element 
+results in an indeterminate value (see [@sec:indeterminate-values]).  Each 
+out-of-bounds element is a different indeterminate value.  Sub-access operations 
+with constant index may be convereted to sub-index operations even though it
```
azidar, line 1770:
```suggestion
with constant index may be converted to sub-index operations even though it 
```

spec.md
```diff
@@ -1665,15 +1749,26 @@ module MyModule :
...
 The sub-access expression dynamically refers to a sub-element of a vector-typed
 expression using a calculated index. The index must be an expression with an
-unsigned integer type.  An access to an out-of-bounds element results in an 
-indeterminate value (see [@sec:indeterminate-values]).  Each out-of-bounds 
-element is a different indeterminate value.  Sub-access operations with constant
-index may be convereted to sub-index operations even though it converts
-indeterminate-value-on-out-of-bounds behavior to a compile-time error.
+unsigned integer type.  If the expression is of a constant vector type, the 
+sub-element shall be of a constant type.  An access to an out-of-bounds element 
+results in an indeterminate value (see [@sec:indeterminate-values]).  Each
```
azidar, line 1768: Is this saying that the indeterminate value is `const` ?

spec.md
darthscsi: The spec doesn't give a way to initialize non-constant memories either.   I suspect in the near term, the existing defacto annotation or $readmem* intrinsic would work until all memories have a uniform initialization method available in the spec.

azidar (approved):
Looks great! Just please fix the minor clarifications prior to merging. I personally feel we should make memories of const-type illegal until we actually have a complete story for ROMs, but I don't have a super strong opinion about it.

spec.md
azidar: I'm not sure that the `mem` construct is actually what we want to represent a ROM though. To me, it seems much more akin to an aggregate literal expression/construct, than a memory. In that regard, initialization of a memory is very different than a ROM, as the initialization of a ROM is not programmable and thus is synthesizable.

spec.md
darthscsi: This is a good question.  I believe it works as one expects.  An out-of-bounds access would be an indeterminate value of const type.  In practice, it's either going to be 0 or one of the in-bound inputs, both of which are const in this scenario.

PR #70, opened 2023-02-09
[patch] Clarify lowering behavior regarding ports (naming, orientation).

dtzSiFive:
Mostly clarifying output bundles with flips become input ports,
but while visiting address some naming details.


spec.md
```diff
@@ -2603,6 +2605,25 @@ wire a_1_b : UInt<1>
...
+
+When lowered this becomes:
+
+```firrtl
+extmodule Test:
+  input x_a : UInt<1>
+  output x_b : UInt<1>
+```
+
+It is implementation-defined how name conflicts are resolved within or across
```
mwachs5, line 2624: does this have to be an implementation detail? Can it be an error?

darthscsi: For internal deployment, it seems simple enough to opt-in to the stricter semantics for a while.  Since names of ports are known up front, we can check the names very early on and provide clear errors.

That said, it does seem like lower-types shouldn't affect what is legal.  It also seems like all legal things should be deterministically lowered, especially since this is mixing with hand-written code.

We haven't kept the previous rules well, but IIRC, if lowering a port made a name conflict with an existing port, a `_` was added to all names being lowered from a port until no conflict existed.  Thus your example would become:
```
a_b : UInt<1>, a__b : UInt<2>
```

dtzSiFive:
This appears to be subsumed by #76, with a nicer explicit notion of "convention" and details of naming (:tada:).

PR #71, opened 2023-02-09
[patch] Drop language about width of conditionally valid.

dtzSiFive:
Condtionally valid (`validif`) was removed in 9890f17f2d .


seldridge (approved):

mwachs5 (commented):
nit I'd say this is `patch` not `nfc`. `nfc` is like "I changed something in the Makefile on how we generate the PDFs". You are actually changing the words of the spec.

dtzSiFive:
I completely agree.  Since this was just a cleanup to an existing 'major' change not in a release yet I was trying to sneak it under that umbrella :innocent: , but this sure isn't NFC.  Will update shortly.

PR #72, opened 2023-02-27
[minor] Intrinsics via modules

darthscsi:
Reuse the machinery of modules for intrinsics.  This allows a pretty wide range of intrinisics to be defined.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -242,6 +242,29 @@ A common use of an externally defined module is to represent a Verilog module
 that will be written separately and provided together with FIRRTL-generated
 Verilog to downstream tools.
 
+## Implementation Defined Modules (Intrinsics)
+
+Intrinsic modules are modules which represent implementation-defined, 
+compiler-provided functionality.  Intrinsics generally are used for 
+funcitonality which requires knowledge of the implementation or circuit not
+available to a library writer.  What intrinsics are supported by an 
+implementation is defined by the implementation.  The particular intrinsic
+represented by an intrinsic module is encoded in _intrinsic_.  The name of the
+intmodule is only used for instances, it has no semantic meaning.  An 
+implementation shall type-check all ports and parameters.  Ports may be 
+uninferred (either width or reset) if specified by the implementation (which is 
+useful inspecting and interacting with those inference features).
+
+``` firrtl
+intmodule MyExternalModule :
+  input foo: UInt<2>
+  output bar: UInt<4>
+  output baz: SInt<8>
+  intrinsic = IntrinsicName
+  parameter x = "hello"
+  parameter y = 42
+```
+
 # Types
 
 FIRRTL has two classes of types: _ground_ types and _aggregate_ types.  Ground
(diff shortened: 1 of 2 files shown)


ekiwi (approved):
I like this a lot better than treating intrinsics as functions (which requires you to define a whole bunch of machinery that does not exist yet)

spec.md
```diff
@@ -242,6 +242,29 @@ A common use of an externally defined module is to represent a Verilog module
...
+## Implementation Defined Modules (Intrinsics)
+
+Intrinsic modules are modules which represent implementation-defined, 
+compiler-provided functionality.  Intrinsics generally are used for 
+funcitonality which requires knowledge of the implementation or circuit not
+available to a library writer.  What intrinsics are supported by an 
+implementation is defined by the implementation.  The particular intrinsic
+represented by an intrinsic module is encoded in _intrinsic_.  The name of the
+intmodule is only used for instances, it has no semantic meaning.  An 
+implementation shall type-check all ports and parameters.  Ports may be
```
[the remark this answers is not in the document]
darthscsi: I can't make parametric types for ports.  Hence we can't make a single module which captures the possible types.  TL;DR; temporary until actual parameters exist.

spec.md
```diff
@@ -2728,6 +2751,11 @@ extmodule = "extmodule" , id , ":" , [ info ] , newline , indent ,
               [ "defname" , "=" , id , newline ] ,
               { "parameter" , id , "=" , ( string | int ) , newline } ,
             dedent ;
+intmodule = "intmodule" , id , ":" , [ info ] , newline , indent ,
+              { port , newline } ,
+              [ "intrinsic" , "=" , id , newline ] ,
+              { "parameter" , "=" , ( string | int ) , newline } ,
+            dedent ;
```
seldridge, line 2758:
```suggestion
intmodule = "intmodule" , id , ":" , [ info ] , newline , indent ,
              { port , newline } ,
              "intrinsic" , "=" , id , newline ,
              { "parameter" , "=" , ( string | int ) , newline } ,
            dedent ;
```

As currently formulate, I think the `intrinsic` is mandatory as opposed to optional?

spec.md
```diff
@@ -242,6 +242,29 @@ A common use of an externally defined module is to represent a Verilog module
...
+useful for inspecting and interacting with those inference features).
+
+``` firrtl
+intmodule MyIntrinsicModule :
+  input foo: UInt
+  output bar: UInt<4>
+  output baz: SInt<8>
+  intrinsic = IntrinsicName
+  parameter x = "hello"
+  parameter y = 42
```
seldridge, line 265: For clarity, it may be good to call this `intmodule IntrinsicModule_xhello_y42` or something that helps get the point across that this is a name used to uniquely identify an instantiation.

spec.md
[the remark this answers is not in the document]
seldridge: Consider: "The name of the intmodule is only used to identify a specific instantiation." You could also consider some language that indicates that this is a limitation of lack of parameters and is expected to change in the future. However, I'm not sure on the wording and that type of telegraphing isn't appropriate for the spec.

seldridge (approved):
Awesome. I really, really like this. 

I am particularly excited because this can entirely subsume memories with a simple, general primitive.

spec.md
darthscsi: yes

PR #74, opened 2023-02-28
[patch] Fix missing ports in indeterminate example

azidar:
My assumption is that the examples provided are missing the `input v` port that was defined in the earlier example. Otherwise, `v` is undefined in the circuit.


dtzSiFive (approved):
LGTM, thanks!

PR #77, opened 2023-03-03
[patch] Add syntax support for type modifiers (e.g., const).

dtzSiFive:
Fix rendering const types in PDF.

### DIFF
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -28,6 +28,9 @@
       <item>Reset</item>
       <item>AsyncReset</item>
     </list>
+    <list name="typemodifiers">
+      <item>const</item>
+    </list>
     <list name="conditionals">
       <item>when</item>
       <item>else</item>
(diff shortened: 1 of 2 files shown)


include/firrtl.xml
```diff
@@ -128,6 +132,10 @@
       <context name="info" attribute="Info" lineEndContext="#stay">
         <DetectChar char="]" context="#pop"/>
       </context>
+      <context name="typemodifier" attribute="Keyword" lineEndContext="#pop">
+        <keyword String="types" context="#pop!type" attribute="Keyword"/>
+        <DetectChar char="{" context="#pop!field" attribute="Separator"/>
+      </context>
```
seldridge, line 138: This state transition seems a bit odd as it is dually encoding that `types` -> `types` and `{` -> `field`. Is it simpler to do something like popping on any whitespace that shows up?

This could probably work just like existing keywords which is just a `#stay`.

WDYT?

dtzSiFive: Just using `#stay` in the default and field contexts should do the trick for valid FIRRTL, indeed.  This is a little more specific, but probably doesn't help us any, so let's go with simple.  Thanks!

dtzSiFive: (updated accordingly)

seldridge (approved):

PR #79, opened 2023-03-03
[nfc] Fix more unstyled FIRRTL snippets

dtzSiFive:
Leave single-chars as-is for now.


seldridge (approved):
LGTM

Feel free to land these types of nfc changes without review.

PR #80, opened 2023-03-03
[patch] Render inline annotations as JSON, fix typo in example

dtzSiFive:
Apparently we can bounce to other syntax definitions, use that to make inline annotations render nicely.

Also fix unstyled inline anno while visiting.

Modeled this after https://github.com/KDE/syntax-highlighting/blob/2880570764a754bdf2ed402d6fd24024a35bf5bc/data/syntax/markdown.xml#L421 , seems to work for our use case.


seldridge (approved):
Thank you for doing this!

I had read about it in the documentation, but never did it. This is a big readability benefit for the inline format. 💯

jackkoenig (approved):

PR #86, opened 2023-03-20
[patch] Clarify Int/String Types, clarify SE-Ints

seldridge:
Add language that explicitly defines literal types:

  1. Integer literals
  2. String literals
  3. Raw String literals
  4. String-encoded integer literals

This only clarifies existing behavior which was left ambiguous or documented in different locations in the spec.

Define that string-encoded integer literals can be used any place that an integer literal can except where specified otherwise.  This is used to differentiate where an external module parameter that looks like a string-encoded integer literal is still a string.  This is also used to preserve the existing SFC behavior where a string-encoded integer literal is not allowed to be used for memory integers.

Notably, this is enshrining the fact that you can do things today like:

```
circuit Foo:
  module Foo:
    input clock: Clock
    input a: UInt<"hF">["o10"]
    output b: UInt<"b1111">[8]

    b <= a
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -224,7 +224,15 @@ references.  Each name--value parameter statement will result in a value being
 passed to the named parameter in the resulting Verilog.  Every port or port
 sub-element of reference type must have exactly one `ref`{.firrtl} statement.
 
-An example of an externally defined module is:
+The widths of all externally defined module ports must be specified.  Width
+inference, described in [@sec:width-inference], is not supported for externally
+defined module ports.
+
+A common use of an externally defined module is to represent a Verilog module
+that will be written separately and provided together with FIRRTL-generated
+Verilog to downstream tools.
+
+An example of an externally defined module with parameters is:
 
 ``` firrtl
 extmodule MyExternalModule :
@@ -235,14 +243,6 @@ extmodule MyExternalModule :
   parameter x = "hello"
   parameter y = 42
 ```
-
-The widths of all externally defined module ports must be specified.  Width
-inference, described in [@sec:width-inference], is not supported for externally
-defined module ports.
-
-A common use of an externally defined module is to represent a Verilog module
-that will be written separately and provided together with FIRRTL-generated
-Verilog to downstream tools.
 
 An example of an externally defined module with references is:
 
@@ -259,6 +259,38 @@ current FIRRTL design. While they are part of the FIRRTL-level interface to the
 external module, they are not expected to correspond to a particular Verilog
 construct. They exist to carry information about the implementation of the
 extmodule necessary for code generation of the current circuit.
+
+The types of parameters may be any of the following literal types.  See
+[@sec:literals] for more information:
+
+1. Integer literal, e.g. `42`{.firrtl}
+1. String literal, e.g., `"hello"`{.firrtl}
+1. Raw String Literal, e.g., `'world'`{.firrtl}
+
+An integer literal is lowered to a Verilog literal.  A string literal is lowered
+to a Verilog string.  A raw string literal is lowered verbatim to Verilog.
+
+As an example, consider the following external module:
+
+``` firrtl
+extmodule Foo:
+  parameter foo = 'hello'
+  parameter bar = "world"
+  parameter baz = 42
+```
+
+This is lowered to a Verilog instantiation site as:
+
+``` verilog
+Foo #(
+  .foo(hello),
+  .bar("world")
+  .baz(42)
+) bar();
+```
+
+A parameter that is a string-encoded integer literal is treated as a string
+literal.
 
 ## Implementation Defined Modules (Intrinsics)
 
@@ -282,6 +314,69 @@ intmodule MyIntrinsicModule_xhello_y64 :
   parameter x = "hello"
   parameter y = 42
 ```
+
+The types of intrinsic module parameters may only be literal integers or
+string literals.
+
+A parameter that is a string-encoded integer literal is treated as a string
+literal.
+
+# Literals
+
+FIRRTL has both integer, string, and raw string literals.
+
+An integer literal is a signed or unsigned decimal integer.  The following are
+examples of integer literals:
+
+``` firrtl
+42
+-9000
+```
+
+A string literal is a sequence of characters with a leading `"`{.firrtl} and a
+trailing `"`{.firrtl}.  The following is an example of a string literal:
+
+``` firrtl
+"hello"
+```
+
+A raw string literal is a sequence of characters with a leading `'`{.firrtl} and
+a trailing `'`{.firrtl}.  The following is an example of a raw string literal:
+
+``` firrtl
+'world'
+```
+
+## String-encoded Integer Literal
+
+A string-encoded integer literal is a special string literal with one of the
+following leading characters to indicate the numerical encoding:
+
+- `b`{.firrtl} -- for representing binary numbers
+- `o`{.firrtl} -- for representing octal numbers
+- `h`{.firrtl} -- for representing hexadecimal numbers
+
+Signed string-encoded integer literals have their sign following the leading
+encoding character.
+
+The following string-encoded integer literals all have the value `42`:
+
+``` firrtl
+"b101010"
+"o52"
+"h2a"
+```
+
+The following string-encoded integer literals all have the value `-42`:
+
+``` firrtl
+"b-101010"
+"o-52"
+"h-2a"
+```
+
+String-encoded integer literals are usable in any place where an integer would
+be unless explicitly disallowed.
 
 # Types
 
@@ -842,8 +937,8 @@ an operation whose arguments are constant produces a constant.  An operation
 with some non-constant arguments produce a non-constant.  Constants can be used
 in any context with a source flow which allows a non-constant.  Constants may be
 used as the target of a connect so long as the source of the connect is itself
-constant.  These rules ensure all constants are derived from literals or from
-constant-typed input ports of the top-level module.
+constant.  These rules ensure all constants are derived from constant integer
+expressions or from constant-typed input ports of the top-level module.
 
 ``` firrtl
 const UInt<3>
@@ -1543,21 +1638,25 @@ by the following parameters.
 
 1.  A passive type representing the type of each element in the memory.
 
-2.  A positive integer representing the number of elements in the memory.
+2.  A positive integer literal representing the number of elements in the
+    memory.
 
 3.  A variable number of named ports, each being a read port, a write port, or
     readwrite port.
 
-4.  A non-negative integer indicating the read latency, which is the number of
-    cycles after setting the port's read address before the corresponding
-    element's value can be read from the port's data field.
+4.  A non-negative integer literal indicating the read latency, which is the
+    number of cycles after setting the port's read address before the
+    corresponding element's value can be read from the port's data field.
 
-5.  A positive integer indicating the write latency, which is the number of
-    cycles after setting the port's write address and data before the
+5.  A positive integer literal indicating the write latency, which is the number
+    of cycles after setting the port's write address and data before the
     corresponding element within the memory holds the new value.
 
 6.  A read-under-write flag indicating the behavior when a memory location is
     written to while a read to that location is in progress.
+
+Integer literals for the number of elements and the read/write latencies _may
+not be string-encoded integer literals_.
 
 The following example demonstrates instantiating a memory containing 256 complex
 numbers, each with 16-bit signed integer fields for its real and imaginary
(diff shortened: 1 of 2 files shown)


azidar:
This was added to the spec by MFC, this was never supported in SFC if I remember correctly, at least in the original FIRRTL spec this was never the case. Personally, I don't like that one can do `UInt<"hF">` but :shrug: at least its good to clarify existing behavior.

spec.md
```diff
@@ -283,6 +293,69 @@ intmodule MyIntrinsicModule_xhello_y64 :
   parameter y = 42
 ```
 
+The types of intrinsic module parameters may only be literal integers or
+string literals.
+
+A parameter that is a string-encoded integer literal is treated as a string
+literal.
```
azidar, line 300: This is why I don't like this. `Integers` should be `42`, not `"hFFFF"`. The fact that this is now a corner case here is just bad. It should accept strings. The ONLY place that it makes sense to pass in a hex number is when a user cares about a hardware literal value, hence, it should only be special cased in the grammar there.

spec.md
```diff
@@ -283,6 +293,69 @@ intmodule MyIntrinsicModule_xhello_y64 :
...
+A raw string literal is a sequence of characters with a leading `'`{.firrtl} and
+a trailing `'`{.firrtl}.  The following is an example of a raw string literal:
+
+``` firrtl
+'world'
+```
+
+## String-encoded Integer Literal
+
+A string-encoded integer literal is a special string literal with one of the
```
azidar, line 330: Is it a string literal or an integer literal?

spec.md
```diff
@@ -3478,11 +3502,11 @@ linecol = digit_dec , { digit_dec } , ":" , digit_dec , { digit_dec } ;
 info = "@" , "[" , { string , " " , linecol } , "]" ;
 
 (* Type definitions *)
-width = "<" , int , ">" ;
+width = "<" , int_any , ">" ;
```
azidar, line 3505: Can we just not allow this?

spec.md
```diff
@@ -3456,11 +3470,21 @@ digit_hex = digit_dec
...
+    '"' , "b" , [ "-" ] , { digit_bin } , '"'
+  | '"' , "o" , [ "-" ] , { digit_oct } , '"'
+  | '"' , "h" , [ "-" ] , { digit_hex } , '"' ;
+
+(* Any Integer Literal *)
+int_any = int | int_se ;
+
+(* String Literals *)
+string = (* a string that is not defined and assumed to exist *) ;
+string_raw = "'" , string , "'" ;
```
azidar, line 3487: We should make `int_se` be a string in the grammar.

spec.md
```diff
@@ -3456,11 +3470,21 @@ digit_hex = digit_dec
...
+int_se =
+    '"' , "b" , [ "-" ] , { digit_bin } , '"'
+  | '"' , "o" , [ "-" ] , { digit_oct } , '"'
+  | '"' , "h" , [ "-" ] , { digit_hex } , '"' ;
+
+(* Any Integer Literal *)
+int_any = int | int_se ;
+
+(* String Literals *)
+string = (* a string that is not defined and assumed to exist *) ;
```
azidar, line 3486: This is weird to me.. can you clarify?

spec.md
mwachs5: by "it should accept strings" do you mean "it should not accept strings"?

spec.md
```diff
@@ -2221,71 +2298,49 @@ for referring to a declared circuit component, for statically and dynamically
...
+UInt<10>("b101010")
+UInt<10>("o52")
+UInt<10>("h2A")
+UInt<10>("h2a")
 ```
 
 Note that it is an error to supply a bit width that is not large enough to fit
 the given value. If the bit width is omitted, then the minimum number of bits
-necessary to fit the given value will be inferred.
+necessary to fit the given value will be inferred.  Both of the following will
```
mwachs5, line 2319:
```suggestion
necessary to fit the given value will be inferred.  All of the following will
```

spec.md
```diff
@@ -2295,34 +2350,12 @@ be inferred.
 
 ``` firrtl
 SInt(-42)
+SInt("b-101010")
```
mwachs5, line 2353: is there supposed to be some words above that say something about the width of these, or what these are all showing? It seems like these are presented without explanation, unlike the integer version of these which states that the bit width will be 5 for all of them

spec.md
```diff
@@ -3574,17 +3598,18 @@ module = "module" , id , ":" , [ info ] , newline , indent ,
            { port , newline } ,
            { statement , newline } ,
          dedent ;
+type_param = int | string | raw_string ;
```
mwachs5, line 3601: i don't get the difference between string and raw_string other than how they are emitted Didn't we also get rid of the RawString version of this from chisel in terms of the ExtModule use of it? Can we just get rid of raw_string? Is the only point of it to let you put `"` inside the string...?

spec.md
seldridge: I don't know how to define "string" nor do I want to. Other places that I've seen this, e.g., in the Python grammar, they will use something external, like `STRING`. What I'm trying to communicate here is that a person reading this needs to assume the existence of a string and/or this is not specified within the scope of this document.

seldridge:
> This was added to the spec by MFC, this was never supported in SFC if I remember correctly, at least in the original FIRRTL spec this was never the case. Personally, I don't like that one can do UInt<"hF"> but 🤷 at least its good to clarify existing behavior.

SFC compiles the above circuit, too.

```
# cat Foo.fir && firrtl -i Foo.fir && cat Foo.v
circuit Foo:
  module Foo:
    input clock: Clock
    input a: UInt<"hF">["o10"]
    output b: UInt<"b1111">[8]

    b <= a
module Foo(
  input         clock,
  input  [14:0] a_0,
  input  [14:0] a_1,
  input  [14:0] a_2,
  input  [14:0] a_3,
  input  [14:0] a_4,
  input  [14:0] a_5,
  input  [14:0] a_6,
  input  [14:0] a_7,
  output [14:0] b_0,
  output [14:0] b_1,
  output [14:0] b_2,
  output [14:0] b_3,
  output [14:0] b_4,
  output [14:0] b_5,
  output [14:0] b_6,
  output [14:0] b_7
);
  assign b_0 = a_0;
  assign b_1 = a_1;
  assign b_2 = a_2;
  assign b_3 = a_3;
  assign b_4 = a_4;
  assign b_5 = a_5;
  assign b_6 = a_6;
  assign b_7 = a_7;
endmodule
```

I admit that this was likely never the intended behavior, but this reflects the current state of the world _for the SFC_. MFC deviates only in allowing string-encoded integer literals in memories.

I am also amenable to just disallowing this entirely and doing a major spec change that only allows string-encoded integer literals in the construction of hardware integer literals. This makes way more sense to me.

spec.md
seldridge: Here's a small circuit:

```
circuit Foo:
  extmodule Bar:
    output a: UInt<1>
    parameter foo = 'hello'
    parameter bar = "world"

  module Foo:
    output a: UInt<1>

    inst bar of Bar
    a <= bar.a
```

This compiles to Verilog as:
```
```

spec.md
seldridge: Will fix in a rebase.

spec.md
seldridge: I'm on the fence on this... on the one hand, I think it is super useful to be able to put in the grammar what is a legal string-encoded literal is. A reader can grok what this is without having to wonder, "What strings can I put in there?" On the other hand, this very much blurs the lines between syntax (what the grammar should be) and semantics. I tend to fall on the former side of this specific case.

spec.md
seldridge: I would like to! I'm staging a follow-on major change to do this.

seldridge:
Addressed comments and created a follow-on major change that only allows for string-encoded literals to be used by hardware literals: https://github.com/chipsalliance/firrtl-spec/pull/87

I would like to proceed by landing this patch change (documenting the existing behavior of SFC and MFC) and then immediately landing the major change to greatly restrict all the unintended uses of string-encoded literals.

WDYT?

spec.md
azidar: I think my point is that "string-encoded-integers" should be treated in the grammar as a 'subtype' of String. But, I'm actually in favor of the proposal of removing them altogether and just providing more syntaxes for declaring integers from different bases. E.g. h-FFFF or whatever.

spec.md
mwachs5: stupid question, why not actually define this? Do we allow unicode characters? Snowmen and smiley faces? can we point to some external definition of a string? How to escape 'quotes' in the middle of strings? It seems like there would probably be some standard string definitions to use?

spec.md
mwachs5: ok. but again i thought we got rid of the RawStringParam Chisel API so is there still a use for this?

spec.md
```diff
@@ -3456,11 +3471,21 @@ digit_hex = digit_dec
...
+int = [ "-" ] , { digit_bin } ;
+
+(* String-encoded Integer Literals *)
+int_se =
+    '"' , "b" , [ "-" ] , { digit_bin } , '"'
+  | '"' , "o" , [ "-" ] , { digit_oct } , '"'
+  | '"' , "h" , [ "-" ] , { digit_hex } , '"' ;
+
+(* Any Integer Literal *)
+int_any = int | int_se ;
```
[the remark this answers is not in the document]
seldridge: Updated the comment in the !fixup commit.

spec.md
darthscsi: Alternately the user should never be reading firrtl, so it doesn't matter if firrtl doesn't support hex, that has nothing to do with whether they can write hex numbers in chisel.

spec.md
```diff
@@ -283,6 +293,69 @@ intmodule MyIntrinsicModule_xhello_y64 :
...
+``` firrtl
+"b101010"
+"o52"
+"h2a"
+```
+
+The following string-encoded integer literals all have the value `-42`:
+
+``` firrtl
+"b-101010"
```
darthscsi, line 351: I'd rather it be -b101010.  Then at least it looks like negation of an integer literal rather than an identifier minus an interger literal.

Is this supported now or is this being added by this?

spec.md
```diff
@@ -2221,108 +2298,65 @@ for referring to a declared circuit component, for statically and dynamically
 accessing a nested element within a component, for creating multiplexers, for
 performing primitive operations, and for reading a remote reference to a probe.
 
-## Unsigned Integers
+## Unsigned Hardware Integers
```
darthscsi, line 2301: "Consant Interger Expressions"?  In my mind `UIint...` is an expression (operator) which produces a value of unsigned integer type from a syntatic integer literal.

spec.md
```diff
@@ -2221,108 +2298,65 @@ for referring to a declared circuit component, for statically and dynamically
...
+A literal unsigned hardware integer can be created given a non-negative integer
+literal or non-negative string-encoded integer literal.  An optional positive
+bit width may be specified. Hardware integer literals are of constant type.  All
+of the following examples create a 10-bit unsigned hardware integer representing
+the number `42`:
 
 ``` firrtl
 UInt<10>(42)
+UInt<10>("b101010")
+UInt<10>("o52")
```
darthscsi, line 2312: the quotes hurt my soul.

spec.md
```diff
@@ -2221,108 +2298,65 @@ for referring to a declared circuit component, for statically and dynamically
...
-## Unsigned Integers from Literal Bits
-
-A literal unsigned integer can alternatively be created given a string
-representing its bit representation and an optional bit width.  Like with the
-integer representation, the expression has a constant type.
-
-The following radices are supported:
-
-1.`b`{.firrtl} : For representing binary numbers.
+## Signed Hardware Integers
```
darthscsi, line 2330: dito as UInt

darthscsi (commented):
It's not clear that there is any reason for more than one integer representation.  The IR representation has no bearing on what the user is allowed to write in Chisel.  Having more forms just makes parsing harder for no good reason.  Having integers and strings alias and only be resolvable by context is bad too.  These aren't arguing against documenting existing behavior, but arguing to at minimum have an issue tracking normalizing these things.

spec.md
seldridge: I agree. This is supported now, though. 🥲

spec.md
```diff
@@ -283,6 +293,69 @@ intmodule MyIntrinsicModule_xhello_y64 :
...
+```
+
+A string literal is a sequence of characters with a leading `"`{.firrtl} and a
+trailing `"`{.firrtl}.  The following is an example of a string literal:
+
+``` firrtl
+"hello"
+```
+
+A raw string literal is a sequence of characters with a leading `'`{.firrtl} and
```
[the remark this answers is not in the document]
seldridge: I'll add some information about how this is lowered to Verilog in a `!fixup` commit.

spec.md
seldridge: This is the language that I was looking for! Thanks! These are expressions and that helps clarify everything.

seldridge: I talked with Adam a bit about this and he was more in favor of `UInt Literal` and `SInt Literal`. I started to write this and then switched back to `Constant Integer Literal`. The reason for this is that `UInt` and `SInt` are always referred to, in prose, as "unsigned integer type" or "signed integer type". I'd like to be able to keep the following distinctions to avoid collisions with "integer literal" which seems natural to mean `42` and not `UInt(42)`. The current language is as follows:

- "integer literal", e.g., 42
- "unsigned integer type", e.g., `UInt<8>`
- "constant integer expression", e.g., `UInt<8>(42)`

WDYT?

seldridge:
> These aren't arguing against documenting existing behavior, but arguing to at minimum have an issue tracking normalizing these things.

Tracking issue opened here.

Issue #89, opened 2023-03-23
Cleanup String-encoded Literals


seldridge:
The FIRRTL spec currently allows for string-encoded integer literals, e.g., `"hDEADBEEF"`, to be used in almost any location where an integer is accepted. Because this has been how the SFC has worked since its inception, this is going to be enshrined int he spec here: https://github.com/chipsalliance/firrtl-spec/pull/86 

The current rule is (summarizing):

> A string-encoded integer literal can be used any place an integer literal can be except when defining memory depths and memory read/write latencies. Additionally, string-encoded integer literals are treated as strings in external module and intrinsic module parameters (as there is no way to differentiate them from normal strings).

This allows for circuits like the following to be accepted by the SFC and MFC:

```
circuit Foo:
  module Foo:
    input clock: Clock
    input a: UInt<"hF">["o10"]
    output b: UInt<"b1111">[8]

    b <= a
```

However, this was never the intent of the original spec. (This actually stems from an incorrect implementation of the spec in the original FIRRTL grammar.) Additionally, there is concern that having multiple ways to represent numbers is not useful for a language like FIRRTL. There are several options that we can take here:

1. Leave everything as it.
2. Expand the usage of string-encoded integer literals to be used by memories. This would bring the spec in alignment with itself while leaving external module parameters as a corner case. This isn't terrible, but it continues to allow for interesting circuits like the one above.
3. Restrict string-encoded integer literals to only being used in the construction of constant integer expressions, e.g., `UInt("hBADF00D")`. This would then disallow their usage in widths, sub-indices, etc. This brings the spec inline with its original intent.
4. Remove all string-encoded integer literals. This would force `UInt("hBADF00D")` to be represented as `UInt(195948557)`. This is less user-friendly, but is fine for a compiler. This also creates better round-trip behavior for compilers. I.e., it would be good if FIRRTL compilers could have one internal representation for constant integer expressions. However, if a compiler does this and it ingests "hBADF00D", can it then emit 195948557.

I'm of the opinion that we should just take option (4) and be done with it. I think the overhead of having more than one number representation isn't particularly useful. As a slight weakening of this, I would be okay if the numbers for constant expressions were only allowed to be hex. I do think it is a great thing if a compiler doesn't have to track additional state in order to round trip IR.

Other related options to consider:

- Is it useful to have octal and binary representations in addition to hex? I'm of the opinion that no, these aren't useful. If we keep any non-decimal encoding, then it should be hex alone.
[...]

PR #91, opened 2023-03-23
[major] Only allow se-ints in hardware literals, change se-ints to radix-encoded ints

seldridge:
Restrict the usage of string-encoded integer literals to only being
allowable used in the construction of hardware integer literals.

Change string-encoded integer literals to radix-specified integer literals.
E.g., make the following change:

"b10101" to 0b01010

Remove string-encoded integer literals.  Continue to disallow the usage of
radix-specified integer literals in any location other than hardware integer
literal construction.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -308,35 +308,38 @@ a trailing `'`{.firrtl}.  The following is an example of a raw string literal:
 'world'
 ```
 
-## String-encoded Integer Literal
+## Radix-specified Integer Literal
 
-A string-encoded integer literal is a special string literal with one of the
+A radix-specified integer literal is a special integer literal with one of the
 following leading characters to indicate the numerical encoding:
 
-- `b`{.firrtl} -- for representing binary numbers
-- `o`{.firrtl} -- for representing octal numbers
-- `h`{.firrtl} -- for representing hexadecimal numbers
+- `0b`{.firrtl} -- for representing binary numbers
+- `0o`{.firrtl} -- for representing octal numbers
+- `0d`{.firrtl} -- for representing decimal numbers
+- `0h`{.firrtl} -- for representing hexadecimal numbers
 
-Signed string-encoded integer literals have their sign following the leading
+Signed radix-specified integer literals have their sign following the leading
 encoding character.
 
 The following string-encoded integer literals all have the value `42`:
 
 ``` firrtl
-"b101010"
-"o52"
-"h2a"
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -347,36 +341,36 @@ a trailing `'`{.firrtl}.  The following is an example of a raw string literal:
...
 
 The following string-encoded integer literals all have the value `42`:
 
 ``` firrtl
-"b101010"
-"o52"
-"h2a"
+b101010
+o52
+h2a
```
mwachs5, line 361: I have a slight preference for the leading-0 notation since it seems like a future proofing against identifiers/parameters? WDYT?

seldridge: I like it. Going with a leading `0` helps future proof things for when parameters inevitably show up.

I won't update this PR just yet as it may be better to have these discussions on https://github.com/chipsalliance/firrtl-spec/issues/89 as opposed to me thrashing this PR back and forth.

E.g., I would like to push either in the direction of no radix-encoded integer literals (decimal everywhere) or only use hex-encoded integer literals for constant integer expressions and decimal everywhere else.

seldridge: Switched to `0b`/`0o`/`0h`.

spec.md
```diff
@@ -314,36 +308,36 @@ a trailing `'`{.firrtl}.  The following is an example of a raw string literal:
...
-A string-encoded integer literal is a special string literal with one of the
+A radix-specified integer literal is a special integer literal with one of the
 following leading characters to indicate the numerical encoding:
 
-- `b`{.firrtl} -- for representing binary numbers
-- `o`{.firrtl} -- for representing octal numbers
-- `h`{.firrtl} -- for representing hexadecimal numbers
+- `0b`{.firrtl} -- for representing binary numbers
+- `0o`{.firrtl} -- for representing octal numbers
+- `0h`{.firrtl} -- for representing hexadecimal numbers
```
darthscsi, line 318: It might seem stupid, and maybe it is, but what about `0d`?

darthscsi (approved):
Less overloading, yay!

spec.md
seldridge: Can't hurt to allow it. Added this.

seldridge:
It would be better if this was added as "alternative" syntax to enable a minor release to go out with it before the hard change-over of FIRRTL 3.0.0. 🤔

PR #92, opened 2023-03-23
[patch] Fixup force/release examples.

dtzSiFive:
* Force expects expression as value, not integer literal.
* two-argument force is 'force_initial', fix in example.
* Fix typo/use of wrong variable in when condition.

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -5,6 +5,7 @@ revisionHistory:
   # additions to the specification should append entries here.
   thisVersion:
   # Information about the old versions.  This should be static.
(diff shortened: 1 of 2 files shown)


jackkoenig (approved):
LGTM, should this be included in changelog? Fixing examples is a bug fix of sorts.

dtzSiFive:
> LGTM, should this be included in changelog? Fixing examples is a bug fix of sorts.

If you had to ask, then yes!

> Since this doesn't change the meaning of anything, I went for 'nfc' but wasn't sure.

I wasn't sure.  I'll add it, thanks! (and promote to patch)

PR #93, opened 2023-03-23
[patch] Tweak read grammar to allow read(probe(x)) as in examples.

dtzSiFive:
This isn't really generally useful but might as well allow it as it's useful for examples (such as those in spec already!).

Note you still can't `read(probe(x).y)` but not worth allowing when you might as well `read(probe(x.y))`.


dtzSiFive:
Probably doesn't need to be draft, seems harmless (and a good improvement re:making spec examples legal)! but not eager to rush :+1: .

seldridge (approved):
LGTM

We should keep the grammar consistent with the examples at minimum.

Thanks!

PR #94, opened 2023-04-10
[minor] add enumeration types and match statements

youngar:
To Do
-----
- [x] needs an entry in the revision-history
- [x] remove vector and bundle expressions
- [x] update the grammar (in the document at the end) once the syntax choice settles
- [x] update the the syntax highlighting to support the new constructs
- [x] tag PR as `[minor]`

### DIFF
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -111,8 +111,8 @@
         <keyword attribute="Conditional" context="#stay" String="conditionals" />
         <Detect2Chars char="(" char1="*" attribute="Info" context="info"/>
         <DetectChar char= ";" context="comment" attribute="Comment"/>
+        <Detect2Chars char="{" char1="|" context="field" attribute="Separator"/>
         <DetectChar char="{" context="field" attribute="Separator"/>
-        <Detect2Chars char="{" char1="|" context="enumField" attribute="Separator"/>
         <Detect2Chars char="&lt;" char1="=" attribute="Operator" context="#stay"/>
         <Detect2Chars char="&lt;" char1="-" attribute="Operator" context="#stay"/>
         <Detect2Chars char="=" char1="&gt;" attribute="Operator" context="#stay"/>
@@ -158,10 +158,10 @@
       <context name="outertype" attribute="Keyword" lineEndContext="#pop">
         <DetectChar char="&lt;" attribute="Operator" context="#stay"/>
         <keyword String="types" attribute="Keyword" context="type"/>
+        <Detect2Chars char="{" char1="|" context="field" attribute="Separator"/>
+        <Detect2Chars char="|" char1="}" attribute="Separator" context="#pop#pop"/>
         <DetectChar char="{" context="field" attribute="Separator"/>
         <DetectChar char="}" attribute="Separator" context="#pop#pop"/>
-        <Detect2Char char="{" char1="|" context="enumField" attribute="Separator"/>
-        <Detect2Char char="}" char1="|" attribute="Separator" context="#pop#pop"/>
         <DetectChar char= ";" context="comment" attribute="Comment"/>
         <AnyChar String="&gt;," attribute="Operator" context="#pop"/>
         <DetectChar char="[" attribute="Operator" context="widthOrDepthOrLit"/>
@@ -170,6 +170,7 @@
         <AnyChar String="&lt;[(" attribute="Operator" context="widthOrDepthOrLit"/>
         <AnyChar String="," attribute="Separator" context="#pop"/>
         <AnyChar String="}" attribute="Separator" context="#pop#pop"/>
+        <Detect2Chars char="|" char1="}" attribute="Separator" context="#pop#pop"/>
         <DetectChar char= ";" context="comment" attribute="Comment"/>
         <AnyChar String="&gt;" attribute="Operator" context="#pop"/>
       </context>
@@ -182,17 +183,10 @@
         <keyword String="types" attribute="Keyword" context="type"/>
         <keyword String="outertype" attribute="Keyword" context="outertype"/>
         <keyword String="typemodifiers" attribute="Keyword" context="#stay"/>
+        <Detect2Chars char="{" char1="|" attribute="Separator" context="field"/>
+        <Detect2Chars char="|" char1="}" attribute="Separator" context="#pop"/>
         <DetectChar char="{" attribute="Separator" context="field"/>
         <DetectChar char="}" attribute="Separator" context="#pop"/>
-        <AnyChar String=":," attribute="Separator" context="#stay"/>
-        <DetectChar char= ";" context="comment" attribute="Comment"/>
-      </context>
-      <context name="enumField" attribute="ID" lineEndContext="#stay">
-        <keyword String="types" attribute="Keyword" context="type"/>
-        <keyword String="outertype" attribute="Keyword" context="outertype"/>
-        <keyword String="typemodifiers" attribute="Keyword" context="#stay"/>
-        <Detect2Char char="{" char1="|" attribute="Separator" context="field"/>
-        <Detect2Char char="}" char1="|" attribute="Separator" context="#pop"/>
         <AnyChar String=":," attribute="Separator" context="#stay"/>
         <DetectChar char= ";" context="comment" attribute="Comment"/>
       </context>
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -44,6 +44,7 @@
     <list name="conditionals">
       <item>when</item>
       <item>else</item>
+      <item>match</item>
     </list>
     <list name="primops">
       <item>attach</item>
@@ -111,6 +112,7 @@
         <Detect2Chars char="(" char1="*" attribute="Info" context="info"/>
         <DetectChar char= ";" context="comment" attribute="Comment"/>
         <DetectChar char="{" context="field" attribute="Separator"/>
+        <Detect2Chars char="{" char1="|" context="enumField" attribute="Separator"/>
         <Detect2Chars char="&lt;" char1="=" attribute="Operator" context="#stay"/>
         <Detect2Chars char="&lt;" char1="-" attribute="Operator" context="#stay"/>
         <Detect2Chars char="=" char1="&gt;" attribute="Operator" context="#stay"/>
@@ -158,6 +160,8 @@
         <keyword String="types" attribute="Keyword" context="type"/>
         <DetectChar char="{" context="field" attribute="Separator"/>
         <DetectChar char="}" attribute="Separator" context="#pop#pop"/>
+        <Detect2Char char="{" char1="|" context="enumField" attribute="Separator"/>
+        <Detect2Char char="}" char1="|" attribute="Separator" context="#pop#pop"/>
         <DetectChar char= ";" context="comment" attribute="Comment"/>
         <AnyChar String="&gt;," attribute="Operator" context="#pop"/>
         <DetectChar char="[" attribute="Operator" context="widthOrDepthOrLit"/>
@@ -180,6 +184,15 @@
         <keyword String="typemodifiers" attribute="Keyword" context="#stay"/>
         <DetectChar char="{" attribute="Separator" context="field"/>
         <DetectChar char="}" attribute="Separator" context="#pop"/>
+        <AnyChar String=":," attribute="Separator" context="#stay"/>
+        <DetectChar char= ";" context="comment" attribute="Comment"/>
+      </context>
+      <context name="enumField" attribute="ID" lineEndContext="#stay">
+        <keyword String="types" attribute="Keyword" context="type"/>
+        <keyword String="outertype" attribute="Keyword" context="outertype"/>
+        <keyword String="typemodifiers" attribute="Keyword" context="#stay"/>
+        <Detect2Char char="{" char1="|" attribute="Separator" context="field"/>
+        <Detect2Char char="}" char1="|" attribute="Separator" context="#pop"/>
         <AnyChar String=":," attribute="Separator" context="#stay"/>
         <DetectChar char= ";" context="comment" attribute="Comment"/>
       </context>
--- a/spec.md
+++ b/spec.md
@@ -2467,26 +2467,6 @@ the data type is `UInt<0>(0)`, where it is implicitly defined to be
 {|some: UInt<8>, None|}(Some, x)
 ```
 
-## Vector Expressions
-
-A vector can be constructed by specifying a list of values and a return type
-specifier.
-
-``` firrtl
-[UInt(2), UInt(3), UInt(4)] : UInt<8>[3]
-[io.a, io.b, io.c] : UInt<32>[3]
-```
-
-## Bundle Expressions
-
-A bundle can be constructed by specifying the field values using a bundle-like
-syntax, and a return type specifier.
-
-``` firrtl
-{a: UInt(0)} : {a : UInt<8>}
-{enable: io.in} : {enable : UInt<1>}
-```
-
 ## References
 
 A reference is simply a name that refers to a previously declared circuit
--- a/spec.md
+++ b/spec.md
@@ -553,13 +553,22 @@ Analog     ; analog type with inferred width
 
 Enumerations are structural disjoint union types.  An enumeration has a number
 of variants, each with a type.  The different variants are specified with tags.
-A variant may optionally omit the data type, in which case it is implicitly
-defined to be `UInt<0>`. The variant types of an enumeration must all be
-passive and cannot contain analog types.
+The variant types of an enumeration must all be passive and cannot contain
+analog or probe types.
+
+In the following example, the first variant has the tag `a` with type `UInt<8>`,
+and the second variant has the tag `b` with type `UInt<16>`.
+
+``` firrtl
+{|a: UInt<8>, b: UInt<16>|}
+```
+
+A variant may optionally omit the type, in which case it is implicitly defined
+to be `UInt<0>`. In the following example, all variants have the type
+`UInt<0>`.
 
 ``` firrtl
 {|a, b, c|}
-{|some: UInt<8>, none|}
 ```
 
 ## Aggregate Types
@@ -1022,8 +1031,9 @@ It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
 The `AsyncReset`{.firrtl} type can be connected to another
 `AsyncReset`{.firrtl} or to a `Reset`{.firrtl}.
 
-Two enumeration types are equivalent if both have the same number of variants,
-and both the enumeration's i'th variants have matching names and types.
+Two enumeration types' are equivalent if both have the same number of variants,
+and both the enumeration's i'th variants have matching names and equivalent
+types.
 
 Two vector types are equivalent if they have the same length, and if their
 element types are equivalent.
@@ -1527,9 +1537,9 @@ when c : a <= b else : e <= f
 
 ### Match Statements
 
-Match statements are used to discriminate the active variant of a enumeration
+Match statements are used to discriminate the active variant of an enumeration
 typed expression.  A match statement must exhaustively test every variant of an
-enumeration. An optional binder may be specified to extract the data of the
+enumeration.  An optional binder may be specified to extract the data of the
 variant.
 
 ``` firrtl
@@ -3548,6 +3558,18 @@ The lowering algorithm for the scalarized convention operates as follows:
    suffix, `_<name>`, to the field called `name`. Fields are scalarized
    recursively, depth-first, and left-to-right.
 
+5. Enumeration-typed ports are scalarized to a tag port and a value port. The
+   tag port is appended with the suffix `_tag` and is a `UInt` type with the
+   minimum number of bits needed to represent the tags of all members of the
+   enumeration type.  The data port is appended with the suffix `_data` and is
+   a `UInt` with the minimum number of bits needed to represent the largest
+   data of all members of the enumeration type. Each member of the enumeration
+   has its data mapped to the lowered data port starting at the most
+   significant byte and taking up as many bits as they are wide.  Aggregate
+   data is mapped recursively, depth-first, and left-to-right to contiguous
+   bits of the data port. Enumeration data maps the tag followed by the value
+   port.
+
 E.g., consider the following port:
 
 ``` firrtl
@@ -3695,7 +3717,7 @@ primop = primop_2expr | primop_1expr | primop_1expr1int | primop_1expr2int ;
 (* Expression definitions *)
 expr =
     ( "UInt" | "SInt" ) , [ width ] , "(" , int_any , ")"
-  | type_enum , "(" , id , [ expr ] , ")"
+  | type_enum , "(" , id , [ "," , expr ] , ")"
   | reference
   | "mux" , "(" , expr , "," , expr , "," , expr , ")"
   | "read" , "(" , ref_expr , ")"
--- a/spec.md
+++ b/spec.md
@@ -1534,9 +1534,9 @@ variant.
 
 ``` firrtl
 match x:
-  #Some(x):
-    a <= x
-  #None:
+  #some(v):
+    a <= v
+  #none:
     e <= f
 ```
 
@@ -2457,8 +2457,10 @@ SInt("h-2a")
 
 ## Enum Expressions
 
-An enumeration can be constructed by specifying the variant tag, an optional
-data value expression, and a return type specifier.
+An enumeration can be constructed by specifying the variant tag, a data value
+expression, and a return type specifier. The data value expression may be
+omitted when the data type is `UInt<0>(0)`, where it is implicitly defined to
+be `UInt<0>(0)`.
 
 ``` firrtl
 #a : [#a, #b, #c]
--- a/spec.md
+++ b/spec.md
@@ -3665,7 +3665,8 @@ type_aggregate = "{" , field , { field } , "}"
                | type , "[" , int_any , "]" ;
 type_ref = ( "Probe" | "RWProbe" ) , "<", type , ">" ;
 field = [ "flip" ] , id , ":" , type ;
-type = ( [ "const" ] , ( type_ground | type_aggregate ) ) | type_ref ;
+type = ( [ "const" ] , ( type_ground | type_enum | type_aggregate ) )
+     | type_ref ;
 
 (* Primitive operations *)
 primop_2expr_keyword =
--- a/spec.md
+++ b/spec.md
@@ -3659,6 +3659,8 @@ info = "@" , "[" , lineinfo, { ",", lineinfo }, "]" ;
 width = "<" , int_any , ">" ;
 type_ground = "Clock" | "Reset" | "AsyncReset"
             | ( "UInt" | "SInt" | "Analog" ) , [ width ] ;
+type_enum = "{|" , { field_enum } , "|}" ;
+field_enum = id, [ ":" , type ] ;
 type_aggregate = "{" , field , { field } , "}"
                | type , "[" , int_any , "]" ;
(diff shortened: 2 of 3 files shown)


youngar:
The bundle and vector expressions don't really belong in the PR, but I wanted to be able to show what they might look like and how they can compose with the enum expression.

spec.md
```diff
@@ -1459,6 +1477,21 @@ The `else`{.firrtl} branch may also be added to the single line:
...
+### Match Statements
+
+Match statements are used to discriminate the active variant of a enumeration
+typed expression.  A match statement must exhaustively test every variant of an
+enumeration. An optional binder may be specified to extract the data of the
+variant.
+
+``` firrtl
+match x:
+  #Some(x):
```
darthscsi, line 1489: What about reusing the existing keyword "with" instead of "#"?

youngar: We don't really need a keyword or anything here at all, e.g. it could just be:
```firrtl
match x:
  Some(x):
```

I added the `#` for mainly for enum create expressions, e.g. `#a : [#a, #b, #c]`, and kept it in the match statements for consistency.

youngar:
I think the most controversial part here is adding new syntax for the enum tags instead of some prim-op, e.g. `#a : [#a, #b, #c]` vs maybe something like `create_enum<a, b, c>("a")`. Happy for people to bikeshed.

spec.md
```diff
@@ -547,6 +547,19 @@ Analog<32> ; 32-bit analog type
...
+Enumerations are structural disjoint union types.  An enumeration has a number
+of variants, each with a type.  The different variants are specified with tags.
+A variant may optionally omit the data type, in which case it is implicitly
+defined to be `UInt<0>`. The variant types of an enumeration must all be
+passive and cannot contain analog types.
+
+``` firrtl
+[#a, #b, #c]
+[#some: UInt<8>, #none]
+```
```
jackkoenig, line 561: It's fine if we don't allow defined enumeration encodings initially, but I think it's going to be a necessary extension so hopefully we have an idea of how that should look in the spec.

spec.md
```diff
@@ -2374,6 +2407,36 @@ SInt("h-2A")
 SInt("h-2a")
 ```
 
+## Enum Expressions
+
+An enumeration can be constructed by specifying the variant tag, an optional
+data value expression, and a return type specifier.
```
youngar, line 2413: Should probably clarify that the data value expression is only optional when the data type is a `UInt<0>`, and is implicitly defined as `UInt<0>(0)`.

dtzSiFive:
Woohoo!  Love how neat this is, and look forward to the aggregate literal expressions :+1: !

Please update the grammar (in the document at the end) once the syntax choice settles, as well as the syntax highlighting to support the new constructs (or file an issue) in the PDF.

I believe the intention is to not support reference-types in these enums, which is already covered by the `passive` restriction but consider making that explicit for clarity.

spec.md
jackkoenig: It can't be `UInt<0>(0)` though right since, 1) if you have multiple alternatives, they should have different values and 2) if undefined, the compiler/synthesis tool should be free to choose any value.

spec.md
mwachs5: sorry are Some and None example enumerations or keywords themselves here?

mwachs5: if they are just examples can we stick with the `some` and `none` capitalization throughout?

youngar: example enumerations

spec.md
youngar: Yeah, I am guessing the two things we will want are to specify the width of the tags and the values of the tags.  I think specifying the values of tags should be easy (e.g. `[#a = 1, #b = 2, #c = 4]`), but I'm not sure how to specify the overall width.  For any more complicated encoding, it would probably be best to write a chisel decode helper function to convert from a UInt to the enum type.

spec.md
mwachs5: i think the assumption is if they dont carry a value, they can never be converted to a hardware value / used in a register...? I'm not sure the use cases given the current set of features how it would be used to have multiple enums with the same 0.U value?

youngar:
Right now there is no plan to let people use "tags as values", e.g.
```
input i : [#a, #b]
when (eq(getTag(i), #a)):
  ...
node n = #b
```

spec.md
mwachs5: is the `match x... some(x)` using the same `x` meaningful? If not, could we use different variables for the two usages?

spec.md
```diff
@@ -2374,6 +2407,36 @@ SInt("h-2A")
...
+
+``` firrtl
+#a : [#a, #b, #c]
```
mwachs5, line 2416: In human language, does this mean "a thing #a is of the anonymous enum type that might be either #a, #b, #c".

It seems that the most common cases for enums isn't covered by this PR, of, (to give the classic example) "A thing Add, that may be of Add (represented as 0) , Sub (represnted as 1), Mul(represented as 2), .... "

Is there a reason we are omitting that 'common' case and then adding Some of a totally unknown value ...?

I guess I'd call this PR "union" instead of "enum" in its current form...?

youngar: Yeah these correspond to discriminated unions.  The name `enum` comes from what it means in [Rust](https://doc.rust-lang.org/reference/items/enumerations.html) and [Swift](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/enumerations/). The idea is that if every variant of an enum has no data (i.e. the data is a `UInt<0>`), then we will lower this to an actual enum.  If there is a data "payload", then it will be lowered to a tagged union, something like
```verilog
struct {
  enum tag {a, b, c}
  union {
    int a_data,
    int b_data,
    int c_data
  } data
}
```

mwachs5: would you be willing to write an "opcode" like example, that has the same aha I see what you are doing here as you are doing with `some` and `none`"? Sorry I'm not getting it :)

mwachs5: i mean, it's not like a_data, b_data, c_data can be any number.

youngar: A simple enum example:
```
module Executor:
  in opcode : [#add, #sub, #mul]
  in arg0: UInt<8>
  in arg1 : UInt<8>
  out result : UInt<8>
  match opcode:
    #add:
      result <= add(arg0, arg1)
    #sub:
      result <= sub(arg0, arg1)
    #mul:
      result <= mul(arg0, arg1)
```

youngar: An enum with a payload:
```
module Foo:
  in signal : { data : [#valid: UInt<8>, #invalid], flip ready: UInt<1>}
  output o : UInt<8>

  signal.ready <= UInt<1>(1)
 
  match signal.data:
    #valid(x):
      o <= x
    #invalid:
      o <= UInt<8>(0)
```

spec.md
jackkoenig: You'll need to be able to specify width because users may want a one-hot encoding (as you showed in your example 🙂).

Per more complicated encodings, yeah I'm not sure if/how we could represent encodings that are more complicated than a 1-1 relationship between the tag and some numerical value or bitvector. It's an interesting question but probably beyond the scope of this current work.

spec.md
jackkoenig: My confusion was the use of `UInt<0>` as defined above as the implicit data type associated with tags that notionally do not have one. It might be worth calling that out even if it is slightly repetitive, eg. something like:
```
An enumeration can be constructed by specifying the variant tag, an optional
data value expression, and a return type specifier.
The data value is required unless the tag's associated data type is `UInt<0>`
which indicates there is no data associated with that tag.
```
Probably something better than that, but that's the idea I'm going for.

I agree with @youngar's proposed clarification.

spec.md
youngar: I'm starting to wonder if specifying the width explicitly is actually helpful, I can't really figure out how it would be observable. Right now the default width is just the minimum required to hold the largest tag value.

We don't have the ability to "unsafe bitcast" a random `UInt` to an enum type, and even if we did you could just extract the right bits for the tag and cast that (and if there is data, concat tag bits with the data bits):
```
input i : UInt<4>
output o : [#a = 0, #b = 1]
o <= asEnum<[#a = 0, #b = 1]>(head(tail(i, 1), 1)
```

FWIW right now you have to manually write some kind of decoder, which I hope should usually be quite cheap in HW.
```
input i : UInt<4>
output o : [#a = 0, #b = 1]
node n = head(tail(i, 1), 1)
when (n == 0):
  o <= #a : [#a = 0, #b = 1]
else when (n == 1)
  o <= #b : [#a = 0, #b = 1]
else:
  o is invalid
```

spec.md
mwachs5: thanks! great and clarifying examples. not sure if it is too many to add to the spec.

spec.md
```diff
@@ -1459,6 +1477,21 @@ The `else`{.firrtl} branch may also be added to the single line:
 when c : a <= b else : e <= f
 ```
 
+### Match Statements
+
+Match statements are used to discriminate the active variant of a enumeration
+typed expression.  A match statement must exhaustively test every variant of an
+enumeration. An optional binder may be specified to extract the data of the
+variant.
```
seldridge, line 1485: This could clarify that the "binder" is `v` below. 

We should also disallow duplicate matches.
```

**Review by seldridge — APPROVED** [review]
The vector/bundle expressions should be split out.

This needs a `[minor]` tag on the PR and an update to the version history. Otherwise, LGTM. I'm very excited to see this happening.

File: spec.md (hunk 5 above)
Comment (darthscsi) [reply spec.md:1485]: I'd suggest making things easy and requiring matches be in tag order.

File: spec.md [hunk 6]
```diff
@@ -2374,6 +2407,38 @@ SInt("h-2A")
...
 
+## Enum Expressions
+
+An enumeration can be constructed by specifying the variant tag, a data value
+expression, and a return type specifier. The data value expression may be
+omitted when the data type is `UInt<0>(0)`, where it is implicitly defined to
+be `UInt<0>(0)`.
+
+``` firrtl
+#a : [#a, #b, #c]
```
darthscsi, line 2418: What about the same syntax as UInt construction?  `[#some : UInt<8>, #none](#some UInt<8>(1))`

spec.md
```diff
@@ -2374,6 +2407,38 @@ SInt("h-2A")
...
+
+A vector can be constructed by specifying a list of values and a return type
+specifier.
+
+``` firrtl
+[UInt(2), UInt(3), UInt(4)] : UInt<8>[3]
+[io.a, io.b, io.c] : UInt<32>[3]
+```
+
+## Bundle Expressions
```
darthscsi, line 2432: Can we leave at least the bundle expression off?  I don't want to create specification for field aliasing between declarations by accident.

spec.md
```diff
@@ -2374,6 +2407,38 @@ SInt("h-2A")
...
+#some(UInt<8>(1)) : [#some: UInt<8>, #none]
+```
+
+## Vector Expressions
+
+A vector can be constructed by specifying a list of values and a return type
+specifier.
+
+``` firrtl
+[UInt(2), UInt(3), UInt(4)] : UInt<8>[3]
```
darthscsi, line 2428: It's slightly cute, but why not just follow UInt/SInt precedent for `type(value)` form?

spec.md
darthscsi: I do think type-unsafe casts are going to be wanted eventually.  You want to be able to specify an enum which matches a standards-defined bit-layout (within reason) and not have to write a reencoder.

jackkoenig:
I don't think it needs to be included in this first PR, but it would be nice to make sure this is easily targetable from ChiselEnum. In particular, we need to ability to cast to and from UInts (yes this is not necessarily safe and [Chisel will warn you if it isn't](https://www.chisel-lang.org/chisel3/docs/explanations/chisel-enum.html#casting)), as well as the ability to compare with `eq` and `neq`.

spec.md
```diff
@@ -547,6 +547,19 @@ Analog<32> ; 32-bit analog type
...
+
+Enumerations are structural disjoint union types.  An enumeration has a number
+of variants, each with a type.  The different variants are specified with tags.
+A variant may optionally omit the data type, in which case it is implicitly
+defined to be `UInt<0>`. The variant types of an enumeration must all be
+passive and cannot contain analog types.
+
+``` firrtl
+[#a, #b, #c]
+[#some: UInt<8>, #none]
```
darthscsi, line 560: Final bikesheding round:
`{{ some : UInt<8>, none}}` or `{| some : UInt<8>, none |}` or `{# some : UInt<8>, none #}`
Why?  These are closest in structure to structs and I don't like this being the one place field identifiers are prefixed.
(hash is sometimes associated with comments, so I'd prefer not that form.  Those examples are I think in order of bikeshed preference)

spec.md
```diff
@@ -1459,6 +1477,21 @@ The `else`{.firrtl} branch may also be added to the single line:
...
+### Match Statements
+
+Match statements are used to discriminate the active variant of a enumeration
+typed expression.  A match statement must exhaustively test every variant of an
+enumeration. An optional binder may be specified to extract the data of the
+variant.
+
+``` firrtl
+match x:
+  #some(v):
```
darthscsi, line 1489: final bikesheding round:
If identifiers aren't prefixed in the type, no reason to prefix them here.

spec.md
darthscsi, line 2418: Final bikesheding round: 
`{| some : UInt<8>, none|}(some, UInt<8>(1))`
Why?  For uninformity with UInt.
Same for vector and bundle, though those really should be a different PR.

youngar:
I think I have addressed all the open items here.  I have gone with the latest suggestion to use a syntax like the following: `{| Some: UInt<8>, None |}`.  I implemented parser support in CIRCT over [here](https://github.com/llvm/circt/pull/5116/files#diff-fd665ef4a1ffa13f71ea4a5e9468d7b49786e2a5fca6b41f78e88447fc58b68aR1146). I didn't really know what I was doing with the KDE grammar, I just copied what the code used for bundle types.

spec.md
```diff
@@ -549,6 +549,19 @@ Analog<32> ; 32-bit analog type
 Analog     ; analog type with inferred width
 ```
 
+### Enumeration Types
+
+Enumerations are structural disjoint union types.  An enumeration has a number
+of variants, each with a type.  The different variants are specified with tags.
+A variant may optionally omit the data type, in which case it is implicitly
+defined to be `UInt<0>`. The variant types of an enumeration must all be
+passive and cannot contain analog types.
```
seldridge, line 558: Nits:

The third sentence uses `data type` when the second sentence uses `type`. Using the latter consistently would be better.

Does passivity exclude probe types or should this be explicitly called out?

spec.md
```diff
@@ -549,6 +549,19 @@ Analog<32> ; 32-bit analog type
...
+
+``` firrtl
+{|a, b, c|}
+{|some: UInt<8>, none|}
+```
```
seldridge, line 563: Some comments or commentary that `a`, `b`, and `c` are "tags" and `UInt<8>` is a "type" would help make it obvious what is going on.

spec.md
```diff
@@ -1009,6 +1022,9 @@ It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
 The `AsyncReset`{.firrtl} type can be connected to another
 `AsyncReset`{.firrtl} or to a `Reset`{.firrtl}.
 
+Two enumeration types are equivalent if both have the same number of variants,
+and both the enumeration's i'th variants have matching names and types.
```
seldridge, line 1026: Mega nit: `enumeration types'`. Also, I'd change "matching names and types" to "matching names and equivalent types" as "matching" isn't defined on types. (I supremely nit that "matching names" should be "equal names" and make it clear that this is string equality, but this is not done for `Bundle` today.)

spec.md
```diff
@@ -1507,6 +1525,21 @@ The `else`{.firrtl} branch may also be added to the single line:
 when c : a <= b else : e <= f
 ```
 
+### Match Statements
+
+Match statements are used to discriminate the active variant of a enumeration
```
seldridge, line 1530: Supreme nit: 
```suggestion
Match statements are used to discriminate the active variant of an enumeration
```

spec.md
```diff
@@ -1507,6 +1525,21 @@ The `else`{.firrtl} branch may also be added to the single line:
...
+typed expression.  A match statement must exhaustively test every variant of an
+enumeration. An optional binder may be specified to extract the data of the
```
seldridge, line 1532: Colossal nit (use the same period spacing convention within a paragraph 😅 ): 
```suggestion
enumeration.  An optional binder may be specified to extract the data of the
```

spec.md
```diff
@@ -3614,11 +3659,14 @@ info = "@" , "[" , lineinfo, { ",", lineinfo }, "]" ;
 width = "<" , int_any , ">" ;
 type_ground = "Clock" | "Reset" | "AsyncReset"
             | ( "UInt" | "SInt" | "Analog" ) , [ width ] ;
+type_enum = "{|" , { field_enum } , "|}" ;
```
seldridge, line 3662: EBNF is really stupid and doesn't have a one-or-more operator, so this is currently saying `{||}` is legal. I think you meant:

```suggestion
type_enum = "{|" , field_enum , { "," , field_enum } , "|}" ;
```

The EBNF hasn't been updated to indicate that commas are whitespace, so this should include the explicit `","`.

include/firrtl.xml
```diff
@@ -183,6 +187,15 @@
...
       </context>
+      <context name="enumField" attribute="ID" lineEndContext="#stay">
+        <keyword String="types" attribute="Keyword" context="type"/>
+        <keyword String="outertype" attribute="Keyword" context="outertype"/>
+        <keyword String="typemodifiers" attribute="Keyword" context="#stay"/>
+        <Detect2Char char="{" char1="|" attribute="Separator" context="field"/>
+        <Detect2Char char="}" char1="|" attribute="Separator" context="#pop"/>
+        <AnyChar String=":," attribute="Separator" context="#stay"/>
+        <DetectChar char= ";" context="comment" attribute="Comment"/>
+      </context>
```
seldridge, line 198: This looks good to me! Thanks for keeping this esoteric highlighter format up-to-date.

spec.md
```diff
@@ -3647,6 +3695,7 @@ primop = primop_2expr | primop_1expr | primop_1expr1int | primop_1expr2int ;
 (* Expression definitions *)
 expr =
     ( "UInt" | "SInt" ) , [ width ] , "(" , int_any , ")"
+  | type_enum , "(" , id , [ expr ] , ")"
```
seldridge, line 3698:
```suggestion
  | type_enum , "(" , id , [ "," , expr ] , ")"
```

spec.md
```diff
@@ -3697,6 +3746,9 @@ statement =
   | "when" , expr , ":" [ info ] , newline ,
     indent , statement, { statement } , dedent ,
     [ "else" , ":" , indent , statement, { statement } , dedent ]
+  | "match" , expr , ":" , [ info ] , newline ,
+    [ indent , { id , [ "(" , id , ")" ] , ":" , newline , 
+    [ indent , { statement } , dedent ] } , dedent ]
```
seldridge, line 3751: Nit: Increasing the indentation for the second indent would help clarify what is going on here. Also, can't this repeat and isn't optional? You would likely need to define a separate `match_statement` and then show that that repeats:

seldridge (commented):
Some nits and some fixes for the EBNF grammar.

This doesn't touch on the lowering, e.g., anything about the resultant sizes of the types. Should this?

spec.md
youngar: Hey @seldridge thanks for taking a close look at this. I did actually intend for the empty enumeration`{||}` to be legal, and this is also why the we don't require at least one "case" in the match statement.  I do see that the grammar is enforcing that there is at least one field in a bundle  (`type_aggregate = "{" , field , { field } , "}"`), when in practice we do allow empty bundles, so I wouldn't be averse to copying that "mistake".

spec.md
youngar: I was trying to make the "case"s entirely optional in the grammar.  It was kind of tricky to do without accidentally allowing each "case" can have a different level of indentation.
e.g. the following is illegal because of the change in indentation, but having empty statement lists is fine:

```
match x:
  a:

   b:
```

There are two `{ }` operators nested, but the outer one was really hard to see the way I had flowed the text.  This might be breaking it up a bit too much but might help read the pattern:
```
  | "match" , expr , ":" , [ info ] , newline ,
    [ indent ,
      { id , [ "(" , id , ")" ] , ":" , newline , 
        [ indent ,
          { statement } ,
        dedent ]
      } ,
    dedent ]
```

spec.md
youngar: Thanks 😅 What is weird here is that bundle types are missing the `","`, but I do see it in most of the rest of the grammar (e.g. primops).

youngar:
> This doesn't touch on the lowering, e.g., anything about the resultant sizes of the types. Should this?

Yeah it should mention something in the scalarized convention at least.  On the implementation side, I haven't thought too much about what happens under the scalarized convention, so here is a quick write up (1st draft) of what we might lower to:
```
5. Enumeration-typed ports are scalarized to a tag port and a value port. The
   tag port is appended with the suffix `_tag` and is a `UInt` type with the
   minimum number of bits needed to represent the tags of all members of the
   enumeration type.  The data port is appended with the suffix `_data` and is
   a `UInt` with the minimum number of bits needed to represent the largest
   data of all members of the enumeration type. Each member of the enumeration
   has its data mapped to the lowered data port starting at the most
   significant byte and taking up as many bits as it is wide.  Aggregate
   data is mapped recursively, depth-first, and left-to-right to contiguous
   bits of the data port.
```

spec.md
youngar: Fixed + `contain analog or probe types.`

spec.md
youngar: Reworded this section, please take a look.

spec.md
youngar: Fixed.  I am a little worried because the types should exactly match with no trunc or ext ("An unsigned integer type is always equivalent to another unsigned integer type regardless of bit width"), similar to array types.  But the wording for arrays uses "equivalent types" as well, so it should be fine.

spec.md
```diff
@@ -3503,6 +3558,18 @@ The lowering algorithm for the scalarized convention operates as follows:
...
+   tag port is appended with the suffix `_tag` and is a `UInt` type with the
+   minimum number of bits needed to represent the tags of all members of the
+   enumeration type.  The data port is appended with the suffix `_data` and is
+   a `UInt` with the minimum number of bits needed to represent the largest
+   data of all members of the enumeration type. Each member of the enumeration
+   has its data mapped to the lowered data port starting at the most
+   significant byte and taking up as many bits as they are wide.  Aggregate
+   data is mapped recursively, depth-first, and left-to-right to contiguous
+   bits of the data port. Enumeration data maps the tag followed by the value
+   port.
```
seldridge, line 3571: Using `type` instead of `member` would be clearer. (Is that right?)

The PR mentions "byte" when I think it means "bit".

Could the PR show an example of the lowering to make it obvious?

seldridge (approved):

seldridge (commented):
To target this from Chisel, we should have the ability for the user to specify the tag value.

You may also want to sanity check that the use cases here will work (potentially with Chisel doing some more complicated code generation of FIRRTL to make it work).

spec.md
```diff
@@ -549,6 +549,28 @@ Analog<32> ; 32-bit analog type
...
 ```
 
+### Enumeration Types
+
+Enumerations are structural disjoint union types.  An enumeration has a number
+of variants, each with a type.  The different variants are specified with tags.
+The variant types of an enumeration must all be passive and cannot contain
+analog or probe types.
+
+In the following example, the first variant has the tag `a` with type `UInt<8>`,
```
dtzSiFive, line 559: It's a pain, but consider adding the `{.firrtl}` to these, so they are highlighted (consistently).

spec.md
```diff
@@ -1009,6 +1031,10 @@ It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
 The `AsyncReset`{.firrtl} type can be connected to another
 `AsyncReset`{.firrtl} or to a `Reset`{.firrtl}.
 
+Two enumeration types' are equivalent if both have the same number of variants,
```
dtzSiFive, line 1034:
```suggestion
Two enumeration types are equivalent if both have the same number of variants,
```

spec.md
```diff
@@ -1009,6 +1031,10 @@ It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
...
+and both the enumeration's i'th variants have matching names and equivalent
```
dtzSiFive, line 1035:
```suggestion
and both the enumerations' i'th variants have matching names and equivalent
```

spec.md
```diff
@@ -1377,7 +1403,9 @@ node mynode = mux(pred, a, b)
 
 ## Conditionals
```
dtzSiFive, line 1404: Nit: Add blurb between a header and its first sub-header.  Here, introducing the two types of conditionals seems like a good idea.

spec.md
```diff
@@ -2422,6 +2465,18 @@ SInt("h-2A")
 SInt("h-2a")
 ```
 
+## Enum Expressions
+
+An enumeration can be constructed by applying an enumeration type to a variant
+tag and a data value expression. The data value expression may be omitted when
+the data type is `UInt<0>(0)`, where it is implicitly defined to be
```
dtzSiFive, line 2472:
```suggestion
the data type is `UInt<0>`{.firrtl}, where it is implicitly defined to be
```

The type is `UInt<0>`, the value is `UInt<0>(0)`.  Added the `{.firrtl}` while visiting.

spec.md
```diff
@@ -3614,11 +3681,14 @@ info = "@" , "[" , lineinfo, { ",", lineinfo }, "]" ;
 width = "<" , int_any , ">" ;
 type_ground = "Clock" | "Reset" | "AsyncReset"
             | ( "UInt" | "SInt" | "Analog" ) , [ width ] ;
+type_enum = "{|" , { field_enum } , "|}" ;
+field_enum = id, [ ":" , type ] ;
```
dtzSiFive, line 3685: 1) Can enum data be const (probably?)?
2) Regardless, consider changing this to not include ref types-- either by splitting the `type` entry (e.g., `type_hw` or something) or just specifying `( type_ground | type_enum | type_aggregate )` here.  This aligns with how it's parsed (FIRRTLBaseType, presently, I believe).  Not critical to reject grammar-level all invalid types, but nice when possible.

dtzSiFive (commented):
Looking good!! Added some comments.

spec.md
darthscsi: 1) I don't see how if the enum itself isn't.  The "const on the bundle" implementation of const talked about elsewhere anticipates this as if the outer enum is constant, the inner data is, but not vice versa.  (e.g. const(non) makes sense, non(const) is really weird).
2) yea, no ref.  I mean, if you had a constant enum, then ref would work out I think, but let's not.

darthscsi:
> Yeah it should mention something in the scalarized convention at least.

I would rather we just didn't scalarize these.  They are, in a sense, already scalar.

spec.md
```diff
@@ -549,6 +549,28 @@ Analog<32> ; 32-bit analog type
 Analog     ; analog type with inferred width
 ```
 
+### Enumeration Types
```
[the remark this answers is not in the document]
darthscsi: Yea, this is a hard one.  they are pretty ground-like in that they don't get decomposed by lower types, but they are more aggregate like.  I've moved them to aggregate, though this may require some other changes.

darthscsi:
Cleaned up and continued here: https://github.com/chipsalliance/firrtl-spec/pull/111

darthscsi:
Merged in #111

PR #100, opened 2023-04-29
[minor] Add "regreset" alternative syntax

seldridge:
Add alternative syntax for specifying a register with a reset.  This is done to simplify parsing, readability, and writability.  This also brings the register more in line with other FIRRTL operations.

The new syntax looks like:

    regreset a: UInt<8>, clk, rst, init

The old syntax that this is an alternative to is:

    reg a: UInt<8>, clk with: (reset => (rst, init))

This syntax is intended to be introduced in the v2.3.0 minor release and made mandatory in the v3.0.0 release.

---

An alternative to this syntax would be to use:

    reg a: UInt<8>, clk, rst, init

I didn't do this as it seems easier to just use a new keyword as opposed to making `reg` variadic.

### DIFF
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -117,6 +117,7 @@
         <AnyChar String=":=.,()" attribute="Separator" context="#stay"/>
         <DetectChar char="[" attribute="Operator" context="widthOrDepthOrLit"/>
         <DetectChar char="&quot;" attribute="String" context="string"/>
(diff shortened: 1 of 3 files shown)


dtzSiFive (approved):
LGTM, this is much nicer! Thanks!

seldridge:
CIRCT parsing support added here: https://github.com/llvm/circt/commit/dae218eb0f91e178504bdb62a64bbdbf0311ae0f

PR #101, opened 2023-05-01
[major] Require Bundle fields are legal IDs

seldridge:
Change the spec to require that Bundle fields are legal identifiers. Currently, the SFC has historically accepted anything, including numbers for fields.  This stems from the MixedVec Chisel utility which emits numeric fields.  Change this to make lexing FIRRTL easier.  Currently, there is extreme ambiguity for something like "0.0".  This could be either a subfield or a floating point number.  By restricting subfields to require legal identifiers, this becomes illegal and the ambiguity is resolved.

Note: the EBNF description is already correct here.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -584,7 +584,8 @@ UInt<16>[10][20]
 ### Bundle Types
 
 A bundle type is used to express a collection of nested and named types.  All
-fields in a bundle type must have a given name, and type.
+fields in a bundle type must have a given name, and type.  All fields must be
+legal identifiers.
 
 The following is an example of a possible type for representing a complex
 number. It has two fields, `real`{.firrtl}, and `imag`{.firrtl}, both 10-bit


mwachs5 (approved):

seldridge:
This support was apparently intentional (see: https://github.com/chipsalliance/chisel/pull/517, https://github.com/chipsalliance/chisel/pull/531, and https://github.com/chipsalliance/firrtl/pull/462) to avoid `__` (double underscore) when expanding programmatically-generated bundles.

It is ideal if we can avoid fields that are integers as it requires context-sensitive lexing.

My read of the original issue is that users want to be able to have a bundle like:

```
output a: { 0 : UInt<1> }
```

And have it lowered to:

```
output a_0: UInt<1>
```

It would be simpler to change the lowering algorithm to say that a leading `_` is added only if a leading underscore is not already present. This _should_ be equivalent to the lowering today except in situations where a leading `_` was already used.

darthscsi:
Would identifier or quoted string be sufficient?

seldridge:
Quoted string is another alternative, e.g.:

```
module Foo:
  input a: { 0: UInt<1> }
  output b: UInt<1>

  b <= a."0"
```

This would also remove the lexing confusion.

This may be better than adding another minor rule to the scalarized module convention.

Three options that I'm aware of:

  1. Revert this and require complicated lexing.
  2. Change the scalarized convention to special case a leading `_`.
  3. Require integers fields to be quoted in subfields.

Of these (1) and (3) should allow for complete equivalence with existing FIRRTL lowerings today. (2) would change some lowerings. (3) probably seems like the best option?

WDYT?

jackkoenig:
A slight variant on using single quotes is to use backticks*, this would allow us to reserve single quotes for another future purpose (raw or multiline strings or something else). We could also then move toward using backticks any time an identifier would be hard to lex (keyword, literal, etc.).

*Shamelessly borrowed from Scala

PR #102, opened 2023-05-01
[major] Add literal identifiers, e.g., `0`

seldridge:
Add literal identifiers which are identifiers that are escaped with
backticks.  This is used to encode an identifier which (extremely
narrowly) is an identifier but starts with a leading number.  This enables
extremely simple lexing and avoids any confusion with other possible
number types, e.g., floats or versions.

Add language that describes how this works for subfields.  E.g., to
subfield into an aggregate "foo" with field "`0`" you would do "foo.`0`".

---

Implements the @darthscsi suggestion from https://github.com/chipsalliance/firrtl-spec/pull/101#issuecomment-1529312052 with suggestions from @jackkoenig to use backticks.

The main use of this is to allow for numeric identifiers to not require context-sensitive lexing. E.g., `foo.0.0.bar` requires context passed from the parser to the lexer so that lexer doesn't see `0.0` and make it a float.

Note: the language in this patch is looser than needed. It allows any quoted string including:

```
circuit Foo:
  module Foo:
    input a: {`🎄`: UInt<1>}
    output b: {`🎄`: UInt<1>}

    connect b.`🎄`, a.`🎄`
```

Further note: the SFC has always accepted any string, e.g., `a: {🎄: UInt<1>}` is fine by the SFC. The MFC has never accepted this.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -3403,13 +3403,16 @@ FIRRTL allows for two types of identifiers:
 Identifiers may only have the following characters: upper and lower case
 letters, digits, and `_`{.firrtl}. Identifiers cannot begin with a digit.
 
-Literal identifiers allow for using a string as an identifier which is not a
-legal identifier.  Such an identifier is encoded using leading and trailing
-backticks, `\``{.firrtl}.  E.g., it is legal to use `\`0\``{.firrtl} as a
-literal identifier in a Bundle field (or anywhere else an identifier may be
-used).  A FIRRTL compiler is allowed to change a literal identifier to an
-identifier of the target language (e.g., Verilog) if the literal identifier is
-not representable in the target language.
+Literal identifiers allow for using an expanded set of characters in an
+identifier.  Such an identifier is encoded using leading and trailing backticks,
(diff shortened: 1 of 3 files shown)


spec.md
```diff
@@ -2490,6 +2490,15 @@ module MyModule :
   out.a <= in ; out.a is of type const UInt
 ```
 
+A sub-field referring to a field whose name is a literal identifier is shown
+below:
+
+``` firrtl
```
jackkoenig, line 2496:
```suggestion
```firrtl
```

jackkoenig (approved):

spec.md
seldridge: Either one of these will work properly and it's used somewhat inconsistently within the document with most of them having a space. The ones just before this use the style with a space, so I'll keep this. 

The reason for this is because this happens to be the setting my editor has when I do a triple-backtick. It prompts me for the format and then adds a space. 🤷‍♀️

seldridge:
Literal identifier parsing support added to CIRCT in: https://github.com/llvm/circt/commit/c5ce48d54d4b566885751420c6dd19d59c0ee6b3

PR #103, opened 2023-05-03
[Minor?] Start documenting and defining the ABI for FIRRTL

darthscsi:
Start documenting the ABI for firrtl.  This is put in a new file as the semantics of FIRRTL are specified in the spec, whereas the mapping to verilog makes additional requirements which are independent of what is needed to ensure semantic preserving translation.

### DIFF
--- a/abi.md
+++ b/abi.md
@@ -1,0 +1,177 @@
+---
+author:
+- The FIRRTL Specification Contributors
+title: Specification for the FIRRTL Language
+date: \today
+# Options passed to the document class
+classoption:
+- 12pt
+# Link options
+colorlinks: true
+linkcolor: blue
+filecolor: magenta
+urlcolor: cyan
+toccolor: blue
+# General pandoc configuration
+toc: true
+numbersections: true
+# Header Setup
+pagestyle:
+  fancy: true
+# Margins
+geometry: margin=1in
+# pandoc-crossref
+autoSectionLabels: true
+figPrefix:
+  - Figure
+  - Figures
+eqnPrefix:
+  - Equation
+  - Equations
+tblPrefix:
+  - Table
+  - Tables
+lstPrefix:
+  - Listing
+  - Listings
+secPrefix:
+  - Section
+  - Sections
+# This 'lastDelim' option does not work...
+lastDelim: ", and"
+---
+
+# Introduction
+
+FIRRTL defines a language/IR for describing synchronous hardware circuits.  This
+document specifies the mapping of FIRRTL constructs to verilog in a manner 
+similar to an application binary interface (ABI) which enables predictability of
+the output of key constructs necessary for the interoperability between circuits 
+described in FIRRTL and between other languages and FIRRTL output.
+
+This document describes multiple versions of the ABI, specifically calling 
+specific changes in later versions.  It is expected that a conforming FIRRTL
+compile can lower to all specified ABIs.  This mechanism exist to allow 
+improved representations when using tools which have better verilog support and
+allow migration of existing development flows to significant representational 
+changes.
+
+# FIRRTL System Verilog Interface
+
+To use a circuit describe in FIRRTL in a predictable way, the mapping of certain 
+behaviors and boundary constructs in FIRRTL to System Verilog must be defined.  
+Where possible this ABI does not impose constraints on implementation, 
+concerning itself primarily with the boundaries of the circuit.
+
+## On Modules
+
+### The Circuit and Top Level
+
+The top level module, as specified by the circuit name, shall be present as a
+System Verilog module of the same name.  This firrtl module is considered a 
+"public" module and subject to the lowering constraints for public modules.
+
+### External Modules
+
+An external module may be presumed to exist following the lowering constraints
+for public modules.
+
+###  Public Modules
+
+Any module considered a "public" module shall be implemented in verilog in a
+consistent way.  Any public module shall exist as a verilog module of the same 
+name.
+
+Each public modules with definitions (e.g. not external modules) shall be placed 
+in a file with the same name.  No assumption is made of the filename of an
+implementation of an external module, it is the user's job to include files in
+such a way in their tools to resolve the name.
+
+### Port Lowering ABIv1
+
+Ports are generally lowered to netlist types, except where Verilog's type system
+prevents it.
+
+Ports of integer types shall be lowered to netlist ports (`wire`) as a packed 
+vector of equivalent size.
+
+```
+// FIRRTL:
+circuit Top :
+  module Top :
+    output out: UInt<16>
+    input b: UInt<32>
+
+// SV:
+module Top(
+    output wire [15:0] out,
+    input wire [31:0] in
+);
+```
+
+Ports of aggregate type shall be lowered according to the "Aggregate Type 
+Lowering" description in the firrtl spec.
+
+Ports of enum type shall be lowered to ports of a type consistent with the rules
+for lowering firrtl enum types to verilog.
+
+Ports of ref type shall be lowered to a verilog macro of the form 
+"`define ref_<circuit name>_<module name>_<portname> <internal path from module>"
+in a file with name "ref_<circuit name>_<module name>.sv".
+
+### Port Lowering ABIv2
+
+Ports are lowered per the v1 ABI above, except for aggregate types.
+
+Vectors shall be lowered to verilog packed vectors.
+
+Bundles shall be recursively split as per "Aggregate Type Lowering", except instead of recursively converting bundles to ground types, the recursion stops at passive types.
+
+Passive bundles shall be lowered to verilog packed structs.
+
+Reference types in ports shall be logically split out from aggregates and named
+as though "Aggregate Type Lowering" was used.
+
+## On Types
+
+Types are only guaranteed to follow this lowering when the verilog type is 
+visible for some reason related to the visibility of the element.  These include
+use in ports and elements exported by reference through a public module.
+
+Ground types are lowered to the `logic` data type (SV.6.3.1), which is the 
+4-valued type.  It is important to distinguish the `logic` data type from using
+`logic` as a keyword to declare a variable instead of net-list object.
+
+Unsigned integers produce an unsigned packed bit vector.  Whether variable or 
+netlist depends on the construct being used.
+
+Signed integers produce an unsigned packed bit vector.  Whether variable or 
+netlist depends on the construct being used.
+
+Passive bundles, when lowered, are lowered to packed structs with their fields
+recursively following these lowering rules.
+
+Vectors, when lowered, are lowered to packed vectors with their element type 
+recursively following these rules.
+
+
+# Versioning Scheme of this Document
+
+This is the versioning scheme that applies to version 1.0.0 and later.
+
+The versioning scheme complies with
+[Semantic Versioning 2.0.0](https://semver.org/#semantic-versioning-200).
+
+Specifically,
+
+The PATCH digit is bumped upon release which only includes non-functional changes,
+such as grammar edits, further examples, and clarifications.
+
+The MINOR digit is bumped for feature additions to the spec.
+
+The MAJOR digit is bumped for backwards-incompatible changes such as features
+being removed from the spec, changing their interpretation, or new required
+features being added to the specification.
+
+In other words, any `.fir` file that was compliant with `x.y.z` will be compliant
+with `x.Y.Z`, where `Y >= y`, `z` and `Z` can be any number.
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -8,6 +8,7 @@ revisionHistory:
     - Delineate string and single-quoted/double-quoted string in grammar.
     - Deprecate reference-first statements.
     - Tweak grammar of 'read' to support 'read(probe(x))' as in examples.
+    - Initial ABI description.
 
   # Information about the old versions.  This should be static.
   oldVersions:
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -6,6 +6,8 @@ revisionHistory:
   thisVersion:
     - Fix typos in force/release examples, force takes expr not int literal.
     - Delineate string and single-quoted/double-quoted string in grammar.
+    - Deprecate reference-first statements.
+  
   # Information about the old versions.  This should be static.
   oldVersions:
     - version: 2.0.1
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ IMG_EPSS=$(IMG_SRCS:include/img_src/%.dot=build/img/%.eps)
 .PHONY: all clean images
 .PRECIOUS: build/ build/img/
 
-all: build/spec.pdf
+all: build/spec.pdf build/abi.pdf
 
 clean:
 	rm -rf build
@@ -23,6 +23,9 @@ PANDOC_FLAGS=\
 	--metadata version:$(VERSION)
 
 build/spec.pdf: spec.md revision-history.yaml include/spec-template.tex include/firrtl.xml include/ebnf.xml $(IMG_EPSS) | build/
+	pandoc $< --metadata-file=revision-history.yaml $(PANDOC_FLAGS) -o $@
+
+build/abi.pdf: abi.md revision-history.yaml include/spec-template.tex include/firrtl.xml $(IMG_EPSS) | build/
 	pandoc $< --metadata-file=revision-history.yaml $(PANDOC_FLAGS) -o $@
 
 build/img/%.eps: include/img_src/%.dot | build/img/
--- a/spec.md
+++ b/spec.md
@@ -1070,6 +1070,18 @@ connected to the left-hand side field.  Conversely, if the i'th field is
 flipped, then the left-hand side field is connected to the right-hand side
 field.
 
+### Alternate Syntax
+
+Connects may also be specified by keyword.  This form is identical to the `<=` 
+form in operand order
+
+``` firrtl
+module MyModule :
+  input myinput: UInt
+  output myoutput: UInt
+  connect myoutput, myinput
+```
+
 ## Statement Groups
 
 An ordered sequence of one or more statements can be grouped into a single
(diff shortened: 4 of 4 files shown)


abi.md
```diff
@@ -0,0 +1,177 @@
...
+
+FIRRTL defines a language/IR for describing synchronous hardware circuits.  This
+document specifies the mapping of FIRRTL constructs to verilog in a manner 
+similar to an application binary interface (ABI) which enables predictability of
+the output of key constructs necessary for the interoperability between circuits 
+described in FIRRTL and between other languages and FIRRTL output.
+
+This document describes multiple versions of the ABI, specifically calling 
+specific changes in later versions.  It is expected that a conforming FIRRTL
+compile can lower to all specified ABIs.  This mechanism exist to allow
```
mwachs5, line 54: Compiler, exists

abi.md
```diff
@@ -0,0 +1,177 @@
...
+An external module may be presumed to exist following the lowering constraints
+for public modules.
+
+###  Public Modules
+
+Any module considered a "public" module shall be implemented in verilog in a
+consistent way.  Any public module shall exist as a verilog module of the same 
+name.
+
+Each public modules with definitions (e.g. not external modules) shall be placed
```
mwachs5, line 85: module

revision-history.yaml
```diff
@@ -8,6 +8,7 @@ revisionHistory:
     - Add connect, invalidate to syntax highlighting
     - Add alternative `regreset` syntax
     - Add literal identifiers to allow for legal numeric fields
+    - Initial ABI description.
```
mwachs5, line 11: Shouldn't this go into a different revision history yml file...?

mwachs5 (commented):
Woohoo! Thanks for writing this up. Some minor spelling issues

revision-history.yaml
darthscsi: Is this a change log for the repository or just for the spec?  Are they versioned together?  The ABI has more than passive ties to the constructs described in the spec.

Adding enums, say, will require some ABI specification since they can appear on ports.

I was taking the strategy of co-versioning the documents, but having the ABI versions be orthogonal to the spec version.  That way when enums are added, they can be specified for both the type-lowering abi (v1) and the aggregate preservation abi (v2).

mwachs5: I don't know the answer yet to whether the spec version and the repo version should match, today they are identical but we could have branches for the api docs too but that starts getting crazy...? 

The way it's set up now works pretty well to make point releases of the spec on branches of the repo. Not sure how branchy/different the abi doc would be from the spec. We could always make branches specific for the abi doc point versions if we find ourselves needing it

azidar:
LOVE IT!! Thank you for doing this!

abi.md
```diff
@@ -0,0 +1,177 @@
...
+  - Section
+  - Sections
+# This 'lastDelim' option does not work...
+lastDelim: ", and"
+---
+
+# Introduction
+
+FIRRTL defines a language/IR for describing synchronous hardware circuits.  This
+document specifies the mapping of FIRRTL constructs to verilog in a manner
```
mruff-sifive, line 47: I believe Verilog is a proper name so s/verilog/Verilog/

abi.md
```diff
@@ -0,0 +1,177 @@
...
+
+### The Circuit and Top Level
+
+The top level module, as specified by the circuit name, shall be present as a
+System Verilog module of the same name.  This firrtl module is considered a 
+"public" module and subject to the lowering constraints for public modules.
+
+### External Modules
+
+An external module may be presumed to exist following the lowering constraints
```
trilorez, line 76: Is it worth mentioning that external modules have the optional `defname` override for the Verilog name?

revision-history.yaml
seldridge: The contents of this are used to populate the Version History section of the spec. If the spec and the ABI are co-versioned, then there's not much issue other than we should add two sub-sections here and have the version history pulled out differently, e.g.,:

```yaml
thisVersion:
  spec:
    - change legal identifiers to be emoji only
  abi:
    - redstone ABI defined
```

mwachs5: @seldridge how would that work if we go to an actual version? Like let's say we release 1.1.0 of ABI doc and version 3.1.0 of spec doc, then start adding more to `thisVersion` for both... what does that look like?

To me if we want different docs we just have different versions of this file with their own version history parts below `thisVersion`...?

Then i think the release of this repo should just follow the spec version or if we really care we could stick spec and ABI version into the release tag

abi.md
darthscsi: yes, quite.

revision-history.yaml
seldridge: The question is if the spec and ABI documents are under the same version. If they are, then this will work. If they are separate, then a separate file would be better. Lenharth is arguing that they need the same version because the spec is going to add stuff which the ABI needs to talk about. We are expecting to have multiple ABIs which is why there are v1 and v2 of the ABI internal to the document.

Mechanically, every version will have a `spec` and `abi` keys whose values are a list of changes.

seldridge:
I landed all the build flow enhancements for this to work on `main` and forcibly rebased this.

revision-history.yaml
mwachs5: ahhh ok, i misread this as having different versions of the ABI document. I agree if they are all versioned the same as a document with different versions of the ABI in each document, that is much simpler

abi.md
```diff
@@ -0,0 +1,146 @@
+---
+author:
+- The FIRRTL Specification Contributors
+title: Specification for the FIRRTL ABI
+revisionHistoryAbi: true
+---
+
```
mwachs5, line 7: @seldridge does something need to be added here to make the revision history show up? Or does that happen automatically due to what you did in the previous PR?

seldridge: Nothing else needs to be added. I landed the support earlier. This specific config option was added in https://github.com/chipsalliance/firrtl-spec/commit/9bb1272145484afcd698015e2b6de648123629ee.

This is mega-jank (due to limitations of using a single LaTeX template, what pandoc lets you do, and my own skill), but there are two configuration options you can define up front:

```yaml
revisionHistoryAbi:
revisionHistorySpec:
```

Each of these, if defined, controls, respectively, whether the `abi` or `spec` parts of the `revision-history.yaml` are included in that Revision History section of a document. This clearly doesn't scale well to multiple documents and it would be nice to customize the LaTeX template based on the value associated with a key as opposed to the existence of a key (but pandoc doesn't let you do this as far as I know). It's good enough for today, though.

seldridge: Here is what this looks like in the built `.pdf`s:

Screenshot 2023-05-03 at 15 24 43
Screenshot 2023-05-03 at 15 25 21

mwachs5: Looks great!

abi.md
```diff
@@ -0,0 +1,147 @@
...
+This document describes multiple versions of the ABI, specifically calling
+specific changes in later versions.  It is expected that a conforming FIRRTL
+compiler can lower to all specified ABIs.  This mechanism exists to allow
+improved representations when using tools which have better Verilog support and
+allow migration of existing development flows to significant representational
+changes.
+
+# FIRRTL System Verilog Interface
+
+To use a circuit describe in FIRRTL in a predictable way, the mapping of certain
```
dtzSiFive, line 25:
```suggestion
To use a circuit described in FIRRTL in a predictable way, the mapping of certain
```

abi.md
```diff
@@ -0,0 +1,147 @@
...
+document specifies the mapping of FIRRTL constructs to Verilog in a manner
+similar to an application binary interface (ABI) which enables predictability of
+the output of key constructs necessary for the interoperability between circuits
+described in FIRRTL and between other languages and FIRRTL output.
+
+This document describes multiple versions of the ABI, specifically calling
+specific changes in later versions.  It is expected that a conforming FIRRTL
+compiler can lower to all specified ABIs.  This mechanism exists to allow
+improved representations when using tools which have better Verilog support and
+allow migration of existing development flows to significant representational
```
dtzSiFive, line 20: "allow migration.. to significant representational changes"? What does this mean?

abi.md
```diff
@@ -0,0 +1,147 @@
...
+
+```verilog
+module Top(
+    output wire [15:0] out,
+    input wire [31:0] in
+);
+```
+
+Ports of aggregate type shall be lowered according to the "Aggregate Type
+Lowering" description in the firrtl spec.
```
dtzSiFive, line 80: `s/firrtl/FIRRTL/g` (in the text anyway).

dtzSiFive (commented):
Awesome, thanks!!! Added some comments/feedback.

abi.md
```diff
@@ -0,0 +1,147 @@
...
+
+Ports of enum type shall be lowered to ports of a type consistent with the rules
+for lowering firrtl enum types to Verilog.
+
+Ports of ref type shall be lowered to a Verilog macro of the form `` `define
+ref_<circuit name>_<module name>_<portname> <internal path from module>`` in a
```
dtzSiFive, line 86: As we add other non-hw "things" to FIRRTL (?), it may be useful to have them all in a single "link-interface" sort of file, vs naming a single file for each "kind" of information per module?

Probably premature to try to accommodate those needs, especially as the nature of what information is needed is not yet known, but want to get that in front of us so we don't end up with too many side files describing higher-level details of the emitted Verilog code.

abi.md
```diff
@@ -0,0 +1,147 @@
...
+Where possible this ABI does not impose constraints on implementation,
+concerning itself primarily with the boundaries of the circuit.
+
+## On Modules
+
+### The Circuit and Top Level
+
+The top level module, as specified by the circuit name, shall be present as a
+System Verilog module of the same name.  This firrtl module is considered a
+"public" module and subject to the lowering constraints for public modules.
```
mmaloney-sf, line 36: "public" under who's notion of publicness? Is there a public in SystemVerilog?

mmaloney-sf: Ah. I see it's defined below.

abi.md
```diff
@@ -0,0 +1,147 @@
...
+implementation of an external module, it is the user's job to include files in
+such a way in their tools to resolve the name.
+
+### Port Lowering ABIv1
+
+Ports are generally lowered to netlist types, except where Verilog's type system
+prevents it.
+
+Ports of integer types shall be lowered to netlist ports (`wire`{.verilog}) as a
+packed vector of equivalent size.  For example, consider the following FIRRTL:
```
mmaloney-sf, line 61: Should it be mandated as well that the numbering be done with the schema `[n-1:0]` for an integer type of n-bits?

abi.md
```diff
@@ -0,0 +1,147 @@
...
+    output wire [15:0] out,
+    input wire [31:0] in
+);
+```
+
+Ports of aggregate type shall be lowered according to the "Aggregate Type
+Lowering" description in the firrtl spec.
+
+Ports of enum type shall be lowered to ports of a type consistent with the rules
+for lowering firrtl enum types to Verilog.
```
mmaloney-sf, line 83: This seems like a punt. Could we either be explicit about what we mean by "rules for lowering firrtl enum types" or else make a note admitting that we aren't ready to commit to a single rule at this point in time?

abi.md
```diff
@@ -0,0 +1,147 @@
...
+
+Ports of ref type shall be lowered to a Verilog macro of the form `` `define
+ref_<circuit name>_<module name>_<portname> <internal path from module>`` in a
+file with name `ref_<circuit name>_<module name>.sv`.
+
+### Port Lowering ABIv2
```
mmaloney-sf, line 89: It looks like we're defining multiple ABIs in this document. If that is the case, we should make sure to motivate why we need more than one. (Especially for the initial version of the doc).

abi.md
```diff
@@ -0,0 +1,147 @@
...
+Passive bundles shall be lowered to Verilog packed structs.
+
+Reference types in ports shall be logically split out from aggregates and named
+as though "Aggregate Type Lowering" was used.
+
+## On Types
+
+Types are only guaranteed to follow this lowering when the Verilog type is
+visible for some reason related to the visibility of the element.  These include
+use in ports and elements exported by reference through a public module.
```
mmaloney-sf, line 108: This might be made clearer by defining a notion of a "public" type analogous to a public module. It isn't unclear as it stands, but it might be good to have that definition in our back pocket for later.

abi.md
```diff
@@ -0,0 +1,147 @@
...
+);
+```
+
+Ports of aggregate type shall be lowered according to the "Aggregate Type
+Lowering" description in the firrtl spec.
+
+Ports of enum type shall be lowered to ports of a type consistent with the rules
+for lowering firrtl enum types to Verilog.
+
+Ports of ref type shall be lowered to a Verilog macro of the form `` `define
```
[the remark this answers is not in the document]
mwachs5: > Is the build system responsible for picking up these ABI side-files (and bringing them to where they're needed for separate compilation purposes)? Glad the "what's needed to link against" information is split out.

I think the answer written is "yes" -- we could say that the compiler has to add `include anywhere the port is used but that is not necessary for now

abi.md
mwachs5: there are different versions used even in the same spec, for public vs internal modules

abi.md
```diff
@@ -0,0 +1,147 @@
...
+The PATCH digit is bumped upon release which only includes non-functional changes,
+such as grammar edits, further examples, and clarifications.
+
+The MINOR digit is bumped for feature additions to the spec.
+
+The MAJOR digit is bumped for backwards-incompatible changes such as features
+being removed from the spec, changing their interpretation, or new required
+features being added to the specification.
+
+In other words, any `.fir` file that was compliant with `x.y.z` will be compliant
```
mwachs5, line 146:
```suggestion
In other words, any `.sv` file that was emitted in a fashion compliant with `x.y.z` will be compliant
```

abi.md
```diff
@@ -0,0 +1,147 @@
...
```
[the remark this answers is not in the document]
mwachs5: made a suggestion below

abi.md
```diff
@@ -0,0 +1,147 @@
...
+visible for some reason related to the visibility of the element.  These include
+use in ports and elements exported by reference through a public module.
+
+Ground types are lowered to the `logic`{.verilog} data type (SV.6.3.1), which is
+the 4-valued type.  It is important to distinguish the `logic`{.verilog} data
+type from using `logic`{.verilog} as a keyword to declare a variable instead of
+net-list object.
+
+Unsigned integers produce an unsigned packed bit vector.  Whether variable or
+netlist depends on the construct being used.
```
mmaloney-sf, line 116: This isn't to do with this document, but what is the distinction between a "variable" and a "netlist" here? Is this `reg` vs `wire` in Verilog? Or am I completely off-base?

abi.md
```diff
@@ -0,0 +1,147 @@
...
+
+Signed integers produce an unsigned packed bit vector.  Whether variable or
+netlist depends on the construct being used.
+
+Passive bundles, when lowered, are lowered to packed structs with their fields
+recursively following these lowering rules.
+
+Vectors, when lowered, are lowered to packed vectors with their element type
```
mmaloney-sf, line 124: It might be good hygiene to give explicit examples, in this section and earlier, of the FIRRTL types and the lowered SystemVerilog types that come from its lowering.

abi.md
darthscsi: internal modules have no ABI constraints.
v1 is essentially documenting what the contract is now.
v2 is aggregate-preserving.
They are not remotely compatible.  These things can be stated explicitly.

abi.md
darthscsi: reg/logic v.s. wire in verilog.

abi.md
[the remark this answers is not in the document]
darthscsi: I think a workable solution for type lowered abi is in now.

abi.md
darthscsi: When we know, we'll do it.

abi.md
darthscsi: Good catch.  Done.

abi.md
darthscsi: put the forward reference in.  Will add the lowering.

abi.md
darthscsi: Since types in firrtl are structural, I don't want to claim a particular lowering for a type, but rather for circuit elements which are visible and have that type.

abi.md
darthscsi: Yes, but I think that can be a follow on PR.

seldridge:
@darthscsi wrote:
> I think any public module needs to have it's output references defined. I thought the language in the spec already forbid input references on public modules.

It's currently restricted to "using a probe reference", i.e., a `read`. This would expand that definition of "use" to include "public modules".

abi.md
```diff
@@ -0,0 +1,164 @@
...
+such as grammar edits, further examples, and clarifications.
+
+The MINOR digit is bumped for feature additions to the spec.
+
+The MAJOR digit is bumped for backwards-incompatible changes such as features
+being removed from the spec, changing their interpretation, or new required
+features being added to the specification.
+
+In other words, any verilog generated that was emitted with `x.y.z` will be 
+compliant with `x.Y.Z`, where `Y >= y`, `z` and `Z` can be any number.
```
seldridge, line 164: This can just say that this document is co-versioned with the FIRRTL specification.

seldridge (approved):
LGTM

abi.md
```diff
@@ -0,0 +1,164 @@
...
+###  Public Modules
+
+Any module considered a "public" module shall be implemented in Verilog in a
+consistent way.  Any public module shall exist as a Verilog module of the same
+name.
+
+Each public module with definitions (e.g. not external modules) shall be placed
+in a file with the same name.  No assumption is made of the filename of an
+implementation of an external module, it is the user's job to include files in
+such a way in their tools to resolve the name.
```
seldridge, line 60: This requirement of public modules being in separate files is reasonable. However, this has some complications:

1. Where is the sub-hierarchy of a public module emitted? What happens if two public modules instantiate the same private sub-module?
2. This disallows any single-file emission mode for any circuit which has a public module other than the just the main module. That's possibly fine, but may not be desired.

darthscsi: public non-top modules is a new thing, so I don't think we are too worried about it.  Single-public top still supports single file emission.

as for (1), the more general question is "what are all the things I need to include and in what order?" and I've side stepped that a few places for now.

PR #106, opened 2023-05-08
[Minor] Add type alias

uenoku:
This PR introduces type alias which provides a mechanism to assign names to FIRRTL types.


spec.md
```diff
@@ -703,6 +704,36 @@ Probe types may target `const`{.firrtl} signals, but cannot use
 `RWProbe<const T>`{.firrtl}, as constant values should never be mutated at
 runtime.
 
+## Type Alias
+
+A type alias provides a way to assign names to existing FIRRTL types, enabling
+their reuse across multiple declarations and potentially serving as hints in
+verilog code generation.
```
seldridge, line 711: Nits: use stronger, direct language here. "A type alias is", "Type aliases enable", etc.

The "potentially serving as hints may be confusing. (Is this mandatory, is this optional, etc.?) You may want to leave this out and adds this to the ABI doc when that lands.

spec.md
```diff
@@ -703,6 +704,36 @@ Probe types may target `const`{.firrtl} signals, but cannot use
...
+  input in: Data
+  output out: Data
+  wire w: AnotherWordType
+  w <= in.w
+  ...
+```
+
+The `type` declaration is globally defined and all named types exist in the same
+namespace and thus must all have a unique name. Type alias doesn't share the same
+type scope as modules; hence it is allowed for type aliaes to conflict with module
```
seldridge, line 729:
```suggestion
namespace and thus must all have a unique name. Type alias do not share the same
type scope as modules; hence it is allowed for type aliaes to conflict with module
```

spec.md
```diff
@@ -703,6 +704,36 @@ Probe types may target `const`{.firrtl} signals, but cannot use
...
+names. In the output verilog the given names may not be preserved by the FIRRTL
+compiler, as names must be unique in the output relative to target-dependent scoping
+rules which may require changing names or dropping. Note that type aliases do not
```
seldridge, line 732: Nit:
```suggestion
rules which may require changing or dropping names. Note that type aliases do not
```

spec.md
```diff
@@ -703,6 +704,36 @@ Probe types may target `const`{.firrtl} signals, but cannot use
...
+introduce nominal types unlike bundle field names; the type equivalence is determined
+solely by their structure. For instance types of `w`{.firrtl} and `in.w`{.wirrtl} are
+equivalent even though they are different type alias.
```
seldridge, line 735: The PR could avoid talking about "nominal types" and just say that the types are compared structurally.  This can then refer to the type equivalence section for more information.

spec.md
```diff
@@ -3610,6 +3641,9 @@ type_ref = ( "Probe" | "RWProbe" ) , "<", type , ">" ;
 field = [ "flip" ] , id , ":" , type ;
 type = ( [ "const" ] , ( type_ground | type_aggregate ) ) | type_ref ;
 
+(* Type alias declaration *)
+type_alias_decl = "type", id, "=", type ;
```
seldridge, line 3645: This should probably be included int he EBNF for `type` as well. This forces questions like can you have a `const` in a typedef, can you apply a `const` onto a named type? Etc.

spec.md
mwachs5, line 729:
```suggestion
namespace as modules; hence it is allowed for type aliases to conflict with module
```

spec.md
mwachs5, line 732: can we move the information about how this is lowered to the ABI spec...? 

I think it's pretty important to know whether the types are lowered to verilog or not, because if they aren't lowered I can call them "foobizzle_1" and "TypeAlias_123". Also stability is concerning.

uenoku: Removed information about output verilog. I'll mention in ABI spec once ABI spec was merged.

spec.md
uenoku: Yeah, that's good point.  I updated EBNF to allow typealias to be quantified by const.

spec.md
```diff
@@ -703,6 +704,32 @@ Probe types may target `const`{.firrtl} signals, but cannot use
 `RWProbe<const T>`{.firrtl}, as constant values should never be mutated at
 runtime.
 
+## Type Alias
+
+A type alias is a mechanism to assign names to existing FIRRTL types. Type aliases
+enables their reuse across multiple declarations.
+
```
darthscsi, line 711: stray '

spec.md
```diff
@@ -703,6 +704,32 @@ Probe types may target `const`{.firrtl} signals, but cannot use
...
+  w <= in.w
+  ...
+```
+
+The `type` declaration is globally defined and all named types exist in the same
+namespace and thus must all have a unique name. Type aliases do not share the same
+namespace as modules; hence it is allowed for type aliases to conflict with module
+names. Note that when we compare two types, the equivalence is determined solely by
+their structures. For instance types of `w`{.firrtl} and `in.w`{.wirrtl} are
+equivalent even though they are different type alias.
```
darthscsi, line 731:
```suggestion
equivalent in the example above even though they are different type alias.
```

spec.md
```diff
@@ -3607,8 +3634,13 @@ type_ground = "Clock" | "Reset" | "AsyncReset"
 type_aggregate = "{" , field , { field } , "}"
                | type , "[" , int_any , "]" ;
 type_ref = ( "Probe" | "RWProbe" ) , "<", type , ">" ;
+type_alias = id ;
 field = [ "flip" ] , id , ":" , type ;
-type = ( [ "const" ] , ( type_ground | type_aggregate ) ) | type_ref ;
+type = ( [ "const" ] , ( type_ground | type_aggregate | type_alias ) )
+     | type_ref | type_alias ;
```
darthscsi, line 3640: This second type alias seems wrong.

spec.md
```diff
@@ -3607,8 +3634,13 @@ type_ground = "Clock" | "Reset" | "AsyncReset"
...
+
+(* Type alias declaration *)
+type_alias_decl = "type", id, "=", type ;
```
darthscsi, line 3643: Why use id here but not "type_alias" which is just id above (or alternately why have type_alias at all)?

darthscsi, line 3643: Is this legal?
```
type a = UInt<3>
type b = a
```

uenoku: Yes, this is legal.

uenoku:
I don't have a write permission so it would be great someone could merge it.

PR #107, opened 2023-05-12
[minor] Add remote instantiation

seldridge:
Add remote instantiation with the "remote_inst foo of Foo" syntax.  This is a FIRRTL-native representation of an instantiation that should be lowered to a SystemVerilog bind.

### DIFF
--- a/abi.md
+++ b/abi.md
@@ -110,6 +110,51 @@ Passive bundles shall be lowered to Verilog packed structs.
 Reference types in ports shall be logically split out from aggregates and named
 as though "Aggregate Type Lowering" was used.
 
+### Remotely Instantiated Modules
+
+Remotely instantiated modules are lowered using the SystemVerilog
+`bind`{.verilog} statement.  All bind statements are placed in a separate file
+named `bindings_<circuit name>.sv`.
+
+For example, consider the following circuit `Foo`{.firrtl}` with remotely
+instantiated module `Bar`{.firrtl}:
+
+``` firrtl
+circuit Foo :
+  module Bar :
+    input a: UInt<1>
+
+  module Foo :
+    input b: UInt<1>
+
+    remote_inst bar of Bar
+    remote_inst.a <= b
+```
+
+This will produce the following two SystemVerilog modules:
+
+``` verilog
+module Foo(
+  input b
+);
+endmodule
+
+module Bar(
+  input a
+);
+endmodule
+```
+
+Additionally a `bindings_Foo.sv` file will be produced with the following
+contents:
+
+``` verilog
+bind Foo bar bar(.a(b));
+```
+
+Future changes to this ABI document will enable more fine-grained control of
+remote instantiation.
+
 ## On Types
 
 Types are only guaranteed to follow this lowering when the Verilog type is on an
--- a/abi.md
+++ b/abi.md
@@ -149,7 +149,7 @@ Additionally a `bindings_Foo.sv` file will be produced with the following
 contents:
 
 ``` verilog
-bind Foo bar bar(.a(b));
+bind Foo Bar bar(.a(b));
 ```
 
 Future changes to this ABI document will enable more fine-grained control of
--- a/abi.md
+++ b/abi.md
@@ -127,8 +127,8 @@ circuit Foo :
   module Foo :
     input b: UInt<1>
 
-    remote_inst bar of Bar
-    remote_inst.a <= b
+    remoteinst bar of Bar
+    bar.a <= b
 ```
 
 This will produce the following two SystemVerilog modules:
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -16,7 +16,7 @@
       <item>is</item>
       <item>invalid</item>
       <item>inst</item>
-      <item>remote_inst</item>
+      <item>remoteinst</item>
       <item>of</item>
       <item>wire</item>
       <item>node</item>
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -16,6 +16,7 @@
       <item>is</item>
       <item>invalid</item>
       <item>inst</item>
+      <item>remote_inst</item>
       <item>of</item>
       <item>wire</item>
       <item>node</item>
--- a/spec.md
+++ b/spec.md
@@ -1856,7 +1856,7 @@ module.  (See the FIRRTL ABI for more information on this lowering.)
 Remotely instantiated modules must only contain input ports of passive types or
 probe types.
 
-Remote instantiation uses the `remote_inst`{.firrtl} syntax as shown below:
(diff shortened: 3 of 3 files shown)


spec.md
```diff
@@ -1847,6 +1847,33 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
 itself, either directly, or indirectly through instances of other modules it
 instantiates.
 
+## Remote Instances
+
+FIRRTL modules may also be remotely instantiated.  A remotely instantiated
+module will cause the instantiation to be emitted outside its instantiating
```
mwachs5, line 1853:
```suggestion
module will cause the instantiation to be emitted outside its instantiating
```

is there any "may" or "will" here? are there conditions where we might not choose to do that sort of lowering?

mwachs5: conversely, there are times where we choose to convert things to binds...

spec.md
```diff
@@ -1847,6 +1847,33 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
...
+module.  (See the FIRRTL ABI for more information on this lowering.)
+
+Remotely instantiated modules must only contain input ports of passive types or
```
mwachs5, line 1856: eats shoots and leaves? Would a comma before the or be more clarifying? I'm not sure if I am allowed to have an input probe port...?

spec.md
```diff
@@ -3679,7 +3706,7 @@ statement =
   | "regreset" , id , ":" , type , "," , expr , "," , expr , "," , expr ,
     [info]
   | memory
-  | "inst" , id , "of" , id , [ info ]
+  | ( "inst" | "remote_inst" ) , id , "of" , id , [ info ]
```
mwachs5, line 3709: would this be better expressed as a more general concept of "usage group" of modules? where passes might decide to do things to different groups based on their type of instantiation (omit them entirely, move them to different directories, do them as binds, keep them present, do inlining...)

seldridge: I'm trying to leave the door open for this without solving it right here. This is what the following language in ABI doc is hinting at:

> Future changes to this ABI document will enable more fine-grained control of remote instantiation.

I was talking with @darthscsi about this earlier and he was proposing:

```
group verif : 
  inst bar of Bar
```

As I didn't want to start allowing arbitrary things inside the group other than instances, he preferred remote instantiation over an instance-only group.

That said, we need some answer for how to put things in different bind files and how grouping works.

abi.md
```diff
@@ -110,6 +110,51 @@ Passive bundles shall be lowered to Verilog packed structs.
...
+``` firrtl
+circuit Foo :
+  module Bar :
+    input a: UInt<1>
+
+  module Foo :
+    input b: UInt<1>
+
+    remote_inst bar of Bar
+    remote_inst.a <= b
```
jackkoenig, line 131:
```suggestion
    remote_inst bar of Bar
    bar.a <= b
```

spec.md
```diff
@@ -1847,6 +1847,33 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
...
+``` firrtl
+circuit Foo :
+  module Bar :
+    input a: UInt<1>
+
+  module Foo :
+    input b: UInt<1>
+
+    remote_inst bar of Bar
+    remote_inst.a <= b
```
jackkoenig, line 1870:
```suggestion
    bar.a <= b
```

include/firrtl.xml
```diff
@@ -16,6 +16,7 @@
       <item>is</item>
       <item>invalid</item>
       <item>inst</item>
+      <item>remote_inst</item>
```
jackkoenig, line 19: This is the only keyword that has an `_`. Types use UpperCamelCase and other keywords just elide any punctuation (eg. extmodule, intmodule, defname). Should we maybe make it `remoteinst` or `extinst` or something else?

abi.md
seldridge: Whoops. This is my first time writing FIRRTL, apparently. 😬

include/firrtl.xml
seldridge: Changes to `remoteinst` in https://github.com/chipsalliance/firrtl-spec/pull/107/commits/a3654afff15b34b8c94f6a86868a8ece649262f9.

spec.md
mwachs5: I guess to me the moth logical way to think would be 

inst foo of Foo group "amazing"


Then something in compiler cmd line /ABI could decide in this lowering that "amazing" should be a verilog-style remote instantiation, or omitted entirely, or left in place.

Just seems strange to put such a verilog-specific lowering so tightly into the IR, but still with insufficient information to make use of it (if we agree the idea of grouping is useful)

seldridge: I'm playing with the `group` angle which may be a better way to do this.

I agree with what you're saying. This PR is currently providing a mechanism for remote instantiation without a mechanism for conditional instantiation. We likely need the latter.

seldridge:
The main problem with remote instantiation on the ABI side is that SystemVerilog disallows bind under bind (Section 23.11 of the 2012 spec). We could do conditional instantiation via an ABI That lowers to `ifdef`. 

However, there is no way to do this with a `bind` that is compatible with separate compilation. If a module contains a remote instantiation, then it cannot itself be remotely instantiated. This is incredibly weird as it doesn't compose sanely. I'm not currently seeing how to get bind emission without delaying it to a linking step.

seldridge:
I thought through the implications of doing this with a new `group` operation here: https://github.com/chipsalliance/firrtl-spec/pull/108 This could also work. What may be simplest is to combine the notion of a `defgroup` with the restrictions around what can be in a group, and implement this with remote instantiation. The same restrictions on remote instantiation would apply to groups (no remote instantiation under remote instantiation to enable bind emission and remote instances cannot drive things in their instantiating module, i.e., they cannot have outputs).

abi.md
```diff
@@ -110,6 +110,51 @@ Passive bundles shall be lowered to Verilog packed structs.
...
+module Foo(
+  input b
+);
+endmodule
+
+module Bar(
+  input a
+);
+endmodule
+```
```
uenoku, line 146: I'm not sure it's necessary to mention in the spec, but these `Foo` and `Bar` are empty modules which have caused a number of errors in downstream tools. Though that's codegen internals, but it might be worth mentioning that these modules must be erased/inlined for correctness somewhere.

seldridge: Good point. I was attempting to keep the example compact by keeping the modules empty. However, it occurred to me that these _should_ be removed by optimizations. I can see if there's a way to make the examples in this PR or in #108 obviously block removal without making them overly long.

seldridge:
Closing this in favor of "optional groups" (#108) as those provide more needed features that cannot be provided with remote instantiation. Namely, "optional groups" enable passing probes from one remote instantiation to another which is not obviously possible in a safe way with remote instantiation.

PR #108, opened 2023-05-15
[minor] Add groups

seldridge:
Alternative to adding remote instantiation.  This introduces a "group" concept which is a way of putting a bunch of statements in a box which are destined to live in a bind.

This introduces a "defgroup" which enables a circuit or external module (which are two things which should be fungible) to declare that it has groups.  Because SystemVerilog bind disallows bind-under-bind, we need some way of guaranteeing that a specific extmodule/circuit is not later bound.

There is a middle-ground solution which would add remote instantiation indicating a group (h/t @mwachs5) and `defgroup`.  This would likely simplify some of the complexity by adding `group`.  However, the same rules of a `group` should likely apply to a remote instance, e.g., a remote instance/group cannot drive anything in the circuit except possibly via force.

### DIFF
--- a/abi.md
+++ b/abi.md
@@ -110,6 +110,179 @@ Passive bundles shall be lowered to Verilog packed structs.
 Reference types in ports shall be logically split out from aggregates and named
 as though "Aggregate Type Lowering" was used.
 
+## On Groups
+
+The lowering convention of a declared optional group specifies how an optional
+group will be lowered.  Currently, the only lowering convention that must be
+supported is `"bind"`{.firrtl.}.  FIRRTL compilers may implement other
+non-standard lowering conventions.
+
+### Bind Lowering Convention
+
+The bind lowering convention is indicated by the `"bind"`{.firrtl} string on a
+declared group.  When using this convention, optional groups are lowered to
+separate modules that are instantiated via SystemVerilog `bind`{.verilog}
+statements ("bind instantiation").
+
+Each module that contains group instances will produce one additional module per
+group that has "internal" module convention.  I.e., the modules are private and
+have no defined ABI---the name of each module is implementation defined, the
+instantiation name of a bound instance is implementation defined, and any ports
+created on the module may have any name and may be optimized away.
+
+Practically speaking, additional ports of each generated module must be created
+whenever a value defined outside the group is used by the group.  The values
+captured by the group will create input ports.  Values defined in a group and
+used by a nested group will create output ports because SystemVerilog disallows
+the use of a bind instantiation underneath the scope of another bind
+instantiation.  The names of additional ports are implementation defined.
+
+For each optional group, one binding file will be produced for each group or
+nested group.  The circuit, group name, and any nested groups are used to derive
+a predictable filename of the binding file.  The file format uses the format
+below:
+
+``` ebnf
+filename = "groups_" , circuit , "_", root , { "_" , nested } , ".sv" ;
+```
+
+As an example, consider the following circuit with three optional groups:
+
+``` firrtl
+circuit Foo:
+  declgroup Group1, bind:
+    declgroup Group2, bind:
+      declgroup Group3, bind:
+```
+
+When compiled to Verilog, this will produce three bind files:
+
+```
+groups_Foo_Group1.sv
+groups_Foo_Group1_Group2.sv
+groups_Foo_Group1_Group2_Group3_.sv
+```
+
+The contents of each binding files must have the effect of including all code
+defined in a group or its parent groups.
+
+#### Example
+
+The end-to-end below example shows a circuit with one two groups that is lowered
+to Verilog.  This shows example Verilog output which is not part of the ABI, but
+is included for informative reasons.
+
+Consider the following circuit containing one optional group, `Group1`, and one
+nested optional group, `Group2`.  Module `Foo` contains one instantiation of
+module `Bar`.  Both `Foo` and `Bar` contain groups.  To make the example
+simpler, no constant propagation is done:
+
+``` firrtl
+circuit Foo:
+  declgroup Group1, bind:
+    declgroup Group2, bind:
+
+  module Bar:
+    output _notA: Probe<UInt<1>, Group1>
+    output _notNotA: Probe<UInt<1>, Group1.Group2>
+
+    wire a: UInt<1>
+    connect a, UInt<1>(0)
+
+    group Group1:
+      node notA = not(a)
+      define _notA = probe(notA)
+
+      group Group2:
+        node notNotA = not(notA)
+        define _notNotA = probe(notNotA)
+
+  module Foo:
+    inst bar of Bar
+
+    group Group1:
+      node x = bar._notA
+
+      group Group2:
+        node y = bar._notNotA
+```
+
+The following Verilog will be produced for the modules without the groups:
+
+``` verilog
+module Foo();
+  Bar bar();
+endmodule
+
+module Bar();
+  wire a = 1'b0;
+endmodule
+```
+
+The following Verilog associated with `Group1` is produced.  Note that all
+module names and ports are implementation defined:
+
+``` verilog
+module Bar_Group1(
+  input a
+);
+  wire notA = ~a;
+endmodule
+
+module Foo_Group1(
+  input bar_notA
+);
+  wire x = bar_notA;
+endmodule
+```
+
+The following Verilog associated with `Group2` is produced. Note that all module
+names and ports are implementation defined:
+
+``` verilog
+module Bar_Group1_Group2(
+  input notA
+);
+  wire notNotA = ~notA;
+endmodule
+
+module Foo_Group1_Group2(
+  input bar_notNotA
+);
+  wire y = bar_notNotA;
+endmodule
+```
+
+Because there are two groups, two bindings files will be produced.  The first
+bindings file is associated with `Group1`:
+
+``` verilog
+// Inside file "groups_Foo_Group1.sv" :
+`ifndef groups_Foo_Group1
+`define groups_Foo_Group1
+bind Foo Foo_Group1 group1(.bar_notA(Foo.bar.group1.notA));
+bind Bar Bar_Group1 group1(.a(Bar.a));
+`endif
+```
+
+The second bindings file is associated with `Group2`.  This group, because it
+depends on `Group1` being available, will automatically bind in this dependent
+group:
+
+``` verilog
+// Inside file "groups_Foo_Group1_Group2.sv" :
+`ifndef groups_Foo_Group1_Group2
+`define groups_Foo_Group1_Group2
+`include "groups_Foo_Group1.sv"
+bind Foo Foo_Group1_Group2 group1_group2(.bar_notNotA(Foo.bar.group1_group2.notNotA));
+bind Bar Bar_Group1_Group2 group1_group2(.notA(Bar.group1.notA));
+`endif
+```
+
+The `` `ifdef ``{.verilog} guards enable any combination of the bind files to be
+included while still producing legal SystemVerilog.  I.e., the end user may
+safely include none, either, or both of the bindings files.
+
 ## On Types
 
 Types are only guaranteed to follow this lowering when the Verilog type is on an
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -6,6 +6,7 @@
       <item>module</item>
       <item>extmodule</item>
       <item>intmodule</item>
+      <item>declgroup</item>
     </list>
     <list name="keywords">
       <item>input</item>
@@ -14,6 +15,7 @@
       <item>parameter</item>
       <item>skip</item>
       <item>inst</item>
+      <item>group</item>
       <item>of</item>
       <item>wire</item>
       <item>node</item>
@@ -148,7 +150,7 @@
       <context name="info" attribute="Info" lineEndContext="#stay">
         <DetectChar char="]" context="#pop"/>
       </context>
-      <context name="outertype" attribute="Keyword" lineEndContext="#pop">
+      <context name="outertype" attribute="ID" lineEndContext="#pop">
         <DetectChar char="&lt;" attribute="Operator" context="#stay"/>
         <keyword String="types" attribute="Keyword" context="type"/>
         <Detect2Chars char="{" char1="|" context="field" attribute="Separator"/>
--- a/spec.md
+++ b/spec.md
@@ -179,6 +179,138 @@ physically present in the final circuit. Refer to the description of the
 instance statement for details on how to instantiate a module
 ([@sec:instances]).
 
+## Optional Groups
+
+Optional groups are named collections of statements inside a module.  Optional
+groups contain functionality which will not be present in all executions of a
+circuit.  Optional groups are intended to be used to keep verification,
+debugging, or other collateral, not relevant to the operation of the circuit, in
+a separate area.  Each group can then be optionally included in the resulting
+design.
+
+The `declgroup`{.firrtl} keyword declares an optional group with a specific
+identifier.  An optional group may be declared in a circuit or in another
+optional group declaration.  An optional group's identifier must be unique
+within the current namespace.  I.e., the identifier of a top-level group
+declared in a circuit must not conflict with the identifier of a module,
+external module, or implementation defined module.
+
+Each optional group declaration must include a string that sets the lowering
+convention for that group.  The FIRRTL ABI specification defines supported
+lowering convention.  One such strategy is `"bind"`{.firrtl} which lowers to
+modules and instances which are instantiated using the SystemVerilog
+`bind`{.verilog} feature.
+
+The `group`{.firrtl} keyword defines optional functionality inside a module.  An
+optional group may only be defined inside a module.  An optional group must
+reference a group declared in the current circuit.  Declarations of identifiers
+and references to existing identifiers following the same lexical scoping rules
+as FIRRTL conditional statements (see: [@sec:scoping])---identifiers declared in
+the group definition may not be used outside the group while groups may refer to
+identifiers declared outside the group.  __The statements in a group are
+restricted in what identifiers they are allowed to drive.__ A statement in a
+group may drive no sinks declared outside the group _with one exception_: a
+statement in a group may drive reference types declared outside the group if the
+reference types are associated with the group in which the statement is declared
+(see: [@sec:reference-types]).
+
+The circuit below contains one optional group declaration, `Bar`.  Module `Foo`
+contains a group definition that creates a node computed from a port defined in
+the scope of `Foo`.
+
+``` firrtl
+circuit Foo:
+  declgroup Bar, bind:  ; Declaration of group Bar with convention "bind"
+
+  module Foo:
+    input a: UInt<1>
+
+    group Bar:            ; Definition of group Bar inside module Foo
+      node notA = not(a)
+```
+
+Optional group declarations may be nested.  Optional group declarations are
+declared with the `declgroup`{.firrtl} keyword indented under an existing
+`declgroup`{.firrtl} keyword.  The circuit below contains four optional group
+declarations, three of which are nested.  `Bar` is the top-level group.  `Baz`
+and `Qux` are nested under `Bar`.  `Quz` is nested under `Qux`.
+
+``` firrtl
+circuit Foo:
+  declgroup Bar, bind:
+    declgroup Baz, bind:
+    declgroup Qux, bind:
+      declgroup Quz, bind:
+```
+
+Optional group definitions must match the nesting of declared groups.  Optional
+groups are defined under existing groups with the `group`{.firrtl} keyword
+indented under an existing `group`{.firrtl} keyword.  For the four declared
+groups in the circuit above, the following is a legal nesting of group
+definitions:
+
+``` firrtl
+module Foo:
+  input a: UInt<1>
+
+  group Bar:
+    node notA = not(a)
+    group Baz:
+    group Qux:
+      group Quz:
+        node notNotA = not(notA)
+```
+
+Statements in a nested optional group may only read from ports or declarations
+of the current module, the current group, or a parent group---statements in a
+group may not drive components declared outside the group except reference types
+associated with the same group.  In the above example, `notA` is accessible in
+the group definition of `Quz` because `notA` is declared in a parent group.
+
+In the example below, module `Baz` defines a group `Bar`.  Module `Baz` has an
+output port, `_a`, that is associated with the group, `Bar`.  This port can then
+be driven from inside the group.  In module `Foo`, the port may be read from
+inside the group.  _Stated differently, module `Baz` has an additional port `_a`
+that is only accessible inside a defined group `Bar`_.
+
+``` firrtl
+circuit Foo:
+  declgroup Bar, bind:
+
+  module Baz:
+    output _a: Probe<UInt<1>, Bar>
+
+    wire a: UInt<1>
+
+    group Bar:
+      node notA = not(a)
+      define _a = probe(notA)
+
+  module Foo:
+
+    inst baz of Baz
+
+    group Bar:
+      node _b = baz._a
+```
+
+If a port is associated with a nested group then a period is used to indicate
+the nesting.  E.g., the following circuit has a port associated with the nested
+group `Bar.Baz`:
+
+``` firrtl
+circuit Foo:
+  declgroup Bar, bind:
+    declgroup Baz, bind:
+
+  module Foo:
+    output a: Probe<UInt<1>, Bar.Baz>
+```
+
+Optional groups will be compiled to modules whose ports are derived from what
+they capture from their visible scope.  For full details of the way optional
+groups are compiled, see the FIRRTL ABI specification.
+
 ## Externally Defined Modules
 
 Externally defined modules are modules whose implementation is not provided in
@@ -665,6 +797,11 @@ Probe references must always be able to be statically traced to their target,
 or to an external module's output reference.  This means no conditional
 connections via sub-accesses, multiplexers, or other means.
 
+Reference types compose with optional groups (see [@sec:optional-groups].  A
+reference type may be associated with an optional group.  When associated with
+an optional group, the reference type may only be driven from that optional
+group.
+
 ### Probe Types
 
 Probe types are reference types used to access circuit elements' data remotely.
@@ -686,6 +823,7 @@ Examples:
 ```firrtl
 Probe<UInt> ; readable reference to unsigned integer with inferred width
 RWProbe<{x: {y: UInt}}> ; readable and forceable reference to bundle
+Probe<UInt, A.B> ; readable reference associated with group A.B
 ```
 
 For details of how to read and write through probe types, see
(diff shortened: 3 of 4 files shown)


spec.md
```diff
@@ -210,7 +212,7 @@ extmodule MyExternalModule :
   parameter y = 42
 ```
 
-An example of an externally defined module with references is:
+An example of an externally defined module with references and groups is:
 
 ```firrtl
 extmodule MyExternalModuleWithRefs :
```
mwachs5, line 218: AndGroups?

spec.md
```diff
@@ -1841,6 +1845,48 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
 itself, either directly, or indirectly through instances of other modules it
 instantiates.
 
+## Groups
+
+Groups are named collections of statements inside a module.  Groups contain
+optional functionality for a module that can be included or excluded without
```
mwachs5, line 1851: Given this meaning, should the name of this concept be something like "optional group"?

spec.md
```diff
@@ -1841,6 +1845,48 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
...
+
+The `defgroup`{.firrtl} statement declares a group with a specific name.  A
+`defgroup`{.firrtl} may declare a group for a `circuit`{.firrtl} or for an
+`extmodule`{.firrtl}.  Group names participate specially in the module
+namespace--no module may include a suffix of `_<group-name>` for any declared
+group name.
+
+The `group`{.firrtl} statement creates a group instance within a module.  A
+`group`{.firrtl} must refer to a group declared by a `defgroup`{.firrtl} of the
+current `circtui`{.firrtl}.
```
mwachs5, line 1865: Circuit

spec.md
```diff
@@ -3690,6 +3738,9 @@ statement =
   | "connect" , reference , "," , expr , [ info ]
   | "invalidate" , reference , [ info ]
 
+(* Group definitions *)
+defgroup = "defgroup" , id ;
```
mwachs5, line 3742: What is this top level one for? Can you give an example?

seldridge: It's to declare the groups that can be used later. This makes it way easier to know what groups exist. Otherwise, you need to scan the entire IR and look for any group instances. This is also then used for check if a group instance is legal. If circuit `Foo` is conforming to an interface, it can only have certain legal groups. By making the groups declared up front this is easier.

Here's a bigger example. It would be illegal for any of the modules to have a "wildcat" group like `group qux of Qux` because that was not declared.

```
circuit Foo:
  defgroup Bar           ; declare the group Bar

  module Baz:
    input a: UInt<1>

    group bar of Bar:    ; create the group instance bar
      node notA = not(a)

  module Foo:
    input a: UInt<1>

    group bar of Bar:    ; create the group instance bar
      node notA = not(a)

    inst baz of Baz
    baz.a <= a
```

This will then produce the following Verilog:

```verilog
// Inside file "Foo.sv":
module Foo(
  input a
  Baz baz(.a(a));
);
endmodule


// Inside file "Baz.sv":
module Baz(
  input a
);
endmodule

// Inside file "Foo_Bar.sv":
module Foo_Bar(
  input a
);
  wire notA = ~a;
endmodule

// Inside file "Baz_Bar.sv":
module Baz_Bar(
  input a
);
  wire notA = ~a;
endmodule

// Inside file "bindings_Foo_Bar.sv" :
bind Foo Foo_Bar bar(.a(a));
bind Baz Baz_Bar bar(.a(a));
```

spec.md
seldridge: Possibly. I'll switch to that for now. I also considered "verification groups", but that may be too narrow for their use.

spec.md
seldridge: Fixed in https://github.com/chipsalliance/firrtl-spec/pull/108/commits/0a54caf039ac72960226a34c368eafeb2cc90911

spec.md
seldridge: Fixed in https://github.com/chipsalliance/firrtl-spec/pull/108/commits/0a54caf039ac72960226a34c368eafeb2cc90911

seldridge:
We had a discussion about this today with the following observations:

1. How does this work for hierarchy manipulations, e.g., deduplication and inlining? Can the resulting group modules be deduplicated?
2. Output probes from groups may be necessary to enable a testharness group from reading from the same group inside a design.
3. Groups could have a partial ordering meaning that one group may require another group. E.g., a "Verification" group may require the "Printing" group because it wants to tap from that group. Other groups may have no dependencies on other groups.
4. External modules that have output probes require information about which group they are in in order to avoid accessing something from the incorrect group.
5. Where are the XMRs from groups placed? It seems natural that these should live inside the bind file as they only exist if a specific group is bound in.

youngar:
I just started to read through this, but it reminds me of closures or nested modules which can reference their outer environment:
```
module Foo:
  input a: UInt<1>
  module Bar:
    print(a)
  inst bar of Bar
```
During lowering, anything referenced from the outer scope becomes some kind of port.  

It is sort of weird to consider side effecting statements like asserts as optional, since they change the simulation behaviour, but I get that they are normally stripped out for synthesis.

For your list of concerns above, it sounds like you are going to need optional output ports:
```
module Foo:
  input in : UInt<8>
  group Verification:
     output out : UInt<8>
     out <= in
```
And then presumably the enabled groups are passed in to the compiler as a flag.  This is kind of like saying the circuit is parameterized on which groups are enabled.

seldridge:
> It is sort of weird to consider side effecting statements like asserts as optional, since they change the simulation behaviour, but I get that they are normally stripped out for synthesis.

Agree. This is weird.

This proposal needs optional output _probe_ ports only (not output ports). Are these "phases"? Are they part of the type?

Yes, the enabled groups are like Verilog parameters set by including specific files.

seldridge:
I updated this with nested groups and to allow output, passive probes from groups.  Group nesting has to be pre-declared and external modules need to declare any ports in their groups.

The ABI is then pretty clean.  From a user perspective, you get one bind file per declared group.  Using an `ifdef`/`define` approach, each bindfile can then pull in any dependent binds that are necessary for it to work.  I.e., a nested group will bind its parent group if not already bound.  This enables arbitrary composition of groups by a user and things just work.

spec.md
seldridge: Switched to "optional groups".

mwachs5 (commented):
looks pretty good! Just a few clarifying questions

abi.md
```diff
@@ -110,6 +110,127 @@ Passive bundles shall be lowered to Verilog packed structs.
 Reference types in ports shall be logically split out from aggregates and named
 as though "Aggregate Type Lowering" was used.
 
+## On Groups
+
+Optional groups are lowered to separate modules that are instantiated via
+SystemVerilog `bind`{.verilog} statements.  Each module that contains group
+instances will produce an additional module whose name is derived from the group
+and nested group instance.  Nested groups are appended.  E.g.,
+`<module-name>$g<group-name>` is the name of a generated module for its first
```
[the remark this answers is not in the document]
mwachs5: actually it is more clear down below in the example that you really are appending all the names to the file.

I am still curious if we are trying to have some sort of namespace. Should we state that groups names need to be unique within a circuit, regardless of nesting? Because otherwise I'm not sure how there would not be ambiguity if you directly put a `Group2` thing into a module?

abi.md
```diff
@@ -110,6 +110,127 @@ Passive bundles shall be lowered to Verilog packed structs.
...
+group.
+
+For each optional group, one binding file will be produced with a name derived
+from the group and nested group instance.
+`bindings$c<circuit-name>$g<group-name>.sv`.  The instantiation name for the
```
mwachs5, line 124:
```suggestion
`bindings$c<circuit-name>$g<group-name>[$g<group-name>].sv`.  The instantiation name for the
```

mwachs5: im not sure that nesting the group names in the file names is getting us anything?

spec.md
```diff
@@ -211,15 +213,17 @@ extmodule MyExternalModule :
...
 ```firrtl
-extmodule MyExternalModuleWithRefs :
+extmodule MyExternalModuleWithRefsAndGroups :
   input foo : UInt<2>
   output mysignal : Probe<UInt<1>>
   output myreg : RWProbe<UInt<8>>
   ref mysignal is "a.b"
   ref myreg is "x.y"
+  defgroup MyGroup
+  defgroup MyOtherGroup
```
mwachs5, line 226: can you show what this looks like for a nested group?

spec.md
```diff
@@ -1925,6 +1929,123 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
 itself, either directly, or indirectly through instances of other modules it
 instantiates.
 
+## Optional Groups
```
mwachs5, line 1932: should this Optional Group chapter go before ExtModule since we refer to it in the ExtModule chapter?

spec.md
```diff
@@ -1925,6 +1929,123 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
...
+  module Foo:
+
+    inst baz of Baz
+
+    group bar of Bar:
+      node _b = baz._a
+```
+
+External modules may also have output, passive ports of probe type.  The
+following example shows this:
```
mwachs5, line 2031:
```suggestion
External modules may also have output, passive ports of probe type that are provided by an optional group.  The
following example shows this:
```

spec.md
```diff
@@ -1925,6 +1929,123 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
...
+```
+
+Groups may be nested.  Nested groups are declared with the `defgroup`{.firrtl}
+statement indented under an existing `defgroup`{.firrtl}.  The circuit below
+contains four groups.  `Bar` is the top-level group.  `Baz` and `Qux` are nested
+under `Bar`.  `Quz` is nested under `Qux`.
+
+``` firrtl
+circuit Foo:
+  defgroup Bar:
```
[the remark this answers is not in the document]
seldridge: They are now opening a possibly empty scope. It's the same syntax as empty module (which is much rarer):

```
circuit Foo:
  module Bar:
  module Foo:
```

Maybe there's a better approach, e.g., `:` is optional if there is no scope or this could use 😨 `{}`.

I'd like to keep this consistent with how other scopes work (module, when, etc.) and then do a batch syntax change for all scopes.

spec.md
```diff
@@ -1925,6 +1929,123 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
...
+
+  module Foo:
+
+    inst baz of Baz
+
+    group bar of Bar:
+      node _b = baz._a
+```
+
+External modules may also have output, passive ports of probe type.  The
```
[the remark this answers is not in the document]
seldridge: Yes. Will clarify. 👍

spec.md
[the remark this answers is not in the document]
mwachs5: for some reason i thought you had to write `pass` for an empty module

spec.md
```diff
@@ -3789,6 +3912,9 @@ statement =
   | "connect" , reference , "," , expr , [ info ]
   | "invalidate" , reference , [ info ] ;
 
+(* Group definitions *)
+defgroup = "defgroup" , id ;
```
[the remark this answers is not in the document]
seldridge: Will fix this in a later commit.

spec.md
```diff
@@ -1925,6 +1929,123 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
...
+
+``` firrtl
+extmodule Foo:
+  defgroup Bar:
+    output a: Probe<UInt<1>>
+```
+
+To summarize, optional groups have the following rules:
+
+1. Modules or external modules which contain groups may not be instantiated in
```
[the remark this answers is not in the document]
seldridge: I'll update this. 👍

spec.md
```diff
@@ -1925,6 +1929,123 @@ To disallow infinitely recursive hardware, modules cannot contain instances of
...
+   a group.
+1. Connections in an optional group instance must not connect to sinks declared
+   outside the group or in parent groups.
+1. All other statements and expressions are allowed inside a group instance.
+1. Group instances may have output ports of passive, probe type.
```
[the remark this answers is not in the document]
seldridge: This will get updated in a later commit.

abi.md
seldridge: I'm actually not super keen on this, but there is a big benefit of doing it this way: this allows for the group names to conflict with each other. If the group names _aren't_ appended, then you need to unique these in some way. I.e., this allows:

```
circuit Foo:
  defgroup A:
    defgroup A:
```

Otherwise, all group names, including nested names, need to be unique. This may be fine.

spec.md
seldridge: Will add this in a later commit.

spec.md
```diff
@@ -2044,6 +2049,10 @@ To summarize, optional groups have the following rules:
    outside the group or in parent groups.
 1. All other statements and expressions are allowed inside a group instance.
 1. Group instances may have output ports of passive, probe type.
+1. A first module that instantiates a second module may access ports of the
+   second module declared in the second module's group instance within a group
+   instance in the first module if the group instances are the same or if the
```
mwachs5, line 2054:
```suggestion
   instance in the first module if the group instances are of the same group or if the
```

spec.md
```diff
@@ -2044,6 +2049,10 @@ To summarize, optional groups have the following rules:
...
+   second module's group instance is a parent of the first module's group.
```
mwachs5, line 2055:
```suggestion
   first module's group instance is within the nested group hierarchy of the second module's group.
```

mwachs5: or something -- now that we have two hierarchy systems going i found this confusing the way it was written

spec.md
```diff
@@ -3936,6 +3945,11 @@ intmodule = "intmodule" , id , ":" , [ info ] , newline , indent ,
               { "parameter" , "=" , ( int | string_dq ) , newline } ,
             dedent ;
 
+(* Group definitions *)
+defgroup = "defgroup" , id , ":" , [ info ] , newline , indent ,
+             { port , newline } ,
```
mwachs5, line 3950: i dont fully understand this syntax, but shouldn't it have a { defgroup }, here?

seldridge: Yep.

abi.md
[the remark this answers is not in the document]
seldridge: I'll clean up this section in a follow-on.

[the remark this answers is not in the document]
seldridge: Mentioned below, but this appending avoids the need to have every group name be unique. I think the issue of name collisions is avoided as a `group` has to follow the existing `declgroup` pattern. You can't do the following because `group B` is not declared:

```
circuit Foo:
  declgroup A:
    declgroup B:
  module Foo:
    group B:
```

However, you could do the following:

```
circuit Foo:
  declgroup A:
    declgroup B:
  declgroup B:
  module Foo:
    group B:
```

spec.md
seldridge: I'll rework this to do that. This section was written before the `declgroup` declaration was added which resulted in the ordering.

abi.md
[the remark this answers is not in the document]
mwachs5: ok so the proper way to do the first example correctly then is this:

```
circuit Foo:
  declgroup A:
    declgroup B:
  module Foo:
    groupA:
      group B:
```

?

abi.md
mwachs5: i feel like the unique name thing might be easier to implement 🤷‍♀️

abi.md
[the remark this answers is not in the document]
seldridge: Maybe this is easier to just state that all groups must be unique in the circuit namespace. Will try this as it's much simpler.

spec.md
seldridge: This condition is difficult to write. Cleaning this up in `fixup!` commit.

abi.md
seldridge: Going to just flatten this out. If group names, including nested group names, are unique in the module namespace, then we can do this. From the perspective of a user, they aren't thinking about nested groups, just this circuit has all these groups and I can choose any of them to bind in.

abi.md
```diff
@@ -114,25 +114,28 @@ as though "Aggregate Type Lowering" was used.
...
-elaboration-time choice by an end user.
+instances will produce an additional module whose name is implementation
+defined.
+
+For each optional group, one binding file will be produced for each group or
+nested group.  The circuit and group name are used to derive a predictable
+filename of the binding file.  The file format uses the format below:
+
+```ebnf
+filename = "groups_" , ciruict , "_", group , ".sv"
```
mwachs5, line 125:
```suggestion
filename = "groups_" , circuit , "_", group , ".sv"
```

mwachs5 (approved):
sgtm!

mwachs5:
(dont forget the changelog edits!)

include/firrtl.xml
```diff
@@ -6,6 +6,7 @@
       <item>module</item>
       <item>extmodule</item>
       <item>intmodule</item>
+      <item>defgroup</item>
```
mwachs5, line 9:
```suggestion
      <item>declgroup</item>
```

abi.md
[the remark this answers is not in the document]
seldridge: > ok so the proper way to do the first example correctly then is this:

That's correct.

spec.md
```diff
@@ -179,17 +179,155 @@ physically present in the final circuit. Refer to the description of the
...
+External modules may also have output, passive ports of probe type declared for
+their optional groups.  The following example shows declared group `Bar` with an
+output port `a`.
+
+``` firrtl
+extmodule Foo:
+  declgroup Bar:
+    output a: Probe<UInt<1>>
+```
+
```
seldridge, line 298: It is slightly inconsistent that `declgroup` can have ports when in an external module, but not have ports when in a circuit. Given this, it would be better to use a different keyword here, e.g., `extgroup`, that is specific to declaring a group with ports inside an external module.

mwachs5: `hasgroup` ? `usesgroup`?

jackkoenig: I like `extgroup`. I would also be okay with `group` (makes more sense to me than `declgroup` for this use), but having a different keyword is probably the clearest thing to do.

jackkoenig: Offline @seldridge mention that there are potential namespace issues with groups defined in extmodules so `group` definitely doesn't make sense, I see how `declgroup` made some sense but yeah I think `extgroup` is probably the best option.

seldridge:
There are a few updates to this after talking with @darthscsi:

1. In the ABI, I switched back to using bind file names that include the full group and nested group, e.g., `groups_Circuit_RootGroup_NestedGroup1_NestedGroup2.sv`. This avoids the oddity of group names all have to be unique. This enables (possibly questionable, but better composable) things like having a group `Foo` nested under another group named `Foo`. There's no real reason why we should disallow this.
2. I made it explicit that identifiers declared within groups cannot conflict with identifiers declared outside the group. This is different from what I discussed with @darthscsi, but it solves to problems: (a) this keeps groups consistent with scoping rules around when blocks and (b) it avoids needing to either provide a way to access a masked name (e.g., `this.foo`) or not providing a means to access a masked name and requiring users to user temporaries to get around this. I'd like to revise the spec to define different scoping rules. However, I'd like to do this together with revising `when` to allow masking.
3. The group declaration now includes a lowering convention string. The only one defined for now is `"bind"`. I expect we'll want to define three more in the future: (a) "inline" which inlines the body, (b) "outline" which creates new modules/instances, and (c) "ifdef" which inlines the statements, but guards them with Verilog `ifdef`s.

abi.md
```diff
@@ -110,6 +110,178 @@ Passive bundles shall be lowered to Verilog packed structs.
...
+        define _notNotA = probe(notNotA)
+
+  module Foo:
+    inst bar of Bar
+
+    group Group1:
+      node x = bar.notA
+
+      group Group2:
+        node y = bar.notNotA
```
jackkoenig, line 206:
```suggestion
  module Foo:
    inst bar of Bar

    group Group1:
      node x = bar._notA

      group Group2:
        node y = bar._notNotA
```

I think this is currently referencing the wrong signals? Also I think the Verilog has the same inconsistency.

spec.md
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
 instance statement for details on how to instantiate a module
 ([@sec:instances]).
 
+## Optional Groups
+
+Optional groups are named collections of statements inside a module.  Optional
+groups contain optional functionality for a module that can be included or
+excluded without affecting the functionality of a circuit.  Optional groups are
```
darthscsi, line 186: I understand what you are going for, but maybe better wording?  I'm not sure quite what.  "Optional groups contain functionality which will not be present in all executions of a circuit."

spec.md
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
...
+used to keep verification, debugging, or other collateral, not relevant to the
```
darthscsi, line 187: "are intended to be used to keep"?

spec.md
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
...
+operation of the circuit, in a separate area.  Each group can then be optionally
+included in the resulting design.
+
+The `declgroup`{.firrtl} keyword declares an optional group with a specific
+identifier.  An optional group may be declared in either a circuit, and external
```
darthscsi, line 192: /and/an/

spec.md
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
...
+module, or in another optional group declaration.  An optional group's
```
darthscsi, line 193: how do names work with groups declared in an external module?  do they alias and must have the same implementation as local?  Are they usable from the local circuit?  

I'm wondering if instead of adding bodies to external modules, we declare them using a declgroup root:
`extmodule Foo using declgroup A` with declgroup A being just a normal declaration.  Then there wouldn't be confusion about multiple conflicting As  (linking is then not just textural substitution, but with named types it already might not be).

spec.md
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
...
+convention for that group.  The FIRRTL ABI specification defines supported
+lowering convention.  One such strategy is `"bind"`{.firrtl} which lowers to
+modules and instances which are instantiated using the SystemVerilog
+`bind`{.verilog} feature.
+
+The `group`{.firrtl} keyword defines an optional group.  An optional group may
+only be defined inside a module.  An optional group must reference a group
+declared in the current circuit.  Statements in an optional groups may only read
+from ports or declarations of the current module or the current
+group---statements in a group may not drive components declared outside the
```
darthscsi, line 208: Maybe explicitly mention names are resolved via lexical scoping, as per when and match.

spec.md
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
...
+group.  Declarations of identifiers in a group may not collide with identifiers
```
darthscsi, line 209: The no-drive-outside scope is a big change which should be highlighted.  Maybe this is enough.

spec.md
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
...
+  module Foo:
+    input a: UInt<1>
+
+    group Bar:            ; Definition of group Bar inside module Foo
+      node notA = not(a)
+
+  extmodule Baz:
+    input a: UInt<1>
+    output b: UInt<1>
+    declgroup "bind" Bar: ; Declaration of group Bar for extmodule Baz
```
darthscsi, line 229: Looking at this, I do think not having the declgroups in extmodules might be cleaner:
```
  extmodule Baz using declgroup Bar:
    input a: UInt<1>
    output b: UInt<1>
```
As then you don't have to specify how two declgroups with different implementation strings interact.

darthscsi: Or alternately, maybe we just don't need to know this information?

spec.md
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
...
+of the current module, the current group, or a parent group---statements in a
+group may not drive components declared outside the group.  In the above
+example, `notA` is accessible in the group definition of `Quz` because `notA` is
+declared in a parent group.
+
+Group instances may declare output, passive ports of `Probe`{.firrtl} type.
+These ports are accessible in parent modules which instantiate the child.  In
+the example below, module `Baz` defines a group `Bar` with an output port `_a`.
+This port is accessible inside module `Foo` within its defined group `Bar`.
+_Put differently, module `Baz` has an additional port `_a` that is only
```
darthscsi, line 274: Stray "_".

spec.md
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
...
+
+``` firrtl
+circuit Foo:
+  declgroup "bind", Bar:
+
+  module Baz:
+    wire a: UInt<1>
+
+    group Bar:
+      output _a: Probe<UInt<1>>
```
darthscsi, line 285: I was going to ask how _a was accessed.  It might be better to not have group outputs implicitly part of module outputs.
```
  module Baz:
    output _a: Probe<UInt<1>>
    wire a: UInt<1>
   
    group Bar:
      output _a: Probe<UInt<1>>
      node notA = not(a)
      define _a = probe(notA)

  define _a <= Bar._a
  ```

File: spec.md [hunk 31]
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
...
+    group Bar:
+      node _b = baz._a
+```
+
+External modules may also have output, passive ports of probe type declared for
+their optional groups.  The following example shows declared group `Bar` with an
+output port `a`.
+
+``` firrtl
+extmodule Foo:
```
darthscsi, line 302: Which you do here.

spec.md
```diff
@@ -179,17 +179,163 @@ physically present in the final circuit. Refer to the description of the
...
+  declgroup "bind", Bar:
+    output a: Probe<UInt<1>>
+```
+
+To summarize, optional groups have the following rules:
+
+1. The identifiers of optional group declarations must be unique within the
```
darthscsi, line 309: I don't know that the summary is a good idea.  There isn't that much text above.

darthscsi (commented):
Comments and opinions changed as I went through, so read all of them first.

I think my only major thing would be to have extmodules have groups, not declgroups.

spec.md
seldridge: I'll make this very explicit. Good thinking.

spec.md
seldridge: I'll add a forward reference to it. It's annoying that this is a forward reference, though... (Maybe we can restructure the spec to not require such cyclic references, e.g., define scoping very early on.)

spec.md
seldridge: This is making the sentence italicized. It is terminated at the end of the sentence. 😉

spec.md
seldridge: I want to indicate that the optional output is associated with the group which is why I didn't want to put them on the module. However, I could use something more direct (which is how I would implement this in a compiler)---each probe port has an optional group associated with it:

```
module Baz:
  output _a: Probe<UInt<1>, Bar>
  wire a: UInt<1>

  group Bar:
    node notA = not(a)
    define _a = probe(notA)
```

WDYT?

spec.md
seldridge: I'll strike it.

spec.md
seldridge: Switched to this in the latest version.

spec.md
seldridge: I'm going to punt on how external modules play into this for now. This will be handled in a later PR.

PR #109, opened 2023-05-16
[major] Disallow self-referential registers

seldridge:
Remove a tedious parsing issue stemming from Chisel emitting reset-less registers with a reference to themselves.  This construction is shown below:

    reg r: UInt<1>, clock with (reset => (UInt<1>(0), r))

This construction requires special parsing unless the identifier "r" is entered into the namespace immediately upon parsing.  It is easier to just disallow this and require FIRRTL-targeting languages to either emit reset-less registers or use wires for situations where they have true self-referential behavior.

Note: I'm changing Chisel to stop emitting this here: https://github.com/chipsalliance/chisel/pull/3280

Further note: The above construction is currently special-cased in CIRCT, but somebody could hand-write FIRRTL with weird expressions that include the register in any of the clock, reset signal, or reset value which are easier to just avoid.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1207,6 +1207,20 @@ reg myreg: SInt, myclock with: (reset => (myreset, myinit))
 A register is initialized with an indeterminate value (see
 [@sec:indeterminate-values]).
 
+A register may not be declared with references to itself.  Any circuit that
+requires this must use a wire as a temporary.  E.g., the following shows an
+illegal register that references itself and an equivalent version that avoids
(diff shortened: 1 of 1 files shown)


seldridge:
I'm going to close this as this _should_ be safe to allow. There is some utility in this as a possible shorthand representation and parsing is relatively simple. Names must be added to the namespace while parsing of certain statements are incomplete. This can happen for nodes and registers.

PR #114, opened 2023-06-15
Remove unintended truncation behavior

darthscsi:
truncation was intended to be limited to partial connect, not connect.  This was changed 5 months ago.

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -19,6 +19,7 @@ revisionHistory:
         construction of hardware literals.
       - Change string-encoded integers to radix-encoded integers.
       - Remove legacy connect (`<=`) and invalidate (`is invalid`) syntax
+      - Make connect disallow implicit truncation (again).
     abi:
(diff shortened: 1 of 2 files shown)


seldridge (approved):
Truncation was always the behavior of all FIRRTL compilers even when the spec said differently. 🥲 It's great that we can finally change this.

Please add a `[major]` marker and update the revision history. I can get this into the 3.0.0 release.

seldridge:
The background here being that all FIRRTL compilers never correctly implemented the "connect does not truncate" behavior. I removed the spec fan-fiction here and opted for the gritty reality. I.e., the following has always, unfortunately compiled:

```
circuit Foo:
  module Foo:
    input a: UInt<2>
    output b: UInt<1>

    b <= a
```

To the following Verilog:
```verilog
module Foo(
  input  [1:0] a,
  output       b
);

  assign b = a[0];
endmodule


```

PR #116, opened 2023-06-28
Add optional desiredName to spec

adkian-sifive:
The need for `desiredName` arose from the fact that there is no way to form logical groupings of modules - all groupings are derived implicitly by backends. For example, the CIRCT backend groups modules for deduplication based on the arity and types of ports in modules. 

Desired name is an optional attribute which will allow users to manually specify an identifier for module grouping. For example, all three of the following modules in

```
circuit Foo:
  module Foo
    pass

  module Foo_1
    pass
  
  module Bar
    pass
```
will get grouped into a single module. With the optional desiredName, 

```
circuit Foo:
  module Foo | Foo
    pass

  module Foo_1 | Foo
    pass
  
  module Bar | Bar
    pass
```

the user has the capability to manually group modules into two groups: `Foo` and `Bar`. 

The reason this new identifier is called desiredName (and not something like "groupName") is to maintain compatibility with the existing Chisel `desiredName`, which is the most likely candidate to get emitted as the FIRRTL desiredName identifier.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -158,26 +158,29 @@ circuit MyTop :
 
 ## Modules
 
-Each module has a given name, a list of ports, and a list of statements
-representing the circuit connections within the module. A module port is
-specified by its direction, which may be input or output, a name, and the data
-type of the port.
+Each module has a given name, an optional desired name, a list of
+ports, and a list of statements representing the circuit connections
+within the module. A module port is specified by its direction, which
+may be input or output, a name, and the data type of the port.
 
 The following example declares a module with one input port, one output port,
 and one statement connecting the input port to the output port.  See
 [@sec:connects] for details on the connect statement.
 
 ``` firrtl
-module MyModule :
+module MyModule | MyModuleDesiredName :
   input foo: UInt
   output bar: UInt
   connect bar, foo
 ```
 
-Note that a module definition does *not* indicate that the module will be
-physically present in the final circuit. Refer to the description of the
-instance statement for details on how to instantiate a module
+Note that a module definition does *not* indicate that the module will
+be physically present in the final circuit. Refer to the description
+of the instance statement for details on how to instantiate a module
 ([@sec:instances]).
+
+The desired name provides an additional attribute for modules which
+may be used by backends to group modules.
 
 ## Externally Defined Modules
 
@@ -3791,7 +3794,7 @@ statement =
 
 (* Module definitions *)
 port = ( "input" | "output" ) , id , ":" , type , [ info ] ;
-module = "module" , id , ":" , [ info ] , newline , indent ,
+module = "module" , id , [ "|", id ] , ":" , [ info ] , newline , indent ,
            { port , newline } ,
            { statement , newline } ,
          dedent ;


spec.md
```diff
@@ -158,27 +158,30 @@ circuit MyTop :
...
-Note that a module definition does *not* indicate that the module will be
-physically present in the final circuit. Refer to the description of the
-instance statement for details on how to instantiate a module
+Note that a module definition does *not* indicate that the module will
+be physically present in the final circuit. Refer to the description
+of the instance statement for details on how to instantiate a module
 ([@sec:instances]).
 
+The desired name provides an additional attribute for modules which
+may be used by backends to group modules.
```
mwachs5, line 183: i dont know if "group" is the right verb here. Do we mention dedup at all in the spec / ABI doc?

adkian-sifive: Nope, there's no mention of dedup

seldridge:
As an alternative, we can do this with output properties (possibly marked `dontTouch`). That has the same empirical effect in CIRCT and avoids treating `desiredName` specially in the spec.

mwachs5:
Hm, i didn't get the impression we were planning to examine properties within the firrtl compiler (are properties a substitute for annotations? Should we just say there is a new annotation for "onlyDedupWithGroup"?)

As far as this PR goes I'd rather be more explicit about stating this is for dedup purposes (or what other purpose is it for?) (which requires adding something to the ABI doc explaning that dedup is a thing?)

the fact that we are using desired name to derive a dedup group is convenience on the chisel side i think, unless we are saying `module Cow | Moo` has some expectation to be called `Moo`?

mwachs5:
ok, I'm on board with the property based approach and it makes sense that it will seamlessly work with dedup without having to explain dedup is a thing, but properties aren't in the spec yet so we should probably push on that

jackkoenig:
FWIW, desired name could have an additional use case which is renaming modules after Dedup, I believe this would make naming somewhat more stable.

I agree that it's nice to not have to add a new concept though so perhaps we should just go with properties instead.

darthscsi:
using an output property for this is not conserving concepts.  It's taking one concept and making special-case interpretations of it for another use.  Using a property with a magic name is not really any different from using a port with a magic value.  What if "output OptLevel : UInt" meant "optimization level of the module, and oh, by the way, don't emit this port"?  Properties should not be abused to be side-channels into passes.  (this is just annotations in disguise, but worse now because it overlaps with something the user is expecting to control)

darthscsi:
Please describe the behavior of "desired name" referencing only things in the firrtl spec.  E.g. do not say "dedup" in your explanation.  (or alternately "module group" or whatever).  If you cannot, it doesn't seem appropriate to add to the spec.

Dedup is a workaround for a hole in Chisel (which does not exist in firrtl).  Let's not enshrine this bug into the firrtl spec.

And practically, dedup, renaming, inlining, prefixing, etc all have poorly specified interactions.  Adding yet another source of names without specifying how all that is suppose to work is just making the situation worse.

mwachs5:
I though @seldridge was saying we don't have to add anything to the spec (or change any implementations) for this if we go the property route. Dedup remains a circt implementation detail and no special consideration needs to be applied for that particular property.

jackkoenig:
@darthscsi your comments are sort of at odds with each other

> Properties should not be abused to be side-channels into passes. 

vs. 

> Dedup is a workaround for a hole in Chisel (which does not exist in firrtl). Let's not enshrine this bug into the firrtl spec.

I don't disagree on either account but there are things we already do and rely on and we should be able to improve the status quo without boiling the ocean. I think it would be justifiable to spec out dedup, inlining, and all of these support important features (even though they're ugly) that every user of FIRRTL relies upon. Alternatively, if we wish to maintain the status quo there, then abusing properties might be the correct workaround for improving the behavior of these things without having to commit to the behavior long term via the spec.

I'm honestly fine going either way, but we need to be able to improve behavior without having to fix every problem first.

seldridge:
Alternative approach after discussing offline with @darthscsi: let's just use an annotation for this. Annotations are generally fine as long as they meet the following criteria:

1. An annotation is only understood by a single pass.
2. An annotation is completely removed by that pass.

I.e., annotations are okay if they are effectively "pass arguments".

What would this look like as a `DedupGroupAnnotation`?

My original point was that Chisel can implement this behavior using only a `DontTouchAnnotation` and a property. However, this is relying on a FIRRTL compiler removing that property. Chisel could do a similar thing with a `DontTouchAnnotation` on a node with a unique identifier indicating the dedup group. Chisel has enough information to emit this. However, it's still pretty janky. The annotation approach should be the cleanest.

jackkoenig:
> let's just use an annotation for this. 

Sounds reasonable to me.

> What would this look like as a DedupGroupAnnotation?

Simplest representation would be just `case class DedupGroupAnnotation(module: ModuleTarget, group: String)` with the behavior being that modules with a `DedupGroupAnnotation` can only deduplicate with other modules in the same group. If no group is listed for a Module, then it just belongs in the general or universal group which can all deduplicate together.

We could define a more complex Annotation, eg. `DedupGroupsAnnotation(groups: Seq[(String, Seq[ModuleTarget])])`, essentially just `groupByIntoSeq` called on a `Seq[DedupGroupAnnotation]`. I'd prefer the simpler annotation and just have the pass create the more complex datastructure as it needs it, but either is easy enough for Chisel to emit.

PR #119, opened 2023-07-11
[NFC] Update formatting on match grammar

darthscsi:
The prior formatting made it easy to miss the statement group was split with a newline.


dtzSiFive (approved):
LGTM!

As I read this, this does not change the grammar, only uses whitespace to clarify the existing grammar.

Thanks!

PR #121, opened 2023-07-12
Add initial description of property types.

mikeurbach:
This PR is to open the discussion about a new kind of type: property types. It describes property types, without adding any concrete new types. My hope is to use this PR to at least get commitment that these types will exist, and add concrete property types and related constructs to the spec in future PRs.

### DIFF
--- a/abi.md
+++ b/abi.md
@@ -142,7 +142,8 @@ payload as the first field and a packed bit vector as a second field.  The
 padding for each payload is set to ensure all padded payloads have the same bit
 width as required by Verilog packed unions.
 
-Property types have no defined ABI.
+Property types have no defined ABI, and may not affect any other guarantees of
+the ABI.
 
 # Versioning Scheme of this Document
 
--- a/abi.md
+++ b/abi.md
@@ -142,6 +142,8 @@ payload as the first field and a packed bit vector as a second field.  The
 padding for each payload is set to ensure all padded payloads have the same bit
 width as required by Verilog packed unions.
 
+Property types have no defined ABI.
+
 # Versioning Scheme of this Document
 
 This is the versioning scheme that applies to version 1.0.0 and later.
--- a/spec.md
+++ b/spec.md
@@ -348,7 +348,7 @@ FIRRTL has four classes of types: _ground_ types, _aggregate_ types, _reference_
 types, and _property_ types.  Ground types are fundamental and are not composed
 of other types.  Aggregate types and reference types are composed of one or more
 aggregate or ground types.  Reference types may not contain other reference
-types. Property types represents information about the circuit that is not
+types. Property types represent information about the circuit that is not
 hardware.
 
 ## Ground Types
@@ -943,13 +943,15 @@ FIRRTL property types represent information about the circuit that is not
 hardware. This is useful to capture domain specific knowledge and design intent
 alongside the hardware description within the same FIRRTL.
 
-Property types cannot affect hardware. They only exist to augment the hardware
-description with extra information.
+Property types cannot affect hardware functionality or the hardware ABI. They
+cannot be used in any hardware types, including aggregates. They only exist to
+augment the hardware description with extra information.
 
 Handling of property types is completely implementation defined. A valid FIRRTL
 compiler implementation may do anything with property types as long as the
-existence of property types does not affect the emitted hardware description.
-For example, it is valid to drop property types from the IR completely.
+existence of property types does not affect hardware functionality or the
+hardware ABI. For example, it is valid to drop property types from the IR
+completely.
 
 ## Type Modifiers
 
--- a/spec.md
+++ b/spec.md
@@ -940,14 +940,14 @@ module Top:
 ## Property Types
 
 FIRRTL property types represent information about the circuit that is not
-hardware. This is useful to capture domain specific knowledge and design intent
+hardware. This is useful to capture domain-specific knowledge and design intent
 alongside the hardware description within the same FIRRTL.
 
 Property types cannot affect hardware functionality or the hardware ABI. They
-cannot be used in any hardware types, including aggregates. They only exist to
-augment the hardware description with extra information.
+cannot be used in any hardware types, including aggregates and references. They
+only exist to augment the hardware description with extra information.
 
-Handling of property types is completely implementation defined. A valid FIRRTL
+Handling of property types is completely implementation-defined. A valid FIRRTL
 compiler implementation may do anything with property types as long as the
 existence of property types does not affect hardware functionality or the
 hardware ABI. For example, it is valid to drop property types from the IR
--- a/spec.md
+++ b/spec.md
@@ -344,11 +344,12 @@ integer literals.  Any use in place of an integer is disallowed.
 
 # Types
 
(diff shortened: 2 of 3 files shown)


spec.md
```diff
@@ -936,6 +937,20 @@ module Top:
...
+alongside the hardware description within the same FIRRTL.
+
+Property types cannot affect hardware. They only exist to augment the hardware
+description with extra information.
+
+Handling of property types is completely implementation defined. A valid FIRRTL
+compiler implementation may do anything with property types as long as the
+existence of property types does not affect the emitted hardware description.
+For example, it is valid to drop property types from the IR completely.
+
```
seldridge, line 953: This is good, though is a bit cagey. Are there any examples that we can include in this PR to make things more obvious. E.g., `BigInt` and `String` type?

Or: are you planning to split this out in a very fine-grained way over multiple PRs?

spec.md
```diff
@@ -1064,6 +1081,8 @@ In order for a connection to be legal the following conditions must hold:
 3.  Either the flow of the right-hand side expression is source or duplex, or
     the right-hand side expression has a passive type.
 
+4.  The left-hand side and right-hand side types are not property types.
```
seldridge, line 1084: Are property types allowed in normal FIRRTL aggregates? If so, then this should indicate that neither LHS nor RHS can contain property types. If not, then some language in the property types section would be good.

spec.md
```diff
@@ -344,11 +344,12 @@ integer literals.  Any use in place of an integer is disallowed.
...
 
-FIRRTL has three classes of types: _ground_ types, _aggregate_ types, and
-_reference_ types.  Ground types are fundamental and are not composed of other
-types.  Aggregate types and reference types are composed of one or more
+FIRRTL has four classes of types: _ground_ types, _aggregate_ types, _reference_
+types, and _property_ types.  Ground types are fundamental and are not composed
+of other types.  Aggregate types and reference types are composed of one or more
 aggregate or ground types.  Reference types may not contain other reference
-types.
+types. Property types represents information about the circuit that is not
```
mwachs5, line 351:
```suggestion
types. Property types represent information about the circuit that is not
```

abi.md
```diff
@@ -142,6 +142,8 @@ payload as the first field and a packed bit vector as a second field.  The
 padding for each payload is set to ensure all padded payloads have the same bit
 width as required by Verilog packed unions.
 
+Property types have no defined ABI.
+
```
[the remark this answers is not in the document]
mikeurbach: Yeah, I think this makes sense to add here as well. Something like "Property types have no defined ABI, and may not affect any other guarantees of the ABI."?

spec.md
mikeurbach: I was planning to multiple fine grained PRs, so the next one would be to define a concrete property type like BigInt. If you think it makes more sense to have examples in the first PR, I can do that. Or, if it would help, I can get my act together and send the next PR based on this one.

spec.md
mikeurbach: No, property types won't be allowed in aggregates (they will have their own aggregates eventually). I can include some description about this in the property types section.

spec.md
seldridge: Multiple PRs are fine. 👍

mwachs5: in terms of impacting hardware, and in absence of any description of dedup, is it legal to use properties to decide whether to dedup modules or not... do we need to say anything? Or maybe we could lightly weaken this statement by saying "hardware functionality" vs "hardware"

abi.md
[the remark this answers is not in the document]
mikeurbach: I added the above, let me know what you think.

spec.md
mikeurbach: I am now using the phrasing "cannot affect hardware functionality or the hardware ABI". Still somewhat vague, since we don't define "hardware functionality", but let me know what you all think.

spec.md
mikeurbach: I mentioned that property types "cannot be used in any hardware types, including aggregates"

mikeurbach:
I filed a draft PR on my fork for what will come next: define the first concrete property type, add it to the grammar, and indicate in the Property Types section where these are legal (this list will grow as we add more property constructs). I also included a small example of what has been specced out thus far.

https://github.com/mikeurbach/firrtl-spec/pull/1

Also, if it helps, I've filed Chisel PRs for adding Property types in general: https://github.com/chipsalliance/chisel/pull/3425
and a draft PR for adding Integer type specifically: https://github.com/mikeurbach/chisel3/pull/1

mwachs5 (approved):

seldridge (approved):
LGTM

spec.md
```diff
@@ -936,6 +937,22 @@ module Top:
   node consumer_debug = read(c.out.cref); ; Consumer-side signal
 ```
 
+## Property Types
+
+FIRRTL property types represent information about the circuit that is not
+hardware. This is useful to capture domain specific knowledge and design intent
```
dtzSiFive, line 943:
```suggestion
hardware. This is useful to capture domain-specific knowledge and design intent
```

spec.md
```diff
@@ -936,6 +937,22 @@ module Top:
...
+alongside the hardware description within the same FIRRTL.
+
+Property types cannot affect hardware functionality or the hardware ABI. They
+cannot be used in any hardware types, including aggregates. They only exist to
+augment the hardware description with extra information.
+
+Handling of property types is completely implementation defined. A valid FIRRTL
```
dtzSiFive, line 950:
```suggestion
Handling of property types is completely implementation-defined. A valid FIRRTL
```

spec.md
```diff
@@ -936,6 +937,22 @@ module Top:
...
+compiler implementation may do anything with property types as long as the
+existence of property types does not affect hardware functionality or the
```
dtzSiFive, line 952: Looks like this was intentionally phrased to allow properties to change the hardware emitted, just not what's observable externally re:ABI and functionality? Sounds good/right to me.

spec.md
```diff
@@ -936,6 +937,22 @@ module Top:
...
 ```
 
+## Property Types
+
+FIRRTL property types represent information about the circuit that is not
+hardware. This is useful to capture domain specific knowledge and design intent
+alongside the hardware description within the same FIRRTL.
+
+Property types cannot affect hardware functionality or the hardware ABI. They
+cannot be used in any hardware types, including aggregates. They only exist to
```
dtzSiFive, line 947: Consider:
```suggestion
cannot be used in any hardware types, including aggregates and references. They only exist to
```

AFAIK this is 100% the case, but how/where to convey that in the spec has some flexibility :+1: .

dtzSiFive (approved):
LGTM, some nits/suggestions.

It's a little weird for the spec to mention a new "kind" of type without any use of it -- I'd consider this to be a blocker for cutting a release until something concrete is added -- but I do like idea of adding new types piecemeal for a for reasons..  but let's just not leave that dangling :+1: (don't think there's reason to worry there).

darthscsi (approved):

spec.md
mikeurbach: I think `and references` here is a good place, will adopt that.

PR #122, opened 2023-07-13
[patch] Tweak mux selector width-inference behavior.

dtzSiFive:
Don't require inferring selector to 1-bit:
* Support inferring to zero-bit (will be padded)
* Don't put constraints on operands ("back-propagation").
* Align with behavior in similar constructs: when, subaccess.

As-is selector can be entirely unconstrained otherwise, but will inferred to one-bit when used in a mux.

This masks design bugs and has no known utility.

Further, by requiring inference to 1-bit this can cause hardware that otherwise would be zero-width (and often elided) to become 1-bit and kept around.

Instead treat the selector like any other operand, but still reject if selector is inferred to be larger than 1 bit.

See https://github.com/llvm/circt/issues/5444 for further discussion and examples demonstrating the issue on current and historical implementations.

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -5,6 +5,9 @@ revisionHistory:
   # additions to the specification should append entries here.
   thisVersion:
   # Information about the old versions.  This should be static.
+    spec:
+      - Change/clarify mux selector width inference to align with other operations
+        (must infer to some width by itself, pad if infers to less than 1-bit).
   oldVersions:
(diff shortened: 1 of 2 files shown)


seldridge (approved):
LGTM

Thanks for cleaning up this edge case.

darthscsi (commented):
I'm reviewing an already committed PR just so I can approve it so hard.

PR #123, opened 2023-07-19
Add Integer property type.

mikeurbach:
This adds the first concrete property type: Integer. Integer
properties are arbitrary bit width integer values. This also indicates
where concrete property types are legal, and updates the grammar. For
now, the only legal location is in ports.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -953,6 +953,20 @@ existence of property types does not affect hardware functionality or the
 hardware ABI. For example, it is valid to drop property types from the IR
 completely.
 
+Property types are legal in the following constructs:
+
+* Port declarations on modules and external modules
+
+### Integer Property Types
+
+Integer properties represent arbitrary bit width signed integer values with a
+two's complement representation.
+
+``` firrtl
+module Example:
+  input intProp : Integer ; an input port of Integer property type
+```
+
 ## Type Modifiers
 
 ### Constant Type
@@ -3669,6 +3683,7 @@ type_aggregate = "{" , field , { field } , "}"
                | type , "[" , int , "]" ;
 type_ref = ( "Probe" | "RWProbe" ) , "<", type , ">" ;
 field = [ "flip" ] , id , ":" , type ;
+type_property = "Integer" ;
 type_simple_child = type_ground | type_enum | type_aggregate | id ;
 type = ( [ "const" ] , type_simple_child ) | type_ref ;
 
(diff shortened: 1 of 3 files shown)


spec.md
```diff
@@ -953,6 +953,20 @@ existence of property types does not affect hardware functionality or the
...
 completely.
 
+Property types are legal in the following constructs:
+
+* Port declarations on modules and external modules
+
+### Integer Property Types
+
+Integer properties represent arbitrary bit width signed integer values with a
+two's complement representation.
```
seldridge, line 963: Why does the two's complement encoding matter?

seldridge (commented):
One question about encoding.

Copy-editing wise... the spec is going to get confusing with multiple simultaneous, though qualified uses of "integer". Currently there is "integer type" (`UInt`, `SInt`), "integer literals" (`UInt<1>(0hdeadbeef)`), "integer literal (parameters)" for external modules, and now "integer property types". Keeping the language straight around all this is going to be a challenge. I.e., it may be best to pedantically always refer to each of these by their full, verbose name in order to avoid confusion. Alternatively, there may be some way to restructure the language in the spec to avoid these collisions. These comments are largely out-of-scope of this PR as this does introduce differentiating language for this new flavor of integer.

spec.md
mikeurbach: I guess it doesn't matter, but it seemed worth specifying. I guess if we ever had arithmetic or boolean operations (which I'm not planning anytime soon), it would matter. I'm happy to leave it off. I looked at the Python docs about the builtin int, and it just mentions arbitrary precision at first (is precision a better term? we use bit width elsewhere in this doc). Later, the Python docs mention two's complement, but only when discussing boolean operations on ints.

mikeurbach:
> Copy-editing wise... 

Yeah good point, I will try to be precise here, and use the verbose name, and keep it this way in future additions. I guess we can revisit if there is a better organization to use to help clarify.

spec.md
seldridge: It would matter if bitwise boolean operations are defined on integer properties. I'm not sure we need that, yet. For arithmetic, I'm not sure it matters as long as integers behave like integers.

Interesting find about Python integers! It looks like that is pedantically talking about the behavior of how integers are interpreted by bitwise operators, not the underlying storage:

> Bitwise operations only make sense for integers. The result of bitwise operations is calculated as though carried out in two’s complement with an infinite number of sign bits.

My nit here is that I don't think it makes sense to be prescriptive about the _storage_ of integers in the specification when that storage is likely invisible to a user.

mikeurbach:
I updated this to use the phrase "arbitrary precision", and drop the mention of two's complement. I also was sure to say "Integer property types"... I'll make sure the various property-related additions say "xxx property type" moving forward.

mikeurbach:
Any other concerns on this one?

seldridge (approved):
LGTM

mikeurbach:
Thanks for taking another look @seldridge!

PR #124, opened 2023-07-25
Add property assignment.

mikeurbach:
Defines the propassign statement and legal property assignments.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2412,7 +2412,7 @@ order for a property assignment to be legal, the following conditions must hold:
 5. The left-hand side expression may be used as the left-hand side in at most
    one property assignment.
 
-7. The property assignment may not occur within a conditional scope.
+7. The property assignment must not occur within a conditional scope.
 
 Note that property types are not legal for any expressions with duplex flow.
 
--- a/spec.md
+++ b/spec.md
@@ -1028,6 +1028,8 @@ flipped orientations. Thus all ground types are passive types. Vector types are
 passive if their element type is passive. And bundle types are passive if no
 fields are flipped and if all field types are passive.
 
+All property types are passive.
+
 ## Type Equivalence
 
 The type equivalence relation is used to determine whether a connection between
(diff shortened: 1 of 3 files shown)


spec.md
```diff
@@ -2387,6 +2389,58 @@ module DUT :
...
+
+Instead, property typed expressions are assigned with the `propassign`{.firrtl}
+statement.
+
+Property typed expressions have the normal rules for flow (see [@sec:flows]),
+but otherwise use a stricter, simpler algorithm than `connect`{.firrtl}. In
+order for a property assignment to be legal, the following conditions must hold:
+
+1. The left-hand and right-hand side expressions must be references (see
+   [@sec:references]).
```
seldridge, line 2405: What does this mean?  It seems sufficient to say that th left and right hand side must be property types?

spec.md
```diff
@@ -2387,6 +2389,58 @@ module DUT :
...
+2. The left-hand and right-hand side expressions must be of property types.
+
+3. The types of the left-hand and right-hand side expressions must be the same.
+
+4. The flow of the left-hand side expression must be sink.
+
+5. The flow of the right-hand side expression must be source.
+
+6. The left-hand side expression must be used as the left-hand side in at most
+   one property assignment.
```
seldridge, line 2416: The language with "must" and "at most" are slightly at odds.  I would think it would either be "may be used in at most one" or "must be used in exactly one".  The former doesn't discuss initialization checking while the latter includes initialization checking.

spec.md
```diff
@@ -2387,6 +2389,58 @@ module DUT :
...
+
+7. The property assignment may not occur within a conditional scope.
+
+Note that property types are not legal for any expressions with duplex flow.
+
+There is no notion of a connection algorithm (see
+[@sec:the-connection-algorithm]) for sub-elements of property typed expressions.
+Unlike vectors and bundles, property types are always passive (see
+[@sec:passive-types]).
```
seldridge, line 2425: The connection algorithm discusses how you connect aggregates (albeit trivially, but exhaustively specified to avoid ambiguity).  I'm not sure what the second sentence means as aggregate types are orthogonal to passivity.

spec.md
```diff
@@ -2387,6 +2389,58 @@ module DUT :
...
+
+There is no notion of last connect semantics (see [@sec:last-connect-semantics])
+for property types. Property typed expressions may only be assigned once, and
+must not be assigned in conditional scopes.
+
+This is why property assignment is called "assignment". Property typed
+expressions require static single assignment.
```
seldridge, line 2432: Consider striking this as its more background/motivation/other reading.

spec.md
mwachs5: i think he is saying you can't assign from a literal?

mikeurbach: I was thinking I would explicitly say the expressions where property types are allowed, but I think Schuyler is right that this is just redundant. This is already defined in the section for property types.

spec.md
mikeurbach: Yeah, I think I would say "may be used in at most one". I am not exactly sure yet if the spec should get into initialization checking for properties or leave it open. I am leaning toward the weaker guarantee for now, and I think "may" would be the right choice if that makes sense.

spec.md
mikeurbach: Let me re-word this. I really just want to say property types are passive and must be connected as a whole.

spec.md
```diff
@@ -3776,6 +3830,7 @@ statement =
   | force_release , [ info ]
   | "connect" , reference , "," , expr , [ info ]
   | "invalidate" , reference , [ info ] ;
+  | "propassign" , reference , "," , reference , [ info ]
```
dtzSiFive, line 3833: Should these be static references? (and if so probably should have some language about this)

mikeurbach: Yeah I think static reference is appropriate for the lhs. I am realizing that we will want to allow property literals in the rhs in the future, but those haven't been specified yet.

Where do you think it would make sense to have some language about this? I actually just removed a part of the new section that was trying to talk about the types of expressions allowed in the propassign statement, because we already document the places property types can be used in a separate section: https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#property-types

spec.md
mikeurbach: This now says "may be used in at most one". I also updated the item below from "may not" to "must not".

spec.md
mikeurbach: I changed this to "Unlike vectors and bundles, property types must be connected as a whole." I'm not sure if I even want to say this here, since we aren't even talking about property aggregates yet. Maybe I can just remove this section?

mwachs5: i think its confusing since if we can't assign to a literal (you haven't created that yet) and you haven't defined aggregates, its unclear what you CAN assign to/from other than a pass-through port, so it does seem a bit early for this statement

mikeurbach: Yeah, let me just cut this. I was thinking it would be helpful to constrast to how connects work, but really i think we should just precisely define how propassigns work (which is quite limited for now, as you point out).

mikeurbach: I think I'm going to cut "There is not notion of last connect semantics" as well--it's redundant with the list above, and in the spirit of just defining propassign without comparing it to connect, i think this makes it more simple without losing information.

rwy7 (approved):

seldridge (approved):
LGTM

PR #125, opened 2023-07-28
Add Integer property literals.

mikeurbach:
This defines literal property type expressions in general, and Integer property type literals specifically. Examples are added to show example Integer literals, as well as using literals in propassign statements.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2426,12 +2426,22 @@ module Example:
   propassign propOut, propIn
 ```
 
+The following example demonstrates a property assignment from a property literal
+expression to a module's output property type port.
+
+``` firrtl
+module Example:
+  output propOut : Integer
+  propassign propOut, Integer(42)
+```
+
 # Expressions
 
-FIRRTL expressions are used for creating constant integers, for referring to a
-declared circuit component, for statically and dynamically accessing a nested
-element within a component, for creating multiplexers, for performing primitive
-operations, and for reading a remote reference to a probe.
+FIRRTL expressions are used for creating constant integers, for creating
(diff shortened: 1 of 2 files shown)


seldridge (approved):
I'm inclined to say that we should allow this for any of the radix-specified integer literals, too. E.g., `Integer(0hdeadbeef)`. However, we can add this later. WDYT?

spec.md
```diff
@@ -2485,6 +2495,23 @@ SInt("h-2A")
...
+
+### Integer Property Literal Expresssions
+
+A literal `Integer` property type expression can be created from an integer
+literal. The following examples show literal `Integer` property type
+expressions.
+
+``` firrtl
+Integer(42)
+Integer(-42)
```
mwachs5, line 2512: can you write in hex or any other formats?

seldridge:
Semi-related, I pushed a spec patch change underneath this. The examples just above this section in the spec were still using string-encoded integer literals. I'm not sure how that slipped through. I thought I fully removed those in 3.0.0. 🤦

mikeurbach:
We could support radix-specified integer literals. I know there was a reason to support both integer literals and radix-specified literals for UInt and SInt, but I don't think there is any need to support multiple ways of doing it for properties. So I guess I'd prefer to just start with one format. Would radix-specified be better? I picked base-10 integer literals because the initial Chisel and CIRCT support were using base-10 integers, but we can change it.

seldridge:
Keep it with normal base-10, non-radix specified like you have it, i.e., `42` and `-42`. 👍

PR #126, opened 2023-08-25
[nfc] Tweak syntax highlighting rules.

dtzSiFive:
Allow nesting "outertype", and treat literals
more like normal functions.

Changes `UInt<1>(0)` but also fixes various
places where the type context persisted
after ending, such as when used inside
expressions like `when eq(n, UInt(0)) :`.

Fixes rendering of `RWProbe<T, A.B>`.

Cleanup but also preparation for adding new
types and literal expressions where this is needed.

### DIFF
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -118,7 +118,7 @@
         <DetectChar char="`" context="field" attribute="ID"/>
         <Detect2Chars char="=" char1="&gt;" attribute="Operator" context="#stay"/>
         <AnyChar String=":=.,()" attribute="Separator" context="#stay"/>
-        <DetectChar char="[" attribute="Operator" context="widthOrDepthOrLit"/>
+        <DetectChar char="[" attribute="Operator" context="widthOrDepth"/>
         <DetectChar char="&quot;" attribute="String" context="string"/>
         <StringDetect String="regreset" attribute="Keyword" context="register"/>
         <StringDetect String="reg" attribute="Keyword" context="register"/>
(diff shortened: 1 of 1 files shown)


include/firrtl.xml
```diff
@@ -152,26 +152,26 @@
...
+        <keyword String="outertype" attribute="Keyword" context="outertype"/>
         <keyword String="types" attribute="Keyword" context="type"/>
         <Detect2Chars char="{" char1="|" context="field" attribute="Separator"/>
-        <Detect2Chars char="|" char1="}" attribute="Separator" context="#pop#pop"/>
         <DetectChar char="{" context="field" attribute="Separator"/>
-        <DetectChar char="}" attribute="Separator" context="#pop#pop"/>
         <DetectChar char= ";" context="comment" attribute="Comment"/>
-        <AnyChar String="&gt;," attribute="Operator" context="#pop"/>
-        <DetectChar char="[" attribute="Operator" context="widthOrDepthOrLit"/>
+        <DetectChar char="&gt;" attribute="Operator" context="#pop"/>
```
dtzSiFive, line 160: `outertype` isn't done until the closing `>` is found.

Everything else is stay or goes into another context (which will be popped when done).

include/firrtl.xml
```diff
@@ -152,26 +152,26 @@
...
+        <DetectChar char="," attribute="Separator" context="#stay"/>
       </context>
       <context name="type" attribute="String" lineEndContext="#pop">
-        <AnyChar String="&lt;[(" attribute="Operator" context="widthOrDepthOrLit"/>
+        <AnyChar String="&lt;[" attribute="Operator" context="widthOrDepth"/>
+        <DetectChar char="(" attribute="Separator" context="#pop"/>
         <AnyChar String="," attribute="Separator" context="#pop"/>
-        <AnyChar String="}" attribute="Separator" context="#pop#pop"/>
-        <Detect2Chars char="|" char1="}" attribute="Separator" context="#pop#pop"/>
+        <AnyChar String="}" attribute="Separator" context="#pop" lookAhead="true"/>
```
dtzSiFive, line 168: If in the `type` context, it's because we saw, e.g., `UInt`.  It optionally followed by width/depth.  Any paren indicates the type portion is over (don't treat `UInt<5>(..` as in the "type" context).

Use lookAhead to exit the type context without consuming the character, for when a bundle or enum or outertype contains a type.

include/firrtl.xml
```diff
@@ -152,26 +152,26 @@
...
+        <Detect2Chars char="|" char1="}" attribute="Separator" context="#pop" lookAhead="true"/>
         <DetectChar char= ";" context="comment" attribute="Comment"/>
-        <AnyChar String="&gt;" attribute="Operator" context="#pop"/>
+        <AnyChar String="&gt;" attribute="Operator" context="#pop" lookAhead="true"/>
       </context>
-      <context name="widthOrDepthOrLit" attribute="String" lineEndContext="#stay">
-        <DetectChar char="&lt;" attribute="Operator" context="widthOrDepthOrLit"/>
-        <AnyChar String="&gt;])" attribute="Operator" context="#pop"/>
+      <context name="widthOrDepth" attribute="String" lineEndContext="#stay">
```
dtzSiFive, line 173: Looking to support "literals" that aren't appropriate to be rendered as strings and may be expressions themselves (Property types).  Tweak this to only handle width/depth and treat `T(asdf)` as normal expressions.

Fixes various in-spec rendering minor bugs where closing punctuation was sometimes considered as part of the type.

dtzSiFive (commented):
Some notes/thoughts FWIW.

PR #127, opened 2023-08-25
[minor] Add List property types and literal.

dtzSiFive:


spec.md
```diff
@@ -2650,6 +2661,17 @@ Integer(42)
...
+
+A literal `List`{.firrtl} property type expression can be created from a
+sequence of property values of the element type.  The following examples show
+literal `List`{.firrtl} property type expressions.
+
+```firrtl
+List<Integer>(Integer(42), Integer(0))
+List<List<Integer>>(List<Integer>(), List<Integer>(Integer(5)))
+```
+
```
jackkoenig, line 2674: We should probably use the term "value" instead of "literal" since you can have a List of Integer properties that are not literals, eg.
```
input foo : Integer
output bar : List<Integer>
propassign bar, List<Integer>(foo, Integer(42))
```

dtzSiFive: Haha, okay so I'm happy either way!  Had same doubt and was gently encouraged to keep with literal terminology.

Whatever answer we like best lets be consistent throughout the stack :+1: .

If not a literal will need to tweak the structure a bit but that seems fine and we'll surely want other sorts of expressions at some point.  Will circle back, regardless thanks for the attention to detail and bringing this up!

spec.md
```diff
@@ -1104,6 +1104,17 @@ module Example:
...
 
+### List Property TYpes
+
+List property types represent zero or more property values of the same type.
+List property types are parameterized over the element type.
+
+```firrtl
+module Example:
+  input list : List<Integer> ; an input port of List<Integer> property type
+  input listOfList : List<List<Integer>> ; an input port of List<Integer> property type
```
tymcauley, line 1115: Not sure exactly how you want to phrase it, but I think this was a copy-paste from the line above:

```suggestion
  input listOfList : List<List<Integer>> ; an input port of List<List<Integer>> property type
```

spec.md
dtzSiFive: Totally, that's the right phrasing.  Mostly blocked on either reworking this section to not be called "Property Literal Expressions" or split it out, words are hard.   Anyway, just wanted to note lack of progress is not lingering disagreement :+1:, just need some free writing cycles.

PR #128, opened 2023-09-19
Fix "Type Alias" section location. 

mbty:
This PR fixes the location of the "Type Alias" alias, which was inserted in the middle of the "Reference Types" section. Consequently, some subsections of "Reference Types" are wrongly attached to the "Type Alias" section.

This dates back to b037c7662240450077b0c09e82cd377076412733 and it can be observed in the 3.1.0 version of the specification (see for instance subsection 7.4.0.1, "Width and Reset Inference").

Additionally, this PR enforces the 80 characters maximum width in some places.

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -6,6 +6,8 @@ revisionHistory:
   thisVersion:
     spec:
       - Add optional groups.
+      - Fix position of "Type Alias" (it used to be in the middle of
+        "Reference Types").
     abi:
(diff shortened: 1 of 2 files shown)


seldridge (commented):
Whoops. Nice find.

Can you back out the text flow changes?

mbty:
Sure, here you go.

seldridge (approved):

PR #20, opened 2022-06-08
Adding `public` modules

debs-sifive:
Proposal for `public` modules.

Related: https://github.com/chipsalliance/firrtl-spec/pull/18

PR #130, opened 2023-09-21
[major] Add public modules

### LINKED ISSUE #82: Define public, private ("visibility")
As encountered in #75 and #76 , there's a notion of some modules as "public" and "private" being teased out, but this should be explicitly defined in the FIRRTL spec somewhere if we want to rely on it -- probably in its own PR if not as part of one of those.

seldridge:
Add a notion of "public" modules to FIRRTL.  A public module is a module that has a defined Verilog ABI.  A user can look at a FIRRTL circuit and expect to see Verilog modules with expected ports and names after compilation.  The notion of a circuit is then expanded to allow for any number of public modules.  The instance graphs of public modules may or may not share common children. In effect, a circuit is now collection of multi-rooted instantiation trees.

Any public module that is instantiated by other modules has certain restrictions on its lowering.  It follows one of the port lowering ABIs and no information from its instantiating context(s) may be used to optimize into it.  (You cannot constant prop into a public module.)  However, it is allowable for modules which instantiate the public module to optimize through the public module.

Modify the ABI to require that all public modules produce a filelist whose name is derived from the circuit and public module.

Closes #82.
Closes #20.

### DIFF
--- a/abi.md
+++ b/abi.md
@@ -36,30 +36,68 @@ may evolve as new FIRRTL constructs are added.
 
 ## On Modules
 
-### The Circuit and Top Level
+### The Circuit
 
-The top level module, as specified by the circuit name, shall be present as a
-System Verilog module of the same name.  This FIRRTL module is considered a
-"public" module and subject to the lowering constraints for public modules.
+The circuit is a container of public modules.  The circuit, by itself, does not
+have any ABI.
 
 ### External Modules
 
 An external module may be presumed to exist following the lowering constraints
-for public modules.  The module shall exist with a verilog name matching the
-defname value or, lacking that, the module name.
+for public modules.  The module shall exist with a Verilog name matching the
+defname value or, lacking that, the FIRRTL module name.  The ports of an
+external module shall adhere to one of the port lowering ABIs.
 
-###  Public Modules
+No assumption is made of the filename of an implementation of an external
+module.  It is the user's job to include files in such a way in their tools to
+resolve the name.
 
-Any module considered a "public" module shall be implemented in Verilog in a
-consistent way.  Any public module shall exist as a Verilog module of the same
-name.
+### Public Modules
 
-Each public module with definitions (e.g. not external modules) shall be placed
-in a file with the same name.  No assumption is made of the filename of an
-implementation of an external module, it is the user's job to include files in
-such a way in their tools to resolve the name.
+All public modules shall be implemented in Verilog in a consistent way.  All
+public modules shall exist as Verilog modules of the same name.  Each public
+module shall be placed in a file with a format as follows where `module` is the
+name of the public module:
 
-### Port Lowering ABIv1
+``` ebnf
+filename = module , ".sv" ;
+```
+
+Each public module in a circuit shall produce a filelist that contains the
+filename of the file containing the public module and any necessary files that
+define all public or private module files instantiated under it.  Files that
+define external modules are not included.  This filelist shall have a name as
+follows where `module` is the name of the public module:
+
+``` ebnf
+filelist_filename = "filelist_" , module , ".f" ;
+```
+
+The filelist contents are a newline-delimited list of filenames.
+
+The ports of public modules shall be lowered using one of the Port Lowering ABIs.
+
+No information about the instantiations of a public module may be used to
+compile a public module---a public module is compiled as if it is never
+instantiated.  However, it is legal to compile modules which instantiate public
+modules with full knowledge of the public module internals.
+
+### Private Modules
+
+Private modules have no defined ABI.  If the compilation of private modules
+produces files, then those files shall be included when necessary in public
+module filelists.
+
+FIRRTL compilers shall mangle the names of private modules that are not removed
+by compilation.  The mangling scheme is implementation defined.  This is done to
+avoid name collisions with the private modules produced by other compilations.
+
+### Port Lowering ABIs
+
+There are two supported port lowering ABIs.  These ABIs are applicable to public
+modules or external modules only.
+
+#### Port Lowering ABIv1
 
 Ports are generally lowered to netlist types, except where Verilog's type system
 prevents it.
@@ -68,8 +106,8 @@ Ports of integer types shall be lowered to netlist ports (`wire`{.verilog}) as a
 packed vector of equivalent size.  For example, consider the following FIRRTL:
 
 ```FIRRTL
-circuit Top :
-  module Top :
+circuit :
+  public module Top :
     output out: UInt<16>
     input b: UInt<32>
 ```
@@ -86,16 +124,16 @@ module Top(
 Ports of aggregate type shall be scalarized according to the "Aggregate Type
 Lowering" description in the FIRRTL spec.
 
-Ports of ref type shall be lowered to a Verilog macro of the form `` `define
-ref_<circuit name>_<module name>_<portname> <internal path from module>`` in a
-file with name `ref_<circuit name>_<module name>.sv`.  References to aggregates
-will be lowered to a series of references to ground types.  This ABI does not
-specify whether the original aggregate referent is scalarized or not.
+Ports of ref type on public modules shall, for each public module, be lowered to
+a Verilog macro of the form `` `define ref_<module name>_<portname> <internal
+path from module>`` in a file with name `ref_<module name>.sv`.  References to
+aggregates will be lowered to a series of references to ground types.  This ABI
+does not specify whether the original aggregate referent is scalarized or not.
 
 All other port types shall lower according ot the type lowering in
 section ["On Types"](#On-Types).
 
-### Port Lowering ABIv2
+#### Port Lowering ABIv2
 
 Ports are lowered per the v1 ABI above, except for aggregate types.
 
@@ -137,30 +175,37 @@ used by a nested group will create output ports because SystemVerilog disallows
 the use of a bind instantiation underneath the scope of another bind
 instantiation.  The names of additional ports are implementation defined.
 
-For each optional group, one binding file will be produced for each group or
-nested group.  The circuit, group name, and any nested groups are used to derive
-a predictable filename of the binding file.  The file format uses the format
-below:
+For each optional group and public module, one binding file will be produced for
+each group or nested group.  The public module, group name, and any nested
+groups are used to derive a predictable filename of the binding file.  The file
+format uses the format below where `module` is the name of the public module,
+`root` is the name of the root-level group and `nested` is the name of zero or
+more nested groups:
 
 ``` ebnf
-filename = "groups_" , circuit , "_", root , { "_" , nested } , ".sv" ;
+filename = "groups_" , module , "_", root , { "_" , nested } , ".sv" ;
 ```
 
 As an example, consider the following circuit with three optional groups:
 
 ``` firrtl
-circuit Foo:
+circuit:
   declgroup Group1, bind:
     declgroup Group2, bind:
       declgroup Group3, bind:
+  public module Bar:
+  public module Baz:
 ```
 
-When compiled to Verilog, this will produce three bind files:
+When compiled to Verilog, this will produce six bind files:
 
 ```
-groups_Foo_Group1.sv
-groups_Foo_Group1_Group2.sv
-groups_Foo_Group1_Group2_Group3_.sv
+groups_Bar_Group1.sv
+groups_Bar_Group1_Group2.sv
+groups_Bar_Group1_Group2_Group3_.sv
+groups_Baz_Group1.sv
+groups_Baz_Group1_Group2.sv
+groups_Baz_Group1_Group2_Group3_.sv
 ```
 
 The contents of each binding files must have the effect of including all code
@@ -178,7 +223,7 @@ module `Bar`.  Both `Foo` and `Bar` contain groups.  To make the example
 simpler, no constant propagation is done:
 
 ``` firrtl
-circuit Foo:
+circuit:
   declgroup Group1, bind:
     declgroup Group2, bind:
 
@@ -197,7 +242,7 @@ circuit Foo:
         node notNotA = not(notA)
         define _notNotA = probe(notNotA)
 
-  module Foo:
+  public module Foo:
     inst bar of Bar
 
     group Group1:
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -5,7 +5,12 @@ revisionHistory:
   # additions to the specification should append entries here.
   thisVersion:
     spec:
+      - Add public modules.  Change circuits to support any number of public
+        modules which may have disjoint instance hierarchies.
     abi:
+      - Add ABI for public modules and filelist output.
+      - Changed ABI for group and ref generated files.  These now use the public
+        module and not the circuit.
   # Information about the old versions.  This should be static.
   oldVersions:
     - version: 3.2.0
--- a/spec.md
+++ b/spec.md
@@ -137,24 +137,60 @@ versioned release of this standard to include this preamble.
 
 ``` firrtl
 FIRRTL version 1.1.0
-circuit MyTop...
+circuit :
 ```
 
 # Circuits and Modules
 
 ## Circuits
 
-All FIRRTL circuits consist of a list of modules, each representing a hardware
-block that can be instantiated. The circuit must specify the name of the
-top-level module.
+A FIRRTL circuit is a collection of FIRRTL modules.  Each module is a hardware
+"unit" that has ports, registers, wires, and may instantiate other modules (see:
+[@sec:modules]).  (This is the same concept as a Verilog `module`{.verilog}.)  A
+public module may be instantiated outside the current circuit.  Public modules
+are the exported identifiers of a circuit.  Any non-public module may not be
+instantiated outside the current circuit.
+
+Consider the following circuit.  This contains two modules, `Bar` and `Baz`.
+Module `Baz` is marked public.
 
 ``` firrtl
-circuit MyTop :
-  module MyTop :
-    ; ...
-  module MyModule :
-    ; ...
+circuit :
+  module Bar :
+    input a: UInt<1>
+    output b: UInt<1>
+
+    connect b, a
+
+  public module Baz :
+    input a: UInt<1>
+    output b: UInt<1>
+
+    inst bar of Bar
+    connect bar.a, a
+    connect b, bar.b
 ```
+
+The circuit below is more complex.  This circuit contains three public modules,
+`Foo`, `Bar`, and `Baz`, and one non-public module, `Qux`.  Module `Foo` has no
+common instances with `Bar` or `Baz`.  `Bar` and `Baz` both instantiate `Qux`:
+
+``` firrtl
+circuit :
+  public module Foo :
+
+  public module Bar :
+    inst qux of Qux
+
+  public module Baz :
+    inst qux of Qux
+
+  module Qux :
+```
+
+A circuit that contains no public modules is trivially equivalent to a circuit
+that contains no modules.  It is not enforced that a circuit has at least one
+public module, but it is expected that it does.
 
 ## Modules
 
@@ -178,6 +214,38 @@ Note that a module definition does *not* indicate that the module will be
 physically present in the final circuit. Refer to the description of the
 instance statement for details on how to instantiate a module
 ([@sec:instances]).
+
+### Public Modules
+
+A public module is a module that is marked with the `public`{.firrtl} keyword.
+A public module may be instantiated outside by other circuits.
+
+The example below shows a public module `Foo`:
+
+``` firrtl
+public module Foo:
+  input a: UInt<1>
+  output b: UInt<1>
+
+  connect b, a
+```
+
+A public module has a number of restrictions:
+
+1. A public module may have no ports of uninferred width.
+1. A public module may have no ports of abstract reset type.
+1. A public module may have no ports of input probe type.
+1. A `RWProbe`{.firrtl} may not be used to access a public module's ports.
+1. A public module may be instantiated by other modules within a circuit, but
+   the behavior of the module must not be affected by these instantiations.
+
+For more information on the lowering of public modules, see the FIRRTL ABI
+Specification.
+
+### Private Modules
+
+A private module is any module which is not public.  Private modules have none
+of the restrictions of public modules.
 
 ## Optional Groups
 
@@ -219,7 +287,7 @@ contains a group definition that creates a node computed from a port defined in
 the scope of `Foo`.
 
 ``` firrtl
-circuit Foo:
+circuit:
   declgroup Bar, bind:  ; Declaration of group Bar with convention "bind"
 
   module Foo:
@@ -236,7 +304,7 @@ declarations, three of which are nested.  `Bar` is the top-level group.  `Baz`
 and `Qux` are nested under `Bar`.  `Quz` is nested under `Qux`.
 
 ``` firrtl
-circuit Foo:
+circuit:
   declgroup Bar, bind:
     declgroup Baz, bind:
     declgroup Qux, bind:
@@ -274,7 +342,7 @@ inside the group.  _Stated differently, module `Baz` has an additional port `_a`
 that is only accessible inside a defined group `Bar`_.
 
 ``` firrtl
-circuit Foo:
+circuit:
   declgroup Bar, bind:
 
   module Baz:
@@ -299,7 +367,7 @@ the nesting.  E.g., the following circuit has a port associated with the nested
 group `Bar.Baz`:
 
 ``` firrtl
-circuit Foo:
+circuit:
   declgroup Bar, bind:
     declgroup Baz, bind:
 
@@ -877,8 +945,8 @@ with inferring the input port as `AsyncReset`{.firrtl} if a direct connection
 was used:
 
 ```firrtl
-circuit ResetInferBad :
-  module ResetInferBad :
+circuit :
+  public module ResetInferBad :
     input in : Reset
     output out : AsyncReset
     connect out, read(probe(in))
@@ -888,8 +956,8 @@ The following circuit has all resets inferred to `AsyncReset`{.firrtl},
 however:
 
 ```firrtl
-circuit ResetInferGood :
-  module ResetInferGood :
+circuit :
+  public module ResetInferGood :
     input in : Reset
     output out : Reset
     output out2 : AsyncReset
@@ -961,8 +1029,8 @@ contexts so a single description of the module may be generated.
 The following example demonstrates such an invalid use of probe references:
 
 ```firrtl
-circuit Top:
-  module Top:
+circuit:
+  public module Top:
     input in : UInt<4>
     output out : UInt
 
@@ -1119,7 +1187,7 @@ with some non-constant arguments produce a non-constant.  Constants can be used
 in any context with a source flow which allows a non-constant.  Constants may be
 used as the target of a connect so long as the source of the connect is itself
 constant.  These rules ensure all constants are derived from constant integer
-expressions or from constant-typed input ports of the top-level module.
+expressions or from constant-typed input ports of a public module.
 
 ``` firrtl
 const UInt<3>
@@ -1132,7 +1200,7 @@ control flow is conditioned on an expression which has a constant type.  This
 means if a constant is being assigned to in a `when`{.firrtl} block, the
 `when`{.firrtl}'s condition must be a constant.
 
-Output ports of external modules and input ports to the top-level module may be
+Output ports of external modules and input ports to a public module may be
 constant.  In such case, the value of the port is not known, but that it is
 non-mutating at runtime is known.
 
@@ -2025,12 +2093,12 @@ example demonstrates creating an instance named `myinstance`{.firrtl} of the
 `MyModule`{.firrtl} module within the top level module `Top`{.firrtl}.
 
 ``` firrtl
-circuit Top :
+circuit :
   module MyModule :
     input a: UInt
     output b: UInt
     connect b, a
-  module Top :
+  public module Top :
     inst myinstance of MyModule
 ```
 
@@ -3416,11 +3484,12 @@ namespace.
 Annotations encode arbitrary metadata and associate it with zero or more
 targets ([@sec:targets]) in a FIRRTL circuit.
 
-Annotations are represented as a dictionary, with a "class" field which
-describes which annotation it is, and a "target" field which represents the IR
-object it is attached to. Annotations may have arbitrary additional fields
-attached. Some annotation classes extend other annotations, which effectively
-means that the subclass annotation implies to effect of the parent annotation.
+Annotations are represented as a dictionary, with a required "class" field which
+describes which annotation it is. An annotation has an optional "target" field
+which represents the IR object it is attached to. Annotations may have arbitrary
+additional fields attached. Some annotation classes extend other annotations,
+which effectively means that the subclass annotation implies to effect of the
+parent annotation.
 
 Annotations are serializable to JSON.
 
@@ -3429,7 +3498,15 @@ Below is an example annotation used to mark some module `foo`{.firrtl}:
 ```json
 {
   "class":"myannotationpackage.FooAnnotation",
-  "target":"~MyCircuit|MyModule>foo"
+  "target":"MyModule>foo"
+}
+```
+
+Below is an example of an annotation which does not have a target:
+
+```json
+{
+  "class":"myannotationpackage.BarAnnotation"
 }
 ```
 
@@ -3440,18 +3517,17 @@ example, there may be multiple instances of a module which will eventually
 become multiple physical copies of that module on the die.
 
 Targets are a mechanism to identify specific hardware in specific instances of
-modules in a FIRRTL circuit.  A target consists of a circuit, a root module, an
-optional instance hierarchy, and an optional reference. A target can only
-identify hardware with a name, e.g., a circuit, module, instance, register,
-wire, or node. References may further refer to specific fields or subindices in
+modules in a FIRRTL circuit.  A target consists of a root module, an optional
+instance hierarchy, and an optional reference. A target can only identify
+hardware with a name, e.g., a module, instance, register, wire, or
+node. References may further refer to specific fields or subindices in
 aggregates. A target with no instance hierarchy is local. A target with an
 instance hierarchy is non-local.
 
 Targets use a shorthand syntax of the form:
 
 ```ebnf
-target = “~” , circuit ,
-         [ “|” , module , { “/” (instance) “:” (module) } , [ “>” , ref ] ]
+target = module , [ { “/” (instance) “:” (module) } , [ “>” , ref ] ]
 ```
 
 A reference is a name inside a module and one or more qualifying tokens that
@@ -3469,7 +3545,7 @@ circuit. This consists of four instances of module `Baz`{.firrtl}, two instances
 of module `Bar`{.firrtl}, and one instance of module `Foo`{.firrtl}:
 
 ```firrtl
-circuit Foo:
+circuit:
   module Foo:
     inst a of Bar
     inst b of Bar
@@ -3492,14 +3568,13 @@ representation where each instance is broken out into its own module.
 Using targets (or multiple targets), any specific module, instance, or
 combination of instances can be expressed. Some examples include:
 
-Target                   Description
------------------------  -------------
-`~Foo`                   refers to the whole circuit
-`~Foo|Foo`               refers to the top module
-`~Foo|Bar`               refers to module `Bar`{.firrtl} (or both instances of module `Bar`{.firrtl})
-`~Foo|Foo/a:Bar`         refers just to one instance of module `Bar`{.firrtl}
-`~Foo|Foo/b:Bar/c:Baz`   refers to one instance of module `Baz`{.firrtl}
-`~Foo|Bar/d:Baz`         refers to two instances of module `Baz`{.firrtl}
+Target              Description
+------------------  -------------
+`Foo`               refers to module `Foo`{.firrtl} (or the only instance of module `Foo`{.firrtl})
+`Bar`               refers to module `Bar`{.firrtl} (or both instances of module `Bar`{.firrtl})
+`Foo/a:Bar`         refers just to one instance of module `Bar`{.firrtl}
+`Foo/b:Bar/c:Baz`   refers to one instance of module `Baz`{.firrtl}
+`Bar/d:Baz`         refers to two instances of module `Baz`{.firrtl}
 
 If a target does not contain an instance path, it is a _local_ target.  A local
 target points to all instances of a module.  If a target contains an instance
@@ -3517,11 +3592,11 @@ containing two annotations:
 [
   {
     "class":"hello",
-    "target":"~Foo|Bar"
+    "target":"Bar"
   },
   {
     "class":"world",
-    "target":"~Foo|Baz"
+    "target":"Baz"
   }
 ]
 ```
@@ -3531,14 +3606,14 @@ Annotation JSON in `%[ ... ]`{.firrtl}.  The following shows the above
 annotation file stored in-line:
 
 ``` firrtl
-circuit Foo: %[[
+circuit: %[[
   {
     "class":"hello",
-    "target":"~Foo|Bar"
+    "target":"Bar"
   },
   {
     "class":"world",
-    "target":"~Foo|Baz"
+    "target":"Baz"
   }
 ]]
   module Foo :
@@ -3702,8 +3777,8 @@ As an example illustrating some of these points, the following is a legal FIRRTL
 circuit:
 
 ``` firrtl
-circuit Foo :
-    module Foo :
+circuit :
+    public module Foo :
       skip
     module Bar :
      input a: UInt<1>
@@ -3723,8 +3798,8 @@ and '`\`' itself.
 The following example shows the info tokens included:
 
 ``` firrtl
-circuit Top : @[myfile.txt 14:8]
-  module Top : @[myfile.txt 15:2]
+circuit : @[myfile.txt 14:8]
+  public module Top : @[myfile.txt 15:2]
     output out: UInt @[myfile.txt 16:3]
     input b: UInt<32> @[myfile.txt 17:3]
     input c: UInt<1> @[myfile.txt 18:3]
@@ -3991,7 +4066,7 @@ statement =
 
 (* Module definitions *)
 port = ( "input" | "output" ) , id , ":" , (type | type_property) , [ info ] ;
-module = "module" , id , ":" , [ info ] , newline , indent ,
+module = [ "public" ] , "module" , id , ":" , [ info ] , newline , indent ,
            { port , newline } ,
            { statement , newline } ,
          dedent ;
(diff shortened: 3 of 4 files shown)


dtzSiFive:
cc #82 .

mwachs5:
> The notion of a circuit is then expanded to allow for multiple roots and these roots may have no common children. 

What is the point of this proposal, if they can have no common children?

spec.md
```diff
@@ -144,18 +144,51 @@ circuit MyTop...
...
-  module MyModule :
-    ; ...
+circuit Foo :
+  module Bar :
+    input a: UInt<1>
+    output b: UInt<1>
+
+    connect b, a
+
+  public module Bar :
```
dtzSiFive, line 163:
```suggestion
  public module Baz :
```

spec.md
```diff
@@ -179,6 +212,32 @@ physically present in the final circuit. Refer to the description of the
...
+A public module is a module that is marked with the `public`{.firrtl} keyword.
+This means that this is a module that is usable external to this circuit.  A
+public module thereby has a defined ABI---knowing the FIRRTL representation is
+sufficient to know its compiled representation.
+
+A public module has a number of restrictions:
+
+1. A public module may not be instantiated in the current circuit.
+1. A public module may have no ports of uninferred width.
+1. A public module may have no ports of abstract reset type.
```
dtzSiFive, line 226: FWIW also no input probes, can cannot RWProbe a public module's ports.  This is in the text elsewhere already, and may not need to be repeated here.

seldridge:
It is a point of discussion if public modules should be allowed to be instantiated. This could work one of several ways if allowed:

1. A public, instantiated module is compiled as if it were an external module. I.e., all ports are marked don't touch and the same restrictions apply on uninferred widths and abstract reset.
2. A public, instantiated module is compiled with all optimizations allowed, but the ports have to follow one of the port lowering ABIs. This implies no uninferred widths and no abstract reset.
3. A public, instantiated module is compiled will all optimizations on and you get a Verilog file out, but you have no idea what the ports are until you compile it. 😅

spec.md
```diff
@@ -144,18 +144,51 @@ circuit MyTop...
 
 ## Circuits
 
-All FIRRTL circuits consist of a list of modules, each representing a hardware
-block that can be instantiated. The circuit must specify the name of the
-top-level module.
+A FIRRTL circuit is a named collection of FIRRTL modules.  Each module is a
+hardware "unit" that has ports, registers, wires, and may instantiate other
+modules (see: [@sec:modules]).  (This is the same concept as a Verilog
+`module`{.verilog}.)  A public module is usable outside of the current circuit.
```
mwachs5, line 150: what does it mean to be "usable". Should we say that a public module follows the defined verilog ABI? And that it blocks constant propagation across its boundaries? I'm still unclear if that is what "public" means.

mwachs5: i guess you clarify this below

spec.md
mwachs5, line 163: did you want one of these to be Baz?

Can you show that you can have another `public module SomethingElse` that can also instantiate the non-public module?

spec.md
```diff
@@ -179,6 +212,32 @@ physically present in the final circuit. Refer to the description of the
...
+### Public Modules
+
+A public module is a module that is marked with the `public`{.firrtl} keyword.
+This means that this is a module that is usable external to this circuit.  A
+public module thereby has a defined ABI---knowing the FIRRTL representation is
+sufficient to know its compiled representation.
+
+A public module has a number of restrictions:
+
+1. A public module may not be instantiated in the current circuit.
```
mwachs5, line 224: because of this point 1, i feel like instead of calling this "public", 'root' would be a better word. That captures more the meaning. We may decide that public is applicable to things that do not meet this requirement 1, later.

spec.md
seldridge: Talking about what it means to be "public" is difficult here as we haven't even introduced module, yet. This is more of rewriting the spec to put things in a more logical order. The requirements of `public` are specified in the "Public Modules" subsection of the "Module" section which follows the "Circuit" section. @mmaloney-sf has brought up the fact that this whole thing is hard to read and it may make sense to reverse all these sections. I.e., introduce `module` before `circuit`. This type of thing is largely out of scope of this PR. I agree we should do some big rewrites, though.

spec.md
mwachs5, line 224: would it be acceptable to instantiate a public aka root module *as an ext module (would need a new API)* in the current circuit? If not, why not?

abi.md
```diff
@@ -50,15 +50,23 @@ defname value or, lacking that, the module name.
...
 implementation of an external module, it is the user's job to include files in
 such a way in their tools to resolve the name.
 
+Each public module in a circuit will produce a filelist that contains the name
+of the public module and all modules instantiated under it.  This file will have
+a format as follows where `circuit` is the name of the FIRRTL circuit and
+`module` is the name of the public module:
+
+``` ebnf
+filename = "filelist_" , circuit , "_", module , ".f" ;
```
dtzSiFive, line 67:
```suggestion
filename = "filelist_" , circuit , "_" , module , ".f" ;
```

spec.md
```diff
@@ -144,18 +144,51 @@ circuit MyTop...
...
+  public module Bar :
+    input a: UInt<1>
+    output b: UInt<1>
+
+    inst bar of Bar
+    connect bar.a, a
+    connect b, bar.b
+```
+
+A circuit may contain public modules that instantiate no common instances.
```
dtzSiFive, line 172: Saying same thing you said a few different ways but just to be clear:

Dropping "main module".  Circuits are kinda like packages of public entirely disjoint circuits.
Is that right?

1) Can they instantiate the same extmodule's? Or is the rule about not sharing any instances only for definitions within the circuit?
2) Can they share anything at all, like the recent top-level definitions (class/extclass, properties, type alias, so on)?
3) Other than the filelist detail, I think what's proposed here is very much like what you'd get with different circuits (each with own "top") but instead in same file-- with difference being moving multiple circuits into same circuit requires them to be disjoint (where-as today multiple circuits can of course extmodule and use the other).  IS that about right?

spec.md
```diff
@@ -144,18 +144,51 @@ circuit MyTop...
...
+
+In the circuit `Qux` below, there are two public modules, `Corge` and `Grault`.
+Neither instantiates each other:
```
dtzSiFive, line 175:
```suggestion
Neither instantiates the other:
```
? I don't know but "neither.. each other" seems strange.   It's likely I'm wrong, whatever you think is best :+1:.

dtzSiFive (commented):
Cool! Added some comments for discussion.  Thanks!

spec.md
dtzSiFive: :exploding_head: good question / fun idea!

dtzSiFive:
> It is a point of discussion if public modules should be allowed to be instantiated. This could work one of several ways if allowed:
> 
>     1. A public, instantiated module is compiled as if it were an external module. I.e., all ports are marked don't touch and the same restrictions apply on uninferred widths and abstract reset.
> 
>     2. A public, instantiated module is compiled with all optimizations allowed, but the ports have to follow one of the port lowering ABIs. This implies no uninferred widths and no abstract reset.
> 
>     3. A public, instantiated module is compiled will all optimizations on and you get a Verilog file out, but you have no idea what the ports are until you compile it. 😅

IMO public modules should have their ports fixed -- no new ports, none can be changed, renamed, or removed.  And scalarized convention for use by an extmodule somewhere, unless (somehow) indicated explicitly differently.

I'd suggest not talking about whether optimizations are "allowed" -- they're always allowed the question is what are the semantics of the language under which they operate (as they always must)?

IMO public should mean that we cannot do the sort of optimization that would allow us to:
1) Delete ports -- because we must assume they may have a live user elsewhere
2) Constant-prop /into/ the circuit (optimization proof obligation here being "port is driving to known value in all instantiation sites" -- we must assume such exist we cannot inspect and so conservatively may not optimize a public thing w.r.t its use within the design)

This is not the same as dontTouch -- which I'd summarize in practice as the strongest blocker (and IIRC "preserve name" was part of it?) similar to making the signal /itself/ public s.t it may be remotely written from anywhere at any time.  Public visibility on a module (IMO, we're inventing it) shouldn't necessarily imply that.

For example, a public module driving constants to its ports must preserve all ports (for ABI and for reasons above) but any modules compiled along with that definition can optimize themselves with analysis results over the public module -- that is, if it drives `5` to port `x` then at instantiation sites it's an analysis fact usable for transformation that the port will always have value `5`.  That is, you can analyze "upwards" through "public" but not "downwards" (and transform for purposes of optimization using such analysis results).

In LLVM (IIRC) there's actually a thing called `available_externally` (https://llvm.org/docs/LangRef.html#linkage-types) which can be used to give a definition for an entity (here, module) such that it could be inlined or reasoned about where it's used BUT is not the actual definition (not suggesting we have that here, but same sort of idea except this IS the definition of the thing as well).

Does that make sense / WDYT?

----

Reading your comment again, I think I missed your focus above a little, but think the above is worth adding to the discussion.

In the above framing, I'd say a compiler is free to (basically) inline a public module into its instantiation sites (if we did compiler-driven inlining vs annotation-driven today, but point is it /could/ -- or if we had control of naming / private visibility we could clone a public module and specialize it for our use of it, only obligated to keep it and its tree alone).

Public modules may not be deduplicated, and depending on where things land re:"can share any modules" may mean no dedup across public instantiation trees (as that would cause sharing).

Public modules marked inline maybe is okay (above) but unlike today where it also implies "I do not expect to see this in my output" the public module must always be present in the output, because, that's its whole thing.

dtzSiFive:
May be worth picking the syntax change from https://github.com/chipsalliance/firrtl-spec/pull/20/files#diff-fc53c821ad3e7a3caa0a87f0f46f48202fca3bc55dbb9e56fc34cb0c6ec3ead3R21 but haven't looked at impact with/without that locally yet.

spec.md
mwachs5: i think you read the eats shoots and leaves like i did ... the statement is that they are *allowed* to instantiate the same (non-public) modules and extmodules. They do not HAVE to share any commmon modules and extmodules.

In other words, (https://mentor.ieee.org/myproject/Public/mytools/draft/styleman.pdf)

- public modules `MAY` share common modules and extmodules.
- public modules `SHALL NOT` be instantiated by any other modules.

mwachs5:
Regarding @dtzSiFive comments:

> Public modules may not be deduplicated, and depending on where things land re:"can share any modules" may mean no dedup across public instantiation trees (as that would cause sharing).

I agree that this should be the policy, i dont think that dedup is mentioned anywhere in the spec as even a possibility, but since a public module has to be in a module with the same name as the firrtl module that seems like a given (because otherwise its name would change)?

> Public modules marked inline maybe is okay (above) but unlike today where it also implies "I do not expect to see this in my output" the public module must always be present in the output, because, that's its whole thin

How can you mark a `module` inline, vs an instance? In this current doc, since public modules can't be instantiated by anything, the concept of inlining is not applicable

dtzSiFive:
> I agree that this should be the policy, i dont think that dedup is mentioned anywhere in the spec as even a possibility, but since a public module has to be in a module with the same name as the firrtl module that seems like a given (because otherwise its name would change)?

(Yeah, apologies, dedup is not something in spec and arguably shouldn't be-- just using as example to tease out the sort of thing I'm thinking public should imply, which sounds we're all in agreement on! And yes given what appears to be our agreed notions here, it should be "a given" for sure.  Just making sure we're looking at things same way!)

> How can you mark a `module` inline, vs an instance? In this current doc, since public modules can't be instantiated by anything, the concept of inlining is not applicable (and again the concept of inlining doesn't appear in this doc so far)

This is interesting.  Well, in CIRCT we move all annotations on instances to modules: https://github.com/llvm/circt/blob/6b1f86c3609ffead02bd820148c593da7c876063/lib/Dialect/FIRRTL/FIRRTLAnnotationHelper.cpp#L236 .
And FWIW our inliner does not inline in this instance-path-sensitive sort of way, see https://github.com/llvm/circt/issues/2790 .  So, perhaps as an implementation detail or maybe a bug, in practice inline might as well be an annotation on the module.
And again, you're right re:"since public modules can't be instantiated by anything, the concept of inlining is not applicable" (and as you say isn't relevant to spec).  I was using behavior we know and rely on to describe what "public" would mean-- as you say if we can't instantiate them locally then there's marking it inline --instance or module should be at best a no-op if not illegal.

Glad we're in agreement.

mwachs5:
@dtzSiFive yes you are i think asking a lot of good questions that apply if we allow "modules that are deep within someone else's hierarchy" to be public, but i think this proposal as written is dodging all of those

spec.md
seldridge: We can just remove this requirement. I think Will, you, and others have all convinced me that this isn't useful to enforce. I'm planning to back this out and then add language around the types of optimizations that are allowed here. Specifically, a public module that is instantiated must be compiled as if it were not instantiated. However, a module that instantiates a public module is allowed to optimize through a public module.

E.g., it is legal for a parent to inline a public module that it instantiates. However, that public module must still be produced following the Verilog ABI.

Initial circuit:
```
circuit Foo:
  public module Foo:
    inst bar of Bar

  public module Bar:
    wire w: UInt<1>
```

Final legal circuit:

```
circuit Foo:
  public module Foo:
    wire bar_w: UInt<1>
  public module Bar:
    wire w: UInt<1>
```

E.g., if a parent instantiates a public module and drives a port to a constant that is fed out through an output port, then the parent module can constant prop through the instance. However, the constant cannot be moved into the instance.

Initial circuit:

```
circuit Foo:
  public module Foo:
    output b: UInt<1>
    inst bar of bar
    connect bar.a, UInt<1>(1)
    connect b, bar.b
  public module Bar:
    input a: UInt<1>
    output b: UInt<1>
    connect b, a
```

Optimized circuit:
```
  public module Foo:
    output b: UInt<1>
    connect b, UInt<1>(1)
  public module Bar:
    input a: UInt<1>
    output b: UInt<1>
    connect b, a
```

Illegally optimized circuit:
```
  public module Foo:
    output b: UInt<1>
    connect b, UInt<1>(1)
  public module Bar:
```

spec.md
dtzSiFive: Oh my goodness, totally! Hahaha, thanks!

D'oh!

Thanks for the `MAY` / `SHALL NOT` language :+1: . Okay, great.  That makes a lot of sense.

seldridge: There's no requirement on disjointness. This statement was more to make it clear that the public modules _may_ have entirely disjoint instance graphs. However, they may also share anything. Put differently, circuits are collections of `public` modules. What those `public` modules instantiate or reference, has no real bearing on anybody using them as the rest of this isn't visible via the ABI.

spec.md
seldridge: Made this example longer and this language is no longer present. Good catch.

spec.md
```diff
@@ -179,6 +218,34 @@ physically present in the final circuit. Refer to the description of the
 instance statement for details on how to instantiate a module
 ([@sec:instances]).
 
+### Public Modules
+
+A public module is a module that is marked with the `public`{.firrtl} keyword.
+This means that this is a module that is usable external to this circuit.  A
+public module thereby has a defined ABI---knowing the FIRRTL representation is
+sufficient to know its compiled representation.
```
mwachs5, line 226:
```suggestion
A public module is a module that is marked with the `public`{.firrtl} keyword.
A public module has a defined ABI---knowing the FIRRTL representation is
sufficient to know its compiled representation.
This means that this is a module that is usable external to this circuit.  
```

Suggest switching the order of these things -- it's usable *because* it has a defined ABI.

spec.md
mwachs5: Ah, if you don't enforce it then all @dtzSiFive questions I said didn't matter start mattering again ;) But i think that the rule that a public module has to exist in a verilog file of that name squashes many of the questions.

mwachs5: > Specifically, a public module that is instantiated must be compiled as if it were not instantiated. However, a module that instantiates a public module is allowed to optimize through a public module.

Putting this in my own words:
- Optimizations on a module A that instantiates a public module B MAY treat B as if it were equivalent to an extmodule, but optimizations to A MAY also fully consider the contents of B and perform any modifications to A desired (including not instantiating B at all).

- When a module A instantiates a public module B, optimizations on B must ignore any information known about its context in A

mwachs5: i am a little bit worried about the first bullet, sicne it might imply that you are actually essentially duplicating the public module into a "not public" copy and using that, which might not get the desired effect.

seldridge:
This has been updated to allow for public modules to be instantiated.

spec.md
```diff
@@ -179,6 +218,43 @@ physically present in the final circuit. Refer to the description of the
...
+
+  connect b, a
+```
+
+A public module has a number of restrictions:
+
+1. A public module may have no ports of uninferred width.
+1. A public module may have no ports of abstract reset type.
+1. A public module may have no ports of input probe type.
+1. A `RWProbe`{.firrtl} may not be used to access a public module's ports.
```
mwachs5, line 243: at some point we're going to need to relax this RWProbe thing...

dtzSiFive: In short, this is a rule because of the net propagation problems that violate basic FIRRTL semantics and cause problems in practice that can be a pain to track down (and are not consistently simulated).

So IMO it's not about relaxing it's about not breaking things that need to work.  What motivates this comment / what's the ask?

Since you brought it up :wink:.... none of this is new but perhaps needs reviving:

As long as we don't emit ports as variables (logic), and as nets (as we do now), I think we want that restriction.  I'm not sure how reasonable or not it'd be to move forward with such a change.  Do you have a sense of that?
I think this was discussed long ago and maybe even attempted but I don't know the details.

One idea: require as part of the extmodule/module signature, and ABI, that it may be forced and undefined behavior to have it wrong re:forceability on either side of the boundary.  This could in practice be "emit as variable", but selectively not everywhere (in trade for requiring correct declarations or bugs), basically encode the forceability on the type itself (not the rwprobe, the target port).  This doesn't really feel like a win, but maybe others have other opinions.

Or: we let the net propagation problems be users problem/implementation (/simulator-)defined behavior, hopefully giving them a way to mitigate that isn't manually inserting buffer wires and marking don't touch in order to use this feature or in defense of possibility it's used in a module they don't own (this happens in practice in designs "in the wild" due to adhoc use of force XMR's).  This has benefit of not getting in the way of users who don't want this restriction, but is my least favorite as I'd rather make RWProbe a useful robust concept that acts as folks might expect it to.

WDYT? Is there a way to get what drives this ask a different way?

Basically it's for those reasons it's a requirement, no disagreement that it'd be better to not have such a rule if there's a better place we can take things :+1: .

dtzSiFive: Anyway this probably should be discussed elsewhere, but don't mean to close this down -- please LMK here or elsewhere if have any further thoughts/ideas/suggestions.

mwachs5: Ah, I misread this as "there may not be any RWProbe type ports". Which is not in fact what this says, and now I understand your objection.

spec.md
mwachs5: in the future we could add a thing at the instantiation site that is `inst foo of Foo as public` which means "use the public version of this and I mean it" to block the optimizations the other way.

spec.md
```diff
@@ -144,18 +144,57 @@ circuit MyTop...
 
 ## Circuits
 
-All FIRRTL circuits consist of a list of modules, each representing a hardware
-block that can be instantiated. The circuit must specify the name of the
-top-level module.
+A FIRRTL circuit is a named collection of FIRRTL modules.  Each module is a
+hardware "unit" that has ports, registers, wires, and may instantiate other
+modules (see: [@sec:modules]).  (This is the same concept as a Verilog
+`module`{.verilog}.)  A public module is usable outside of the current circuit.
```
mwachs5, line 150:
```suggestion
`module`{.verilog}.)  A public module, when lowered, will be usable outside of the context of the current circuit.
```

mwachs5: i realized upon re-reading this that we are talking about usability at the verilog level, unless we also mean usability at a hypothetical firrtl-linking level, which i think is not what we are trying to say.

spec.md
```diff
@@ -3724,7 +3800,7 @@ The following example shows the info tokens included:
 
 ``` firrtl
 circuit Top : @[myfile.txt 14:8]
-  module Top : @[myfile.txt 15:2]
+  public module Top : @[myfile.txt 15:2]
     output out: UInt @[myfile.txt 16:3]
     input b: UInt<32> @[myfile.txt 17:3]
     input c: UInt<1> @[myfile.txt 18:3]
```
mwachs5, line 3806: i think we need to add the `public` keyword to the grammar?

And release notes

seldridge:
> Public modules may not be deduplicated, and depending on where things land re:"can share any modules" may mean no dedup across public instantiation trees (as that would cause sharing).

A public module can be deduplicated, but only such that this preserves the public guarantees. I'm talking about replace one instance with another instance, not deleting deduplicated modules, though. I.e., I don't see why we should disallow changes to the instance hierarchy involving public modules just as we don't disallow changes that arise from directed or cost-driven inlining.  Consider the following:

```
circuit Foo:
  public module Bar:
    output a: UInt<1>
    connect a, UInt<1>(0)
  public module Baz:
    output a: UInt<1>
    connect a, UInt<1>(0)
  public module Foo:
    output a: UInt<1>
    inst bar of Bar
    inst baz of Baz
    connect a, xor(bar.a, baz.a)
```

A legal compilation of this to Verilog must produce three modules: `Foo.sv`, `Bar.sv`, and `Baz.sv`. However, there are many legal compilations of the _internals of Foo_. All of the following are legal FIRRTL representations of Foo:

Constant propagation rips out all instances:
```
public module Foo:
  output a: UInt<1>
  connect a, UInt<1>(0)
```

`Baz` deduplicates into `Bar` and instantiations of `Baz` are replaced with `Bar`:
```
public module Foo:
  output a: UInt<1>
  inst bar of Bar
  inst baz of Bar
  connect a, xor(bar.a, baz.a)
```

`Bar` deduplicates into `Baz` and instances of `Bar` and replaced with `Baz`:
```
public module Foo:
  output a: UInt<1>
  inst bar of Baz
  inst baz of Baz
```

> Public modules marked inline maybe is okay (above) but unlike today where it also implies "I do not expect to see this in my output" the public module must always be present in the output, because, that's its whole thing.

Agree. Today, this roughly means "always inline an instantiation of this module into its instantiator".  This still compose with `public`.  It means that the module has to be kept around. (This is just restating your observation.)

spec.md
seldridge: I'm also going to switch to saying "available" and not "usable" as the former language shows up in the LLVM linkage type that Will referenced.

spec.md
seldridge: I think we're punting on FIRRTL linking and saying that this always happens at the Verilog level.  This would be analogous to a C++ header file being enough to determine the name of a function that will be compiled by another process.  There's no C++ "linking" per-se.  However, for this to work there needs to be a paired FIRRTL `extmodule` that has the same name and ports.  (Which is the same as the C++ case.)

dtzSiFive:
> A public module can be deduplicated, but only such that this preserves the public guarantees. I'm talking about replace one instance with another instance, not deleting deduplicated modules, though

Very good point.  Awesome.  And also a non-public module could be dedup'd /into/ a public one, I think.

Thanks for going through my comments and very glad for this direction! :tada:

abi.md
```diff
@@ -69,7 +77,7 @@ packed vector of equivalent size.  For example, consider the following FIRRTL:
 
 ```FIRRTL
 circuit Top :
-  module Top :
+  public module Top :
     output out: UInt<16>
     input b: UInt<32>
 ```
```
dtzSiFive, line 83: Is it intended non-main modules should be something a separate FIRRTL file can make use of via, say, `extmodule`?

Since there's no ability to specify the circuit name for an extmodule (the circuit the module is "imported" from?) ABI relying on circuit name across boundaries are unable to match up.  In particular I mean the ref ABI in the text below.  This is an issue an practice and would be good to fix as part of this.

Since we're requiring the public module name to be the verilog name , and only really "link" at Verilog level, it seems any namespacing provided by including the circuit name is not useful in practice as there would be a naming conflict of Verilog modules already (although I do like the idea, esp leaving door open for such).

While this ABI is clear about a single unit ("what is required/can be expected re:Verilog output"), it doesn't really say how a FIRRTL compiler is supposed to make use of this on the instantiation-side from a separate FIRRTL (or anything else wishing to make use of a FIRRTL-ABI-built "circuit"), so maybe there's some room here-- in addition to how they're used ("module name" needs to be "defname" in practice) it's also unspecified how they're gathered/included such that the macros are available reliably (left to build flow)...  (not unreasonably so, bear with me please)...

One could imagine requiring telling the build flow/driver/linker+compiler that module X is from circuit Y or something (since build flow knows what circuits it built anyway, even if individual builds either don't have that information or we'd like to be able to change the definitions being used (so don't want to bake circuit name in the code)?), leaving circuit name in the filenames for co-located multiple implementations of a thing but macros are just `ref_<module>_` as within any single "linked" design the module name must be unique.

Similar but maybe simpler possibility is to just drop the circuit portion of the ref ABI (`ref_<module>` prefix for files/macros), or if wishing backwards-compat with things built using ABI in current document, then treat all modules as being within their own circuit `ref_<module>_<module>`.

I'll think some more, maybe we should discuss a bit.  Thanks.

seldridge: This makes a lot of sense. Thanks for writing this up. It took me a couple of days to process. 😅 😓 

There's two things here:

1. To properly handle linking across a boundary, only the _module name_ should be used. This implies that some `extmodule Foo` should match up to some `public module Foo` in another compilation or it's Verilog/VHDL equivalent. This _shouldn't_ specify the circuit name as the circuit where this comes from is irrelevant. It's up to the user to figure this out and link it properly.
2. There is a namespacing problem with private modules. At minimum, these should have the circuit name included. At best, these should use some name mangling which cannot be linked from another FIRRTL compilation. E.g., Verilog allows for `$` in module names, but the FIRRTL spec disallows this. We could use this fact to generate a unique private module name which cannot conflict with something else like `<circuit>$<module>`. It's not pretty, but would work. The complication here is that if the spec _isn't_ silent about this, then it imposes an implicit private module ABI which could be used to link things silently (except that the private ABI can't be accessed from FIRRTL due to the use of `$`?).

WDYT?

seldridge: The only problem with using the bare module name without the circuit for this is that conflicts are high. clang/gcc avoid this on the C++ side with more complicated name mangling. We could do something like that that includes the names of the ports and their types in a mangled way. This would be super ugly... Perhaps its best to use the module name, but expect/encourage people to use names that are prefixed intelligently and/or use very descriptive names.

seldridge: I added language that makes things more "public module-centric" in https://github.com/chipsalliance/firrtl-spec/pull/130/commits/ce8aa58bf765c673bb26bcb89b764bd560d1c2e3.  Primarily, this means that the circuit name has no real effect anymore. The only place I see that being used in in the implementation-defined convention for private modules.

E.g., consider the following circuit:

```
circuit Foo:
  module Baz:

  public module Bar:
    inst baz of Baz
```

This must produce a Verilog module called `Bar` in a file `Bar.sv`. If it produces a Verilog module for FIRRTL module `Baz`, it should mangle its name to avoid a collision with another Verilog module. Some examples/ideas:

- `Foo_Baz` (simple, to the point)
- `_Foo_Baz`
- `Foo$Baz` (has the benefit of being legal in FIRRTL, but not in Verilog)

The primary benefit of using the circuit name here is that it is trivial to change without an effect on the rest of the design now. This then fills one of, though not all of, the rolls of the `NestedPrefixModulesAnnotation`.

dtzSiFive: This sounds great!! I don't know how to "manage" the versioning details here -- I'd kinda thought the ABI was mostly versioned orthogonal to the FIRRTL spec, so this change should be ABI v3 and v4 or something?

Whatever seems best, I'm not sure how important that is presently, or if I'm missing something.

Anyway I think everything you've changed and your solutions/answers look great and excited for this direction.  Thanks! :rocket:

revision-history.yaml
```diff
@@ -5,7 +5,10 @@ revisionHistory:
   # additions to the specification should append entries here.
   thisVersion:
     spec:
+      - Add public modules.  Change circuits to support any number of public
+        modules which may have disjoint instance hierarchies.
     abi:
+      - Add ABI for public modules and filelist output.
```
dtzSiFive, line 11: `- Changed ref ABI` (modulo phrasing?)

abi.md
dtzSiFive: Oh, looks like ABI is versioned with FIRRTL spec ? And v1/v2 are parts of that or something.  So probably nevermind :+1:.

revision-history.yaml
dtzSiFive: (mostly "mention if something previously documented to work in a certain way has changed", however)

spec.md
dtzSiFive: Spec currently contains this language:

> Modules have the property that instances can always be *inlined* into the parent
module without affecting the semantics of the circuit.

Inlining I think usually means replacing a thing with its definition, not necessarily deleting the original.  Often the latter is kinda assumed since often after inlining it everywhere the definition is now dead.  But as-is I'd read this to mean it's fine to indeed copy any module's definition out (?) to replace an instance.  With or without public -- but with public it'll never be dead so just want to call this out a bit.

You're absolutely right that the first bullet, by allowing optimizing A using B's definition is essentially equivalent to allowing (at least in some senses) inlining B's definition entirely. :+1:

Maybe we need some way to express/manage that the "identity" (and count) of certain things matter, as well as perhaps their position in some way.  For example, probably don't want to be cloning registers around? But dunno.

Pinning this down would help a lot I think, and has been run into on other related fronts.

Anyway good points, personally I think this should be allowed (copying B into A) and if it's not we should figure out why precisely and pin that down.

revision-history.yaml
seldridge: Good catch. I added language indicating that the ABI for refs and groups changed in https://github.com/chipsalliance/firrtl-spec/pull/130/commits/924afa7e2b0add00e91cf90df5eaebfbe1280e8a.

abi.md
seldridge: I'm glad this makes sense. Thanks for checking the details.

Yes, the ABI _document_ is co-versioned with the FIRRTL spec. However, the actual ABI can have different "versions" within it. ABI v1/v2 is really "lowered aggregates" and "preserved aggregates". This gets a little complicated as maybe we should then do the same thing for the ref and group ABIs given that this patch makes a change to them. I think we can dodge this and change it with the co-versioned version.

I.e., if you are compiling FIRRTL less than version 3 then you should be generating the files with a certain format. If you are compiling FIRRTL version 4 or greater, then you should generate the new format. I would like to avoid this complexity in CIRCT and do a hard switch to version 4. If we have to, we can add in support for version 3 if there is a use case. However, I'm not planning to add it.

spec.md
darthscsi, line 150: Perhaps "Public modules may be instantiated outside the circuit".  This defined "usable" more narrowly.

Further, something about public modules being the exported identifiers for a circuit and all other module kinds not being namable outside the circuit might help clarify.

spec.md
```diff
@@ -144,18 +144,57 @@ circuit MyTop...
...
+
+  public module Baz :
+    inst qux of Qux
+
+  module Qux :
 ```
 
+A circuit that contains no public modules is trivially equivalent to a circuit
+that contains no modules.  It is not enforced that a circuit has at least one
+public module, but it is expected that it does.
```
darthscsi, line 196: Is the empty circuit legal in firrtl currently?

I think I'm find with the empty circuit being legal.

spec.md
```diff
@@ -179,6 +218,49 @@ physically present in the final circuit. Refer to the description of the
 instance statement for details on how to instantiate a module
 ([@sec:instances]).
 
+### Public Modules
+
+A public module is a module that is marked with the `public`{.firrtl} keyword.
+A public module has a defined ABI---knowing the FIRRTL representation is
+sufficient to know its compiled representation.  This means that a public module
+is available external to this circuit.
```
darthscsi, line 226: I don't think this strictly follows.  I could know the representation, but migrate all the contents to the in-circuit instantiators.  Public requires stronger semantics.

spec.md
```diff
@@ -179,6 +218,49 @@ physically present in the final circuit. Refer to the description of the
...
+  output b: UInt<1>
+
+  connect b, a
+```
+
+A public module has a number of restrictions:
+
+1. A public module may have no ports of uninferred width.
+1. A public module may have no ports of abstract reset type.
+1. A public module may have no ports of input probe type.
```
darthscsi, line 242: This pains me, but I don't see a way around it.

spec.md
darthscsi: This is in fact stated later, which makes this partially redundant, but in a way which emphasizes just a part as though that was the most important thing.

spec.md
```diff
@@ -179,6 +218,49 @@ physically present in the final circuit. Refer to the description of the
...
+1. A `RWProbe`{.firrtl} may not be used to access a public module's ports.
+1. A public module's ports are lowered using one of the port lowering ABIs
+   defined in the FIRRTL ABI specification.
+1. A public module may be instantiated by other modules, but must be compiled
```
darthscsi, line 246: Alternately: "A public module may be instantiated by other modules within a circuit, but the behavior of the module must not be affected by these instantiations."

abi.md
```diff
@@ -36,30 +36,69 @@ may evolve as new FIRRTL constructs are added.
...
-Any module considered a "public" module shall be implemented in Verilog in a
-consistent way.  Any public module shall exist as a Verilog module of the same
-name.
+### Public Modules
 
-Each public module with definitions (e.g. not external modules) shall be placed
-in a file with the same name.  No assumption is made of the filename of an
-implementation of an external module, it is the user's job to include files in
-such a way in their tools to resolve the name.
+Any public modules shall be implemented in Verilog in a consistent way.  Any
```
darthscsi, line 57: /Any/All/

darthscsi: in both sentences.

abi.md
```diff
@@ -36,30 +36,69 @@ may evolve as new FIRRTL constructs are added.
...
+filename of the file containing the public module and any necessary files that
+define all public or private module files instantiated under it.  Files that
+define external modules are not included.  This filelist shall have a name as
+follows where `module` is the name of the public module:
+
+``` ebnf
+filelist_filename = "filelist_" , module , ".f" ;
+```
+
+The filelist contents shall include the name of the file containing the public
```
darthscsi, line 76: does file order matter?  Are there tools that would require modules appear before instantiations?  Is this over specifying?

abi.md
```diff
@@ -86,16 +125,16 @@ module Top(
 Ports of aggregate type shall be scalarized according to the "Aggregate Type
 Lowering" description in the FIRRTL spec.
 
-Ports of ref type shall be lowered to a Verilog macro of the form `` `define
-ref_<circuit name>_<module name>_<portname> <internal path from module>`` in a
-file with name `ref_<circuit name>_<module name>.sv`.  References to aggregates
-will be lowered to a series of references to ground types.  This ABI does not
-specify whether the original aggregate referent is scalarized or not.
+Ports of ref type on public modules shall, for each public module, be lowered to
+a Verilog macro of the form `` `define ref_<module name>_<portname> <internal
```
darthscsi, line 129: I'm not sure we should drop the circuit name.   More generally, this would require extmodules to be in a separate circuit and we would need to be able to instantiate modules in other circuits.

abi.md
```diff
@@ -137,13 +176,15 @@ used by a nested group will create output ports because SystemVerilog disallows
...
+For each optional group and public module, one binding file will be produced for
+each group or nested group.  The public module, group name, and any nested
+groups are used to derive a predictable filename of the binding file.  The file
+format uses the format below where `module` is the name of the public module,
+`root` is the name of the root-level group and `nested` is the name of zero or
+more nested groups:
 
 ``` ebnf
-filename = "groups_" , circuit , "_", root , { "_" , nested } , ".sv" ;
+filename = "groups_" , module , "_", root , { "_" , nested } , ".sv" ;
```
darthscsi, line 187: same comment as before, groups_circuit_module seems safer.

darthscsi (approved):
Generally approve.  Things can be cleaned up as we get experience with what works.

abi.md
dtzSiFive: If we have the circuit name -- and no way to declare the circuit extmodules are in -- how does this work? That's why I think dropping circuit name is helpful, but mostly I don't understand what you're saying ("this would require" -- which are you saying requires that--including circuit or not?).

Are you saying you think we should put extmodules in separate circuit (so that we know what circuit-name the ABI will put their macro under? (and compatibility limits use of one compilation unit to "linking" against definitions of that module that use the "expected" circuit name, not any definition?)), or that we don't want to do that?

If so, how does one use a public module (that isn't top) with probes from another FIRRTL circuit?  What does the extmodule look like and how do we find the right ABI bits to use? Thanks!

spec.md
```diff
@@ -179,6 +218,49 @@ physically present in the final circuit. Refer to the description of the
...
+
+  connect b, a
+```
+
+A public module has a number of restrictions:
+
+1. A public module may have no ports of uninferred width.
+1. A public module may have no ports of abstract reset type.
+1. A public module may have no ports of input probe type.
+1. A `RWProbe`{.firrtl} may not be used to access a public module's ports.
```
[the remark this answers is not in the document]
dtzSiFive: If we clarify the difference, that would help yes, and related issue of whether either is expected to actually force the ports themselves necessarily or only have in some sense equivalent behavior.

This restriction (currently in the spec, just described here now) is to ensure a FIRRTL compiler can emit what's needed on /both/ sides of a module with forced ports, to give room for bounce wires or whatever, since "may be forced" is not part of the signature/ABI.

Emitting ports (and other signals) as variables would mostly eliminate the need for this restriction, as you mention, and we should clarify the difference between rwprobe'ing instance results vs the ports from inside the module so that on, e.g., extmodule's that may not have ports as variables, there's no surprises.

Sound good?

spec.md
seldridge: I'll drop this and switch to simpler language about public being capable of being instantiated externally.

abi.md
darthscsi: Or maybe an alternate path is to drop the name from the circuit?  Does it really provide any value anymore?

Basically, the question is does the circuit name matter for anything anywhere?  If it does, it seems we should embed it in the ABI so as to avoid name conflicts when one circuit uses names from another.  This obviously doesn't happen now, but it seems like it should.  Multiple public modules provides most of what you need to have libraries in a namespace.    If it doesn't matter, then sure, drop it all over.

The "would require" thing is referring to specifying the circuit extmodules are in, if circuit is an ABI-affecting namespace.

seldridge: Let's drop it. (Do we go a step further and drop circuits entirely? Probably no, but it's an interesting consideration.) If we don't then it's a hint to the compilation on how to prefix or mangle. However, this is weird as it isn't exactly specified in the ABI.

If we instead remove the name then a prefix can be later added as an optional attribute of the circuit and done in a more exact ABI way.

Other questions/design points: can a circuit be nested or should we design with this in mind? This would make circuit more like a namespace and would probably require dropping the top level circuit. This would require circuit names.

spec.md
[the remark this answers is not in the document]
mmaloney-sf: What constitutes a "use"?

Is there an intuition for why this rule is in place?

spec.md
```diff
@@ -179,6 +220,47 @@ physically present in the final circuit. Refer to the description of the
...
+1. A public module's ports are lowered using one of the port lowering ABIs
+   defined in the FIRRTL ABI specification.
```
mmaloney-sf, line 245: This doesn't seem like it should be a "restriction". This is a commentary on this proposal's relation to the ABI. 

We should scootch it down a few lines, instead of having it appear in this list. And I haven't read the latest version of the ABI spec, but when this says "one of the port lowering ABIs", the question that comes to my mind is, "which one? And who decides?"

mmaloney-sf: Also, remove "The last point requires clarification" from the next sentence. This is just evidence that commentary on ABI is a piece of commentary and not a restriction :)

spec.md
```diff
@@ -179,6 +220,47 @@ physically present in the final circuit. Refer to the description of the
...
+1. A public module may be instantiated by other modules within a circuit, but
+   the behavior of the module must not be affected by these instantiations.
+
+The last point requires clarification.  A public module may or may not be
```
mmaloney-sf, line 249: Let's not use the phrase "may or may not" here, since either one alone sounds like a prescription.

Reading this closer, this whole paragraph is confusing to me. What does it mean to be "compiled" with respect to the rest of the spec? It comes off as if it can be instantiated without being instantiated :)

spec.md
seldridge: Agree. This isn't a restriction. I'll drop it. You're question about which one/who decides is important and the current spec is silent on this. The only way I know of to properly fix this is to add the port lowering convention to each external module and any public module. For now, the spec can choose to be silent about this. However, we should add this.

Will drop the "requires clarification" sentence.

spec.md
seldridge: The problem here is that there is no good way of indicating commentary in the spec. The RISC-V specifications have shaded blocks where they do this and this may be useful to add to the FIRRTL spec. 

None of the body of this paragraph has any bearing on the spec and is only providing commentary on how to interpret the last item (line 246) if you are building a FIRRTL compiler. If you need to produce Verilog from a public module, this will make sense to you. If you are trying to understand FIRRTL, then it has no meaning.

I'll move some of this language into the ABI doc as that is more appropriate.

spec.md
seldridge: Fixed in 90f91c0.

abi.md
seldridge: Circuit names are dropped in [1b7f50b](https://github.com/chipsalliance/firrtl-spec/pull/130/commits/1b7f50badb1376f26543cbb85a0e84132d9c6ca3). This also removes the redundancy between an annotation without a target and an annotation targeting the circuit.

spec.md
```diff
@@ -144,18 +144,57 @@ circuit MyTop...
...
+    output b: UInt<1>
+
+    inst bar of Bar
+    connect bar.a, a
+    connect b, bar.b
+```
+
+A circuit may or may not contain a module with the same name as the circuit.  A
+circuit may contain public modules that instantiate no common instances.  Two
+public modules may share common instances.
```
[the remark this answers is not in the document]
seldridge: I'll just drop this. The second example should be explanation enough.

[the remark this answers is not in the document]
seldridge: Fully removed in 5676d11.

abi.md
```diff
@@ -36,30 +36,69 @@ may evolve as new FIRRTL constructs are added.
...
+```
+
+Each public module in a circuit shall produce a filelist that contains the
+filename of the file containing the public module and any necessary files that
+define all public or private module files instantiated under it.  Files that
+define external modules are not included.  This filelist shall have a name as
+follows where `module` is the name of the public module:
+
+``` ebnf
+filelist_filename = "filelist_" , module , ".f" ;
```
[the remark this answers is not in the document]
seldridge: Not doing this as circuits are unnamed now.

abi.md
seldridge: Yes, this is a weird thing to bake into the ABI. I'll change this to say that this is a newline delimited list of filenames.

seldridge: Changes in 996e307.

PR #135, opened 2023-10-24
[nfc] Define Circuit Components

mmaloney-sf:
As of version 3.2.0, the FIRRTL Spec makes several references to a notion called a "circuit component", including in the introductory chapter.
This latent concept seems to capture the important notion we informally refer to at the Chisel level as "hardware".
I'm adding a section early on to properly define this concept and to raise salience of this important notion of FIRRTL.

The sections these references appear are given with the numbers above each section:

2.1
Chisel designers manipulate circuit components using Scala functions, encode their interfaces in Scala types, and use Scala’s object-orientation features to write their own circuit libraries.

7.2.2
In a connection between circuit components with bundle types, the data carried by the flipped fields flow in the opposite direction as the data carried by the non-flipped fields.

7.3
For use in cross-module references (hierarchical references in Verilog), a reference to a probe of the circuit component is used. See Section 8.14 for details.

7.7
It is inappropriate for some circuit components to be declared with a type that allows for data to flow in both directions. For example, all sub-elements in a memory should flow in the same direction. These components are restricted to only have a passive type.

8.1
The connect statement is used to specify a physically wired connection between two circuit components.

8.1
Connect statements from a narrower ground type component to a wider ground type component will have its value automatically sign-extended or zero-extended to the larger bit width. The behavior of connect statements between two circuit components with aggregate types is defined by the connection algorithm in Section 8.1.1.

8.1.2
When a connection to a sub-element of an aggregate component is followed by a connection to the entire circuit component, the later connection overwrites the earlier sub-element connection.

8.3
A wire is a named combinational circuit component that can be connected to and from using connect statements.

8.4
A register is a named stateful circuit component. Reads from a register return the current value of the element, writes are not visible until after a positive edges of the register’s clock port.

8.5
An invalidate statement is used to indicate that a circuit component contains indeterminate values (see Section 16.1). It is specified as follows: 

```
wire w: UInt 
invalidate w 
```

Invalidate statements can be applied to any circuit component of any type. However, if the circuit component cannot be connected to, then the statement has no effect on the component. This allows the invalidate statement to be applied to any component, to explicitly ignore initialization coverage errors. The following example demonstrates the effect of invalidating a variety of circuit components with aggregate types. See Section 8.5.1 for details on the algorithm for determining what is invalidated.

8.5
The following example demonstrates the effect of invalidating a variety of circuit components with aggregate types. See Section 8.5.1 for details on the algorithm for determining what is invalidated.

8.5.5
The conditional statement creates a new scope within each of its when and else branches. It is an error to refer to any component declared within a branch after the branch has ended. As mention in Section 14, circuit component declarations in a module must be unique within the module’s flat namespace; this means that shadowing a component in an enclosing scope with a component of the same name inside a conditional statement is not allowed.

8.8.6
In the case where a connection to a circuit component is followed by a conditional statement containing a connection to the same component, the connection is overwritten only when the condition holds. Intuitively, a multiplexer is generated such that when the condition is low, the multiplexer returns the old value, and otherwise returns the new value. For details about the multiplexer, see Section 9.8.

8.8.6
The behavior of conditional connections to circuit components with aggregate types can be modeled by first expanding each connect into individual connect statements on its ground elements (see Section 8.1.1 for the connection algorithm) and then applying the conditional last connect semantics.

9
FIRRTL expressions are used for creating constant integers, for creating literal property type expressions, for referring to a declared circuit component, for statically and dynamically accessing a nested element within a component, for creating multiplexers, for performing primitive operations, and for reading a remote reference to a probe.

9.4
A reference is simply a name that refers to a previously declared circuit component. It may refer to a module port, node, wire, register, instance, or memory.

11
The flow of a reference to a declared circuit component depends on the kind of circuit component. A reference to an input port, an instance, a memory, and a node, is a source. A reference to an output port is a sink. A reference to a wire or register is duplex.

12
For all circuit components declared with unspecified widths, the FIRRTL compiler will infer the minimum possible width that maintains the legality of all its incoming connections. If a component has no incoming connections, and the width is unspecified, then an error is thrown to indicate that the width could not be inferred.

14
Each module has an identifier namespace containing the names of all port and circuit component declarations. Thus, all declarations within a module must have unique names.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -599,7 +599,7 @@ since `Passthrough` is not a type.
 
 The principal type of a submodule instance is bundle type determined by its ports.
 Each port creates a field in the bundle of the same name.
-Among these fields, `input` ports are flipped, while `output` fields are unflipped.
+Among these fields, `output` ports are flipped, while `input` fields are unflipped.
 
 For example:
 
--- a/spec.md
+++ b/spec.md
@@ -541,7 +541,7 @@ For both variants of register, the type is given after the colon (`:`{.firrtl}).
 
 Semantically, registers become flip-flops in the design.
 The next value is latched on the positive edge of the clock.
-For registers without a reset, the initial value is indeterminate (see [@sec:indeterminate-values]).
+The initial value of a register is indeterminate (see [@sec:indeterminate-values]).
 
 
 ## Output Ports and Input Ports
--- a/spec.md
+++ b/spec.md
@@ -1267,10 +1267,6 @@ module MyModule :
   connect myoutput, myinput
 ```
 
-TODO: Define a notion of "connectable circuit component".
-TODO: These currently include only wires, registers, and ports.
-TODO: Every "connectable circuit component" has a "principle type" which governs what it can connect to.
-
 In order for a connection to be legal the following conditions must hold:
 
 1.  The types of the left-hand and right-hand side expressions must be equivalent (see [@sec:type-equivalence] for details).
@@ -1385,9 +1381,9 @@ connect c, d
 The empty statement is most often used as the `else`{.firrtl} branch in a conditional statement, or as a convenient placeholder for removed components during transformational passes.
 See [@sec:conditionals] for details on the conditional statement.
 
-## Wires
+## Wire
 
-TODO: Moved to circuit components
+See [@sec:wires].
 
 ## Registers
 
@@ -1402,8 +1398,6 @@ Registers may be declared without a reset using the `reg`{.firrtl} syntax and wi
 
 ### Registers without Reset
 
-TODO: Do I remove this?
-
 The following example demonstrates instantiating a register with the given name
 `myreg`{.firrtl}, type `SInt`{.firrtl}, and is driven by the clock signal
 `myclock`{.firrtl}.
@@ -1415,8 +1409,6 @@ reg myreg: SInt, myclock
 ```
 
 ### Registers with Reset
-
-TODO: Do I remove this?
 
 A register with a reset is declared using `regreset`{.firrtl}.  A
 `regreset`{.firrtl} adds two expressions after the type and clock arguments: a
--- a/spec.md
+++ b/spec.md
@@ -1381,8 +1381,9 @@ TODO: Moved to circuit components
 
 ## Registers
 
-A register is a named stateful circuit component.
-Reads from a register return the current value of the element, writes are not visible until after a positive edges of the register's clock port.
+A register is a stateful circuit component.
+Reads from a register return the current value of the element,
+writes are not visible until after the next positive edge of the register's clock.
 
 The clock signal for a register must be of type `Clock`{.firrtl}.
 The type of a register must be a passive type (see [@sec:passive-types]) and may not be `const`{.firrtl}.
@@ -3100,8 +3101,10 @@ The flow of all other expressions are source.
 
 # Width Inference
 
-For all circuit components declared with unspecified widths, the FIRRTL compiler will infer the minimum possible width that maintains the legality of all its incoming connections.
-If a component has no incoming connections, and the width is unspecified, then an error is thrown to indicate that the width could not be inferred.
+For all circuit components declared with unspecified widths,
+the FIRRTL compiler will infer the minimum possible width that maintains the legality of all its incoming connections.
+If a circuit component has no incoming connections, and the width is unspecified,
+then an error is thrown to indicate that the width could not be inferred.
 
 For module input ports with unspecified widths, the inferred width is the minimum possible width that maintains the legality of all incoming connections to all instantiations of the module.
 
--- a/spec.md
+++ b/spec.md
@@ -492,9 +492,8 @@ Wires represent named expressions whose value is determined by FIRRTL `connect`{
 
 Example:
 
-``` firrtl
+```firrtl
 wire mywire: UInt<1>
-
 connect mywire, UInt<1>(0)
 ```
 
@@ -516,17 +515,17 @@ Optionally, registers may have a reset signal.
 On every cycle, a register will drive its current value
 and then latch the value it will take on for the next cycle.
 
-The `regreset` keyword is used to declare a register with a reset.
-The `reg` keyword is used to declare a register without a reset.
+The `regreset`{.firrtl} keyword is used to declare a register with a reset.
+The `reg`{.firrtl} keyword is used to declare a register without a reset.
 
 Examples:
 
-``` firrtl
+```firrtl
 wire myclock : Clock
 reg myreg : SInt, myclock
 ```
 
-``` firrtl
+```firrtl
 wire myclock : Clock
 reg myreg : SInt, myclock
 ```
@@ -550,7 +549,7 @@ The way a module interacts with the outside world is through its ports.
 
 Example:
 
-``` firrtl
+```firrtl
 input myinput : UInt<1>
 output myinput : SInt<8>
 ```
@@ -569,12 +568,12 @@ Example:
 inst passthrough of Passthrough
 ```
 
-This assumes you have a module named `Passthrough` declared elsewhere in the current circuit.
+This assumes you have a module named `Passthrough`{.firrtl} declared elsewhere in the current circuit.
 The keyword `of`{.firrtl} is used instead of a colon (`:`{.firrtl}).
 
 The type of a submodule instance is bundle type determined by its ports.
 Each port creates a field with the same name in the bundle.
-Among these fields, `output` ports are flipped, while `input` fields are unflipped.
+Among these fields, `output`{.firrtl} ports are flipped, while `input`{.firrtl} fields are unflipped.
 
 For example:
 
@@ -585,7 +584,7 @@ module Passthrough :
   connect out, in
 ```
 
-The type of the submodule instance `passthrough` above is thus:
+The type of the submodule instance `passthrough`{.firrtl} above is thus:
 
 ```firrtl
 { flip in : UInt<8>, out : UInt<8> }
@@ -612,7 +611,7 @@ mem mymem :
 
 The type of a memory is a bundle type derived from the declaration (see [@sec:mem]).
 
-The type named in `data-type` must be passive.
+The type named in `data-type`{.firrtl} must be passive.
 It indicates the type of the data being stored inside of the memory.
 
 See [@sec:mem] for more details.
--- a/spec.md
+++ b/spec.md
@@ -493,7 +493,7 @@ Wires represent named expressions whose value is determined by FIRRTL `connect`{
 Example:
 
 ```firrtl
-wire mywire: UInt<1>
+wire mywire : UInt<1>
 connect mywire, UInt<1>(0)
 ```
 
@@ -505,10 +505,8 @@ The type of a wire is given after the colon (`:`{.firrtl}).
 
 Registers are stateful elements of a design.
 
-Registers can be connected (see [@sec:connects]).
-The state of the register is controlled through what is connected to it.
-
-The state of a register may be any non-`const`{.firrtl} passive type (see [@sec:passive-types]).
+The state of a register is controlled through what is connected to it (see [@sec:connects]).
+The state may be any non-`const`{.firrtl} passive type (see [@sec:passive-types]).
 Registers are always associated with a clock.
 Optionally, registers may have a reset signal.
 
@@ -545,7 +543,7 @@ The initial value of a register is indeterminate (see [@sec:indeterminate-values
 
 ## Output Ports and Input Ports
 
-The way a module interacts with the outside world is through its ports.
+The way a module interacts with the outside world is through its output and input ports.
 
 Example:
 
--- a/spec.md
+++ b/spec.md
@@ -214,18 +214,12 @@ I.e., the identifier of a top-level group declared in a circuit must not conflic
 Each optional group declaration must include a string that sets the lowering convention for that group.
 The FIRRTL ABI specification defines supported lowering convention.  One such strategy is `"bind"`{.firrtl} which lowers to modules and instances which are instantiated using the SystemVerilog `bind`{.verilog} feature.
 
-The `group`{.firrtl} keyword defines optional functionality inside a module.  An
-optional group may only be defined inside a module.  An optional group must
-reference a group declared in the current circuit.  Declarations of identifiers
-and references to existing identifiers following the same lexical scoping rules
-as FIRRTL conditional statements (see: [@sec:conditional-scopes])---identifiers declared in
-the group definition may not be used outside the group while groups may refer to
-identifiers declared outside the group.  __The statements in a group are
-restricted in what identifiers they are allowed to drive.__ A statement in a
-group may drive no sinks declared outside the group _with one exception_: a
-statement in a group may drive reference types declared outside the group if the
-reference types are associated with the group in which the statement is declared
-(see: [@sec:reference-types]).
+The `group`{.firrtl} keyword defines optional functionality inside a module.
+An optional group may only be defined inside a module.
+An optional group must reference a group declared in the current circuit.
+Declarations of identifiers and references to existing identifiers following the same lexical scoping rules as FIRRTL conditional statements (see: [@sec:conditional-scopes])---identifiers declared in the group definition may not be used outside the group while groups may refer to identifiers declared outside the group.
+__The statements in a group are restricted in what identifiers they are allowed to drive.
+__ A statement in a group may drive no sinks declared outside the group _with one exception_: a statement in a group may drive reference types declared outside the group if the reference types are associated with the group in which the statement is declared (see: [@sec:reference-types]).
 
 The circuit below contains one optional group declaration, `Bar`.
 Module `Foo` contains a group definition that creates a node computed from a port defined in the scope of `Foo`.
@@ -1170,16 +1164,14 @@ module Example:
 
 ### Constant Type
 
-A constant type is a type whose value is guaranteed to be unchanging at circuit
-execution time.  Constant is a constraint on the mutability of the value, it
-does not imply a literal value at a point in the emitted design.  Constant types
-may be used in ports, wire, and nodes.  Operations on constant type are well defined.
-With any exception listed in the definition for such operations as have exceptions,
-an operation whose arguments are constant produces a constant.  An operation
-with some non-constant arguments produce a non-constant.  Constants may be
-used as the target of a connect so long as the source of the connect is itself
-constant.  These rules ensure all constants are derived from constant integer
-expressions or from constant-typed input ports of a public module.
+A constant type is a type whose value is guaranteed to be unchanging at circuit execution time.
+Constant is a constraint on the mutability of the value, it does not imply a literal value at a point in the emitted design.
+Constant types may be used in ports, wire, and nodes.
+Operations on constant type are well defined.
+With any exception listed in the definition for such operations as have exceptions, an operation whose arguments are constant produces a constant.
+An operation with some non-constant arguments produce a non-constant.
+Constants may be used as the target of a connect so long as the source of the connect is itself constant.
+These rules ensure all constants are derived from constant integer expressions or from constant-typed input ports of a public module.
 
 ``` firrtl
 const UInt<3>
@@ -1416,15 +1408,12 @@ reg myreg: SInt, myclock
 
 ### Registers with Reset
 
-A register with a reset is declared using `regreset`{.firrtl}.  A
-`regreset`{.firrtl} adds two expressions after the type and clock arguments: a
-reset signal and a reset value.  The register's value is updated with the reset
-value when the reset is asserted.  The reset signal must be a `Reset`{.firrtl},
-`UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}, and the type of initialization
-value must be equivalent to the declared type of the register (see
-[@sec:type-equivalence] for details).  If the reset signal is an
-`AsyncReset`{.firrtl}, then the reset value must be a constant type.  The
-behavior of the register depends on the type of the reset signal.
+A register with a reset is declared using `regreset`{.firrtl}.
+A `regreset`{.firrtl} adds two expressions after the type and clock arguments: a reset signal and a reset value.
+The register's value is updated with the reset value when the reset is asserted.
+The reset signal must be a `Reset`{.firrtl}, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}, and the type of initialization value must be equivalent to the declared type of the register (see [@sec:type-equivalence] for details).
+If the reset signal is an `AsyncReset`{.firrtl}, then the reset value must be a constant type.
+The behavior of the register depends on the type of the reset signal.
 `AsyncReset`{.firrtl} will immediately change the value of the register.
 `UInt<1>`{.firrtl} will not change the value of the register until the next positive edge of the clock signal (see [@sec:reset-type]).
 `Reset`{.firrtl} is an abstract reset whose behavior depends on reset inference.
--- a/spec.md
+++ b/spec.md
@@ -463,7 +463,9 @@ Any use in place of an integer is disallowed.
 # Circuit Components
 
 Circuit components are the named objects which may be contained within a module.
-They are: nodes, wires, registers, submodule instances, and ports.
+They are: nodes, wires, registers, ports, and submodule instances.
+
+TODO: What about `mem`\s?
 
 ## Nodes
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -477,7 +477,9 @@ integer literals.  Any use in place of an integer is disallowed.
 # Circuit Components
 
 Circuit components are the named objects which may be contained within a module.
-They are: nodes, wires, registers, submodule instances, and ports.
+They are: nodes, wires, registers, ports, and submodule instances.
+
+TODO: What about `mem`\s?
```
mmaloney-sf, line 482: Have I captured all the circuit components here? 

I vaguely remember thinking there were 7 for some reason back in April. I know I counted `output` and `input` as separate, but that leaves me feeling like I'm missing one. Maybe I had included `mem`.

spec.md
```diff
@@ -515,15 +517,77 @@ Example:
...
 
-TODO
+Registers are the stateful elements of a design.
 
-## Submodule Instances
+Registers can be connected (see [@sec:connects]).
+The state of the register is controlled through what is connected to it.
+
+The state of a register may be any passive type (see [@sec:passive-types]) and may not be `const`{.firrtl}.
+(TODO: is the `const` restriction a consequence of the passive part or not? Why is this required?)
```
mmaloney-sf, line 526: Is the `const` restriction a consequence of the passive part or not? Why is this required?

Intuitively, I get why it can't be `const`. Maybe it's just an additional restriction. But it might be important to motivate.

spec.md
```diff
@@ -474,6 +474,121 @@ The following string-encoded integer literals all have the value `-42`:
...
+
+Example:
+
+```firrtl
+    node mynode = and(in, UInt<4>(1))
+```
+
+Nodes are given a name and a definition.
+Unlike wires, they can only be assigned to at the point they are defined.
+(TODO: Do all values need to be in scope above the point of definition?)
```
mmaloney-sf, line 497: Do all values need to be in scope above the point of definition in a `node`?

mmaloney-sf: cf the TODO just below in the Wires section:

TODO: Circular definitions? c.f. with nodes.

spec.md
```diff
@@ -474,6 +474,57 @@ The following string-encoded integer literals all have the value `-42`:
 Radix-specified integer literals are only usable when constructing hardware
 integer literals.  Any use in place of an integer is disallowed.
 
+# Circuit Components
+
+Circuit components are the named objects which may be contained within a module.
+They are: nodes, wires, registers, submodule instances, and ports.
```
mmaloney-sf: What else might be missing from this list?

spec.md
```diff
@@ -474,6 +474,122 @@ The following string-encoded integer literals all have the value `-42`:
 Radix-specified integer literals are only usable when constructing hardware
 integer literals.  Any use in place of an integer is disallowed.
 
+# Circuit Components
+
+Circuit components are the named objects which may be contained within a module.
+They are: nodes, wires, registers, ports, and submodule instances.
```
mwachs5, line 480: what about "verification" statements (printfs, asserts, etc)

spec.md
```diff
@@ -474,6 +474,122 @@ The following string-encoded integer literals all have the value `-42`:
...
+```
+
+Nodes are given a name and a definition.
+Unlike wires, they can only be assigned to at the point they are defined.
+(TODO: Do all values need to be in scope above the point of definition?)
+They cannot be connected (see [@sec:connects]).
+
+## Wires
+
+A wire is a named combinational circuit component that can be connected to and from using connect statements.
```
mwachs5, line 502: can we use the `wire`{.firrtl} syntax?

spec.md
mmaloney-sf: It looks like `mem` is a circuit component. Added a section for it.

spec.md
mmaloney-sf: This is a good question. Do you have an opinion on this?

Based on the direction I was leaning, I don't think they are. They:
- do not have a principal type, and they
- may not be connected

spec.md
mmaloney-sf: The concept is wire, the syntax is `wire`{.firrtl}. I considered this, and in this context, it's better to separate the concept from the syntax. (Otherwise, "port" becomes `input` and `output` and "registers" becomes `reg` and `regreset`).

spec.md
mmaloney-sf: I was informed of `mem` earlier. Added a section on memories.

spec.md
```diff
@@ -1436,11 +1587,9 @@ wire w: UInt
 invalidate w
 ```
 
-Invalidate statements can be applied to any circuit component of any
-type. However, if the circuit component cannot be connected to, then the
```
mmaloney-sf, line 1440: As far as the rest of the spec indicates, no circuit component cannot be connected to.

Arguably submodule instances or memories were initially considered, but it is not explicitly forbidden. (Or if it is, we need to reference where and why).

spec.md
```diff
@@ -1155,18 +1293,36 @@ compilable to non-parameterized Verilog.
 
 ## Passive Types
 
-It is inappropriate for some circuit components to be declared with a type that
-allows for data to flow in both directions. For example, all sub-elements in a
-memory should flow in the same direction. These components are restricted to
-only have a passive type.
+Stateful elements, such as registers and memories, may contain data of aggregate types.
```
mwachs5, line 1296: is "data" a precisely defined term?

spec.md
```diff
@@ -1155,18 +1293,36 @@ compilable to non-parameterized Verilog.
...
+Registers with bundle types are especially common.
+However, when using bundle types in stateful elements, the notion of `flip` is meaningless.
+There is no directionality to the data inside a register; the data just is.
```
mwachs5, line 1299:
```suggestion
```

spec.md
```diff
@@ -1155,18 +1293,36 @@ compilable to non-parameterized Verilog.
...
+This is the type of data that flows through the component.
 
-All property types are passive.
+The principal type is used to determine whether two circuit components may be connected.
+
+For wires, registers, ports, the principal component is declared after the colon `:` when the circuit component is declared.
+
+For nodes, the principal type is inferred from the type of the expression on the left hand side of the equals sign `=` when the node is declared.
+
+For submodule instances, TODO: ???
```
mwachs5, line 1325: is there a type other than Principal Type?

spec.md
```diff
@@ -1206,13 +1362,15 @@ Two property types are equivalent if they are the same concrete property type.
...
-interact.
+A module body consists of a sequence of statements.
+Statements declare the circuit components and describe their connectivity.
 
 ## Connects
 
-The connect statement is used to specify a physically wired connection between
-two circuit components.
+Circuit components can be connected together.
+This connection is both physical, as if the components were connected together with a wire,
```
mwachs5, line 1371:
```suggestion
This connection can be both physical, as if the components were connected together with a wire,
```

spec.md
```diff
@@ -1386,6 +1540,8 @@ with a reset using the `regreset`{.firrtl} syntax.
 
 ### Registers without Reset
 
+TODO: Do I remove this?
```
mwachs5, line 1543: why would we remove this?

spec.md
```diff
@@ -1427,20 +1585,17 @@ A register is initialized with an indeterminate value (see
 
 ## Invalidates
 
-An invalidate statement is used to indicate that a circuit component contains
-indeterminate values (see [@sec:indeterminate-values]). It is specified as
-follows:
+An invalidate statement is used to indicate that a circuit component contains indeterminate values (see [@sec:indeterminate-values]).
```
mwachs5: it seems this is purely a whitespace change? This doc has been formatted for 80 character limit, is there a reason you are changing this line?

spec.md
mwachs5: >  no circuit component cannot be connected to.

That is not true, you are allowed to not connect to e.g. the `input` of your top level module in the circuit.

spec.md
```diff
@@ -1673,10 +1828,10 @@ match x:
 
 ### Nested Declarations
 
-If a component is declared within a conditional statement, connections to the
-component are unaffected by the condition. In the following example, register
-`myreg1`{.firrtl} is always connected to `a`{.firrtl}, and register
-`myreg2`{.firrtl} is always connected to `b`{.firrtl}.
+If a circuit component is declared within a conditional statement,
+connections to the it are unaffected by the condition.
```
mwachs5, line 1829: "to the it" --> something wrong

spec.md
mmaloney-sf: @darthscsi explained that it's possible to get a time-varying `reg` in a circuit that only has `const` types.

spec.md
mmaloney-sf: They do.

spec.md
mmaloney-sf: No.

The style that I'm going to start moving things towards is that the first sentence/paragraph of each section is informal motivation. A precise explanation of what "data" is or what it means to "contain data" would require we have rules regarding simulation of a FIRRTL design.

spec.md
mmaloney-sf: In the next few weeks, I want to reorganize a bunch of things in the doc.

The chapter on "Statements" is not the proper grouping. It maybe made sense at one point. But this section is doing much more work than it should be. 

Rather than going statement-by-statement, I am going to move us towards explaining things concept-by-concept. The section I added on Circuit Components causes some tension, begging material from the Statements section to be moved up front. 

Until I have the re-organization finished, some material might get split up (with appropriate cross-references).

spec.md
mmaloney-sf: It looks like I removed the second sentence. (And this was outdated by another change, so just keeping that in mind).

I would like to gradually shift over to using a git-friendly formatting, with one(-ish) sentence per line.

spec.md
mmaloney-sf: Thanks.

spec.md
mmaloney-sf: Thanks.

spec.md
mmaloney-sf: I'll also ensure I have a reasonable line length. I just realized my editor settings don't make it obvious for me to tell when I hit the 80 or 100 col line limit.

spec.md
mmaloney-sf: @dtzSiFive asked me something similar. I added a sentence to clarify this.

tl;dr -- In something like `wire w : UInt<1>`, the *kind* is `wire` but the *principal type* is `UInt<1>`. It's just there to linguistically emphasize we are speaking formally.

spec.md
mmaloney-sf: I pushed a change to cap all of my changes at 100 cols.

Hopefully that isn't controversial.

spec.md
mmaloney-sf: Ah. Thank you. I was trying to work out when this would be restricted.

Is there any problem with me changing this to a component that has Sink or Duplex flow? "Cannot be connected to" puts too much obligation on the reader to read between the lines.

mmaloney-sf:
I removed the "Principal Types in Circuit Components" section and moved the rule for the principal type under each individual component section. 

This gives me a bit more meat under the main heading of Circuit Components to make generic statements about all Circuit Components. Namely: they have a kind (and there are six kinds), they have a principal type, and they can be connected.

I also bolded the words "kind" and "principal type" to indicate we're introducing a technical term.

spec.md
```diff
@@ -1843,7 +2020,7 @@ connect w.a, mux(c, y, x.a)
 connect w.b, x.b
 ```
 
-## Memories
+## Mem
```
mwachs5, line 2023: I don't think we should change this heading...?

spec.md
mwachs5: Sorry why not then just say "type"?

spec.md
```diff
@@ -1155,18 +1331,24 @@ compilable to non-parameterized Verilog.
...
+More precisely, a passive type is defined recursively:
 
-Intuitively, a passive type is a type where all data flows in the same
-direction, and is defined to be a type that recursively contains no fields with
-flipped orientations. Thus all ground types are passive types. Vector types are
-passive if their element type is passive. And bundle types are passive if no
-fields are flipped and if all field types are passive.
+* All ground types are passive.
+* All probe types are passive.
+* All properties types are passive.
```
mwachs5, line 1344: Property types

spec.md
```diff
@@ -1359,23 +1544,15 @@ conditional statement, or as a convenient placeholder for removed components
...
 
-The following example demonstrates instantiating a wire with the given name
-`mywire`{.firrtl} and type `UInt`{.firrtl}.
-
-``` firrtl
-wire mywire: UInt
-```
+See [@sec:wires].
 
 ## Registers
```
mwachs5, line 1551: If you are changing these titles to be named after the keyword, which I don't think really makes sense, shouldn't this one be called Reg?

mwachs5 (commented):
Some more nits

mmaloney-sf:
After spending a few days writing with the full name of "circuit component", I'm getting to the point where it feels really verbose.

Since we already refer to these as "components" in many places in the document (and informally), I may change it uniformly to just be called a "component" formally.

spec.md
```diff
@@ -542,6 +542,172 @@ The following string-encoded integer literals all have the value `-42`:
...
+
+The type of a wire is given after the colon (`:`{.firrtl}).
+
+
+## Registers
+
+Registers are stateful elements of a design.
+
+Registers can be connected (see [@sec:connects]).
+The state of the register is controlled through what is connected to it.
```
seldridge, line 595: What is the "state" of a register? This is again on the intuition versus technical description issue.

spec.md
```diff
@@ -542,6 +542,172 @@ The following string-encoded integer literals all have the value `-42`:
...
+
+The state of a register may be any non-`const`{.firrtl} passive type (see [@sec:passive-types]).
+Registers are always associated with a clock.
+Optionally, registers may have a reset signal.
+
+On every cycle, a register will drive its current value
+and then latch the value it will take on for the next cycle.
```
seldridge, line 602: "latch" is also possibly an inaccurate term that is overloaded with hardware latches.

spec.md
```diff
@@ -542,6 +542,172 @@ The following string-encoded integer literals all have the value `-42`:
...
+## Registers
+
+Registers are stateful elements of a design.
+
+Registers can be connected (see [@sec:connects]).
+The state of the register is controlled through what is connected to it.
+
+The state of a register may be any non-`const`{.firrtl} passive type (see [@sec:passive-types]).
+Registers are always associated with a clock.
+Optionally, registers may have a reset signal.
```
seldridge, line 599: It may be simpler to state that FIRRTL has two types of registers: registers without a reset and registers with a reset.

spec.md
```diff
@@ -542,6 +542,172 @@ The following string-encoded integer literals all have the value `-42`:
...
+wire myclock : Clock
+wire myreset : UInt<1>
+wire myinit : SInt
+regreset myreg : SInt, myclock, myreset, myinit
+```
+
+For both variants of register, the type is given after the colon (`:`{.firrtl}).
+
+Semantically, registers become flip-flops in the design.
+The the next value is latched on the positive edge of the clock.
```
seldridge, line 628:
```suggestion
The next value is latched on the positive edge of the clock.
```

spec.md
```diff
@@ -542,6 +542,172 @@ The following string-encoded integer literals all have the value `-42`:
...
+For registers without a reset, the initial value is indeterminate (see [@sec:indeterminate-values]).
```
seldridge, line 629: There was some very carefully constructed language in the original register statement section that intentionally does not talk about "initial value" and instead uses "reset value". This is extremely intentional as the initial value of a register is assumed to mean "the value of the register before any reset is applied" or "the value of the register at time zero".

Put differently, the initial value of both registers with and without resets is indeterminate.

spec.md
```diff
@@ -542,6 +542,172 @@ The following string-encoded integer literals all have the value `-42`:
...
+
+## Nodes
+
+Nodes are named expressions in FIRRTL.
+
+Example:
+
+```firrtl
+    node mynode = and(in, UInt<4>(1))
+```
```
seldridge, line 568: Nit: Can the PR use a consistent indenting of these examples? I tend to always left align them.

Ditto throughout.

spec.md
```diff
@@ -542,6 +542,172 @@ The following string-encoded integer literals all have the value `-42`:
...
+Each submodule instance must be given a name.
+It must also name the FIRRTL module or external module it is instantiating.
+
+Example:
+
+```firrtl
+inst passthrough of Passthrough
+```
+
+This assumes you have a module named `Passthrough` declared elsewhere in your FIRRTL design.
```
seldridge, line 661:
```suggestion
This assumes you have a module named `Passthrough` declared elsewhere in your FIRRTL `circuit`{.firrtl}.
```

It should be stated somewhere that modules may violate def-before-use.

spec.md
```diff
@@ -542,6 +542,172 @@ The following string-encoded integer literals all have the value `-42`:
...
+The way a module interacts with the outside world is through its ports.
+
+Example:
+
+``` firrtl
+    input myinput : UInt<1>
+    output myinput : SInt<8>
+```
+
+Ports are declared with the keywords `input` and `output`.
```
seldridge, line 643: Nit throughout: there is a special syntax you can use to possible get syntax highlighting of fixed width text:

```
`input`{.firrtl}
```

spec.md
```diff
@@ -542,6 +542,172 @@ The following string-encoded integer literals all have the value `-42`:
...
+The type of the submodule instance `passthrough` above is thus:
+
+```firrtl
+{ flip in : UInt<8>, out : UInt<8> }
+```
+
+
+## Memories
+
+Memories are large stateful elements of a design.
```
seldridge, line 687: Nit as with registers: memories may be small, too. They're yet-another stateful element.

spec.md
```diff
@@ -1223,18 +1387,24 @@ compilable to non-parameterized Verilog.
...
-It is inappropriate for some circuit components to be declared with a type that
-allows for data to flow in both directions. For example, all sub-elements in a
-memory should flow in the same direction. These components are restricted to
-only have a passive type.
+Stateful elements, such as registers and memories, may contain data of aggregate types.
+Registers with bundle types are especially common.
+However, when using bundle types in stateful elements, the notion of `flip` is meaningless.
+There is no directionality to the data inside a register; the data just *is*.
+
+A passive type is a which does not make use of `flip`.
```
seldridge, line 1395:
```suggestion
A passive type is a type which does not make use of `flip`.
```

spec.md
```diff
@@ -1223,18 +1387,24 @@ compilable to non-parameterized Verilog.
...
-flipped orientations. Thus all ground types are passive types. Vector types are
-passive if their element type is passive. And bundle types are passive if no
-fields are flipped and if all field types are passive.
+* All ground types are passive.
+* All probe types are passive.
+* All property types are passive.
+* A vector type is passive if and only if the element type is passive.
+* A bundle type is passive if and only if it contains no field which is marked `flip`
+  and the type of each field is passive.
+* An enum type is passive if and only if the type of each of its variants is passive.
```
seldridge, line 1404: Mega-nit: I think the rest of the spec uses `-` for itemized lists. This doesn't affect the output, just causes my OCD to kick in.

spec.md
```diff
@@ -1223,18 +1387,24 @@ compilable to non-parameterized Verilog.
...
+
+Registers and memories may only be parametrized over passive types.
```
seldridge, line 1406: Can the PR avoid use of "parameterized". This will eventually mean something else.

spec.md
```diff
@@ -1274,13 +1444,16 @@ Two property types are equivalent if they are the same concrete property type.
...
 
 ## Connects
 
-The connect statement is used to specify a physically wired connection between
-two circuit components.
+The components of a module can be connected together using connect statements.
+
+These connections can be both physical, as if the components were connected together with a wire,
+but also logical, allowing metadata such as probes (see [@sec:probe-types;@sec:probe])
+and properties (see [@sec:property-types]) to flow across it during elaboration.
```
seldridge, line 1456: I'd avoid "physical" and "logical" as that introduces new, undefined concepts. Is there a way to do this by talking about different connections for hardware and non-hardware components?

spec.md
```diff
@@ -2643,10 +2808,12 @@ module Example:
...
-FIRRTL expressions are used for creating constant integers, for creating
-literal property type expressions, for referring to a declared circuit
-component, for statically and dynamically accessing a nested element within a
-component, for creating multiplexers, for performing primitive operations, and
+FIRRTL expressions are used for creating constant integers,
+for creating literal property type expressions,
+for referencing a circuit component,
+for statically and dynamically accessing a nested element within a component,
+for creating multiplexers,
+for performing primitive operations, and
```
seldridge, line 2816: Mega-nit: unnecessary manual text flowing.

spec.md
```diff
@@ -3398,11 +3563,11 @@ The flow of all other expressions are source.
...
-For all circuit components declared with unspecified widths, the FIRRTL compiler
-will infer the minimum possible width that maintains the legality of all its
-incoming connections. If a component has no incoming connections, and the width
-is unspecified, then an error is thrown to indicate that the width could not be
-inferred.
+For all circuit components declared with unspecified widths,
+the FIRRTL compiler will infer the minimum possible width that
+maintains the legality of all its incoming connections.
+If a circuit component has no incoming connections, and the width is unspecified,
+then an error is thrown to indicate that the width could not be inferred.
```
seldridge, line 3570: Nit: unnecessary text reflowing.

seldridge (commented):
High level: I agree with the direction! Ping me to take another look at this once comments are addressed.

Nit: The PR is inconsistent with text flow. Sometimes it's one-sentence-per-line, sometimes it's ~80 column. The rest of the document is 80-column. Please use that for consistency. (Yes this should be handled automatically with a formatter to avoid doing it with your editor text reflowing.) I'm not against switching to one-sentence-per-line as this version controls better. However, that should be a conscious decision in a separate PR.

spec.md
```diff
@@ -474,6 +474,183 @@ The following string-encoded integer literals all have the value `-42`:
 Radix-specified integer literals are only usable when constructing hardware
 integer literals.  Any use in place of an integer is disallowed.
 
+# Circuit Components
+
+Circuit components are the named parts of a module corresponding to physical hardware.
+
+There are seven **kinds** of circuit components.
+They are: nodes, wires, registers, output ports, input ports, submodule instances, and memories.
```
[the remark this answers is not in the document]
mmaloney-sf: If we don't distinguish between them, we end up having to make that distinction at another level somewhere else. It's better to treat them as distinct, even though they are tightly related to one another.

spec.md
mmaloney-sf: Changed.

spec.md
mmaloney-sf: Whoops.

spec.md
mmaloney-sf: Added to my todo.

spec.md
mmaloney-sf: There's no formal distinction, so let's just fudge it until there is.

seldridge (approved):
Nit throughout: one-sentence-per-line.

Other high level comments going forward. It would be good to synthesize this into a notion of "reference" or figure out what we should talk about when we say "an expression which is an identifier of a circuit component". Compiler IRs would normally call this a value.

spec.md
```diff
@@ -460,6 +460,161 @@ The following string-encoded integer literals all have the value `-42`:
...
+
+A module in FIRRTL may contain submodules.
+
+Example:
+
+```firrtl
+inst passthrough of Passthrough
+```
+
+This assumes you have a module named `Passthrough`{.firrtl} declared elsewhere in the current circuit.
```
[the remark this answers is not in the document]
mmaloney-sf: Changed to :

This assumes you have a `module`, `extmodule`, or `intmodule` named `Passthrough`{.firrtl} declared elsewhere in the current circuit.

spec.md
```diff
@@ -1035,16 +1189,24 @@ It is, for example, expected that a module with a constant input port be fully c
 
 ## Passive Types
 
-It is inappropriate for some circuit components to be declared with a type that allows for data to flow in both directions.
-For example, all sub-elements in a memory should flow in the same direction.
-These components are restricted to only have a passive type.
+Stateful elements, such as registers and memories, may contain data of aggregate types.
+Registers with bundle types are especially common.
+However, when using bundle types in stateful elements, the notion of `flip` is meaningless.
+There is no directionality to the data inside a register; the data just *is*.
```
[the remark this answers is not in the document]
mmaloney-sf: Reworded to "does not make sense". My regards go out to Bruce.

spec.md
```diff
@@ -3617,7 +3793,9 @@ letter = "A" | "B" | "C" | "D" | "E" | "F" | "G"
        | "o" | "p" | "q" | "r" | "s" | "t" | "u"
        | "v" | "w" | "x" | "y" | "z" ;
 
-(* Tokens: Info *)
+(* Fileinfo communicates Chisel source file and line/column info *)
+linecol = digit_dec , { digit_dec } , ":" , digit_dec , { digit_dec } ;
+lineinfo = string, " ", linecol ;
```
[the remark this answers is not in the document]
mmaloney-sf: Good catch. The merge messed this up.

spec.md
```diff
@@ -460,6 +460,161 @@ The following string-encoded integer literals all have the value `-42`:
...
+
+There are seven **kinds** of circuit components.
+They are: nodes, wires, registers, output ports, input ports, submodule instances, and memories.
+
+Circuit components can be connected together (see [@sec:connects]).
+
+Each circuit component in a module has a type ([@sec:types]).
+It is used to determine the legality of connections.
+
+
```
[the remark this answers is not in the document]
mmaloney-sf: I'll keep that in mind. Fixed wherever I saw it.

spec.md
```diff
@@ -460,6 +460,161 @@ The following string-encoded integer literals all have the value `-42`:
 Radix-specified integer literals are only usable when constructing hardware integer literals.
 Any use in place of an integer is disallowed.
 
+# Circuit Components
+
+Circuit components are the named parts of a module corresponding to hardware.
+
+There are seven **kinds** of circuit components.
+They are: nodes, wires, registers, output ports, input ports, submodule instances, and memories.
```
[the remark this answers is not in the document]
mmaloney-sf: I'll keep my eye on this. It's similar to your other comment about how `Passthrough` "needs to be a module" -- but module is being used in a broad sense to include `module` `extmodule` or `intmodule`. This should be detailed somewhere. But perhaps I'll have it in my notes for a follow-up.

mwachs5:
please mark this as patch and add something to the version history for this since it does actually change the spec document

PR #136, opened 2023-10-28
[patch] Reorganize section on Connections

mmaloney-sf:
The section Connects is in need of reorganization.

The Connect section has been lifted out of the Statements section (which I'm aiming to remove eventually in a future PR) and renamed to Connections.

The section on Flows has been moved under Connections and placed at the start (since the connection rules are defined in terms of Flow). The rules for Flow have been elaborated. Some motivation has been given. I avoid leaning on the phraseology, "all other expressions" in favor of naming the remaining classes of expressions explicitly.

I might not touch it in this PR, but I would like to keep my eye on the Last Connect Semantics section. It is now the only section which isn't related to the legality of connects, and it has interactions with conditionals. This suggests that it will probably be moved at some point in the future.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1162,6 +1162,29 @@ Consequently, `{a:UInt, b:UInt}`{.firrtl} is not equivalent to `{b:UInt, a:UInt}
 
 Two property types are equivalent if they are the same concrete property type.
 
+# Connections
+
+## Flow
+
+An expression's flow partially determines the legality of connecting to and from the expression.
+Every expression is classified as either *source*, *sink*, or *duplex*.
+For details on connection rules refer back to [@sec:connects].
+
+The flow of a reference to a declared circuit component depends on the kind of circuit component.
+A reference to an input port, an instance, a memory, and a node, is a source.
+A reference to an output port is a sink. A reference to a wire or register is duplex.
+
+The flow of a sub-index or sub-access expression is the flow of the vector-typed expression it indexes or accesses.
+
+The flow of a sub-field expression depends upon the orientation of the field.
+If the field is not flipped, its flow is the same flow as the bundle-typed expression it selects its field from.
+If the field is flipped, then its flow is the reverse of the flow of the bundle-typed expression it selects its field from.
+The reverse of source is sink, and vice-versa.
+The reverse of duplex remains duplex.
+
+The flow of all other expressions are source.
+
+
 # Statements
 
 A module body consists of a sequence of statements.
@@ -2730,26 +2753,6 @@ module MyModule :
 The probed expression must be a static reference.
 
 See [@sec:probe-types;@sec:probe] for more details on probe references and their use.
-
-# Flows
-
-An expression's flow partially determines the legality of connecting to and from the expression.
-Every expression is classified as either *source*, *sink*, or *duplex*.
-For details on connection rules refer back to [@sec:connects].
-
-The flow of a reference to a declared circuit component depends on the kind of circuit component.
-A reference to an input port, an instance, a memory, and a node, is a source.
-A reference to an output port is a sink. A reference to a wire or register is duplex.
-
-The flow of a sub-index or sub-access expression is the flow of the vector-typed expression it indexes or accesses.
-
-The flow of a sub-field expression depends upon the orientation of the field.
-If the field is not flipped, its flow is the same flow as the bundle-typed expression it selects its field from.
-If the field is flipped, then its flow is the reverse of the flow of the bundle-typed expression it selects its field from.
-The reverse of source is sink, and vice-versa.
-The reverse of duplex remains duplex.
-
-The flow of all other expressions are source.
 
 # Width Inference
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1225,28 +1391,94 @@ module MyModule :
...
+or else the right-hand side expression has a passive type.
+
+TODO: Why is a passive type acceptable?
+TODO: Is this because passive types "can be flipped for free" intuitively?
+
+The flow of the left-hand side expression must be sink or duplex.
+
+Neither side of the connect may have a property type.
+
+TODO: I think this is fraught. What if a property ends up in a bundle type?
```
mmaloney-sf, line 1472: @mikeurbach I'm looking for some help with this. I think the intention is that property types can never enter bundles. That seems to be suggestive in the grammar.

mwachs5:
I'm not sure how to review this PR since it seems like it has a lot of the same changes as the other or about circuit components. What is your plan with these 2 PRs...? Should you set the base of this one to be that Pr branch so that only the changes in the pr description are being reviewed?

mmaloney-sf:
@mwachs5 This is stacked on top of the changes from #135. Until that merges, it might be hard to pick out the changes.

mikeurbach:
> @mwachs5 This is stacked on top of the changes from https://github.com/chipsalliance/firrtl-spec/pull/135. Until that merges, it might be hard to pick out the changes.

GitHub isn't the best for stacked PRs, but you can make the base branch of this PR the branch from #135, if you want to call out the differences. That way we can review just the diff on this PR, and when #135 merges, GitHub will automatically update the base of this PR.

mmaloney-sf:
@mikeurbach Thanks for the clarification.

I've changed the branch target.

spec.md
```diff
@@ -1374,22 +1374,15 @@ to `{flip a: {b: UInt}}`{.firrtl}.
...
-A module body consists of a sequence of statements.
-Statements declare the circuit components and describe their connectivity.
-
-## Connects
+# Connections
 
 The components of a module can be connected together using connect statements.
 
-These connections can be both physical, as if the components were connected together with a wire,
+These connections are both physical, as if the components were connected together with a wire,
```
mwachs5, line 1381:
```suggestion
These connections can be both physical, as if the components were connected together with a wire,
```

if you just have properties there is nothing physical

spec.md
```diff
@@ -1398,28 +1391,94 @@ module MyModule :
   connect myoutput, myinput
 ```
 
-In order for a connection to be legal the following conditions must hold:
+In the syntax for a `connect` statement, the first argument is referred to as the "right-hand side".
```
mwachs5, line 1394: well, that's all very confusing. Note that `connect rhs, lhs` used to be written as `lhs <- rhs` where that made the most sense. Maybe a footnote for this fact to help readers understand why that is the case (or we rewrite all this??)

spec.md
```diff
@@ -1398,28 +1391,94 @@ module MyModule :
...
+
+The flow of an expression can be one of *source*, *sink*, or *duplex*.
+
+A source expression supplies a signal, and can be used to drive a circuit component.
+A sink expression can be driven by another expression.
+A duplex expression is simply an expression that is both a source and sink.
+
+The rules for the flow of an expression are as follows:
+
+If the expression is a reference, we look at the kind of the circuit component:
```
mwachs5, line 1420: What do you mean "the kind of circuit component"? You mean "the kind of circuit component that the reference is pointing to"?

spec.md
```diff
@@ -1398,28 +1391,94 @@ module MyModule :
...
+
+* Nodes are sources.
+* Wires and registers are duplex.
+* For ports, `input` ports are sources and `output` ports are sinks.
+* Submodule instances are sources.
+* Memories are sources.
```
mwachs5, line 1426:
```suggestion
* References to Nodes are sources.
* References to Wires and registers are duplex.
* For ports, `input` ports are sources and `output` ports are sinks.
* Submodule instances are sources.
* Memories are sources.
```


and so on? I really don't understand the references sentence above

spec.md
```diff
@@ -1398,28 +1391,94 @@ module MyModule :
...
+
+Here are a few comments to help with intuition:
+Nodes may only appear "on the left side" of a connect.
+Wires and registers may appear "on either side of a connect statement".
+Ports are always considered from the perspective of "inside the module".
+Moreover, input ports may only appear "on the left side" of a connect,
+while output ports may only appear "on the right side" of a connect.
```
mwachs5, line 1433: this is confusing because its not obvious how to refer to something on the OUTSIDE of the module. the same "port" is different when viewed from inside or outside the module you are talking about.

We may be well served to use industry standard terminology here. I think "ports" and "pins" are the names for the things viewed "from the inside" vs "from the outside" respectively.

spec.md
mmaloney-sf: I might just change it to a more meaningful name. The "target" and the ... something? I'll see if anyone has any ideas.

mwachs5: i realized after i wrote this that you had written some text below this that can help. I still think a LHS and RHS is ok with the rationale

darthscsi:
This really needed an approval before being merged.  It loses information and introduces errors.

mmaloney-sf:
My bad on this. I had merged this and a few others thinking I had the sign-off. But I guess this repository doesn't have that guard.

PR #139, opened 2023-11-03
Removed flow and replaced with source/sink and relative alignment

azidar:
This is my first attempt to remove the concept of "flow" from the FIRRTL spec. Instead, I've replaced it with source/sink expressions and relative alignment.

Feedback welcome.


mmaloney-sf:
Cool. Let me take some time to look this over.

As an initial thought: You're using the term "subcomponent". Could we wait until we have the term subcomponent formally defined before merging this in?

sequencer:
It’s glad to see firrtl is eventually removing the output and input with
Alignment!
I wonder is this a breaking API in chisel? Or we can seamlessly migrate out
current API to this?

spec.md
```diff
@@ -2887,25 +2911,17 @@ n must be non-negative and less than or equal to the bit width of e.
...
-The flow of a sub-index or sub-access expression is the flow of the vector-typed expression it indexes or accesses.
+An expression can either be used as either a source expression or a sink expression; if an expression is on the left-hand-side of a connect, it is a sink expression.
+All other expressions are source expressions.
 
-The flow of a sub-field expression depends upon the orientation of the field.
-If the field is not flipped, its flow is the same flow as the bundle-typed expression it selects its field from.
-If the field is flipped, then its flow is the reverse of the flow of the bundle-typed expression it selects its field from.
-The reverse of source is sink, and vice-versa.
-The reverse of duplex remains duplex.
+All hardware components can be referenced with a source expression.
```
seldridge, line 2919: The work that @mmaloney-sf has been doing on the grammar and general allows this to be simplified.

```suggestion
All expressions may be used as source expressions.
```

As written, it's not specified if primops, muxes, and other things are legal as sources. And they are!

spec.md
```diff
@@ -2887,25 +2911,17 @@ n must be non-negative and less than or equal to the bit width of e.
...
 
-The flow of all other expressions are source.
+With the following exceptions, all other hardware components/subcomponents can be referenced with a sink expression:
+  - input ports (or subcomponents which are input ports)
+  - data field of memory read ports
+  - rdata field of memory read-write ports
```
seldridge, line 2924: Additionally, no primitive operations may be used as sink expressions. Given @mmaloney-sf's new grammar, it should be safe to say that only `references` are legal sink expressions. Then introduce the restrictions on what references are not legal as sinks.

seldridge (commented):
High level:

1. Flipped/Aligned matters narrowly for connection semantics. Can these changes be sunk into the `Connection Algorithm` section and then used to rewrite that section? The motivation for this is that Flipped/Aligned are concepts which are only relevant to dealing with connects. We can keep the spec tighter by only introducing these concepts there.
2. Can the discussion of relative alignment be simplified to only discuss parent/child alignment once this is sunk? My reasoning here is that relative alignment of siblings I don't think is used and it would be terser to not mention it.

Pulling on (2), I think the essence of the concept is as follows:

A _connection_ sets two things: (a) a base alignment of "aligned" and (b) a base component with a base type specified by the type of either the source or sink expression (choice of source or sink doesn't matter as the types must be equivalent). All leaf sub-components of the source and sink are then connected. Leaf sub-components are connected source-to-sink if the relative alignment of the leaf sub-component and the base component/type are aligned. Leaf sub-components are connected sink-to-source if the relative alignment of the leaf sub-component and base component/type are flipped.

When described as such the notion of sibling alignment doesn't arise.

WDYT?

spec.md
azidar: Thanks!

spec.md
azidar: Excellent, I'll revisit this when I bump.

spec.md
```diff
@@ -648,34 +648,67 @@ Additionally, a field may optionally be declared with a *flipped* orientation.
...
 
-``` firrtl
-{real: {word: UInt<32>, valid: UInt<1>, flip ready: UInt<1>},
- imag: {word: UInt<32>, valid: UInt<1>, flip ready: UInt<1>}}
+Next, all parent/child relationships:
+Because the `flippedChild` field is `Flipped`, it changes its aligment relative to its parent.
 
-```
+- `p` is aligned w.r.t `p.alignedChild`
+- `p` is flipped w.r.t `p.flippedChild`
```
mwachs5, line 673: this is sort of weirdly written. I know it is a symmetrical relationship, but wouldn't it be more intuitive to write it as

`p.aligned child is aligned w.r.t. p`

spec.md
```diff
@@ -1911,7 +1942,9 @@ These statements are detailed below.
 ### Define
 
 Define statements are used to route references through the design, and may be used wherever is most convenient in terms of available identifiers -- their location is not significant other than scoping, and do not have last-connect semantics.
-Every sink-flow probe must be the target of exactly one of these statements.
+Every sink probe expression must be the target of exactly one of these statements.
+
+The define statement takes a sink static reference target and sets it to the specified reference, which must either be a compatible probe expression or static reference source.
```
mwachs5, line 1947:
```suggestion
The `define`{.firrtl} statement takes a sink static reference target and sets it to the specified reference, which must either be a compatible probe expression or static reference source.
```

spec.md
```diff
@@ -2887,25 +2916,18 @@ n must be non-negative and less than or equal to the bit width of e.
...
-If the field is not flipped, its flow is the same flow as the bundle-typed expression it selects its field from.
-If the field is flipped, then its flow is the reverse of the flow of the bundle-typed expression it selects its field from.
-The reverse of source is sink, and vice-versa.
-The reverse of duplex remains duplex.
+With the following exceptions, all other hardware components/subcomponents can be referenced with a sink expression:
 
-The flow of all other expressions are source.
+- input ports (or subcomponents which are input ports)
+- data field of memory read ports
+- rdata field of memory read-write ports
```
mwachs5, line 2930: if we use the nomenclature "ports" (as seen from within a module) and "pins" (as seen from outside a module) shouldn't an output pin not be allowed to be referenced with a sink expression?

PR #140, opened 2023-11-05
Fix grammar of `parameter` in `intmodule`

SpriteOvO:
`parameter`s in `intmodule` should have names, for example

```firrtl
intmodule MyIntModule :
  input in : UInt
  output out : UInt<8>
  intrinsic = testIntrinsic1
  parameter FORMAT = "xyz_timeout=%d\n"
  parameter DEFAULT = 0
  parameter WIDTH = 32
  parameter DEPTH = 32.42
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -3379,7 +3379,7 @@ decl_intmodule =
   "intmodule" , id , ":" , [ info ] , newline , indent ,
     { port , newline } ,
     "intrinsic" , "=" , id , newline ,
-    { "parameter" , "=" , ( int | string_dq ) , newline } ,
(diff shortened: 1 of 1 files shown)


seldridge (approved):
Thanks!

mwachs5:
please mark this as `patch` and add somethign to the version history for this since it does actually change the spec document

SpriteOvO:
@mwachs5 Sorry about that, I didn't checked the contribution guide. Since it's already merged, the only way to change the commit message is force-push, but that's not a good idea for the main branch. I'll pay attention to it next time.

PR #141, opened 2023-11-06
[nfc] Reorganize and Rename Appendices

mmaloney-sf:
- Move "Primitive Operations" section closer to the bottom.
- Rename a "FIRRTL Language Definition" to "Grammar".
- Rename "Details about Syntax" to "Notes on Syntax".

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -5,12 +5,14 @@ revisionHistory:
   # additions to the specification should append entries here.
   thisVersion:
     spec:
-      - Add public modules.  Change circuits to support any number of public
-        modules which may have disjoint instance hierarchies.
+      - Add public modules.
+        Change circuits to support any number of public modules which may have disjoint instance hierarchies.
+      - Rename a "FIRRTL Language Definition" to "Grammar".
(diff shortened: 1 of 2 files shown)


seldridge (approved):

seldridge:
Only thing that would be good for this type of thing is to add entries to the `revision-history.yaml` and to use commit/PR tags. E.g., `[nfc]`.

mmaloney-sf:
@seldridge Done.

I'm a little worried about spamming the revision section with all the changes I want to make during this reorganization effort. Any thoughts on how to mitigate that?

seldridge:
It's fine if there wind up being a lot of entries. Alternatively, the revision-history can just be amended. Obviously, don't edit anything in `oldVersions`. However, `thisVersion` can be updated as opposed to always adding new items.

mmaloney-sf:
I'm also moving the "Literals" section. It feels like it got separated from the herd. It's not important enough for its own top-level section, and I'm sticking it under the "Notes on Syntax" section. 

I'm also removing the heading on radix-specified integer literals. (The content is still there).

PR #143, opened 2023-11-07
[patch] Remove Statements section

mmaloney-sf:
The Statements section in the spec is not so much a listing of all the statements as it is a container for all non-type, non-expression topics in the language.

This PR doesn't change any copy. It only changes the organization of the sections under the Statements heading. 

Reading the commit messages should be sufficient to understand the entirety of what this PR does.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1348,6 +1348,59 @@ Invalidating a component with a bundle type recursively invalidates each sub-ele
 
 Components of reference and analog type are ignored, as are any reference or analog types within the component (as they cannot be connected to).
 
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -1727,7 +1727,7 @@ connect w.a, mux(c, y, x.a)
 connect w.b, x.b
 ```
 
-# Mem
+# Memory Instances
```
seldridge, line 1812: Nice.

seldridge (approved):

PR #144, opened 2023-11-07
[patch] Types Section Reorg

mmaloney-sf:
This PR re-orders a few sections under the Types section. The easiest way to review this is commit-by-commit.
Also, add `type` as a keyword for syntax highlighting.

No copy has been changed.

Some high-level points:

- Float property types up, since they are the four tier of the standard taxonomy (along with ground and aggregate and reference types).
- Move passive types up, since that feature is central and very stable.
- The probes section had a bunch of copy that isn't relevant to the type system and has been moved to the section of probes below.

All of this is prepwork for a heavy upcoming rewrite of the types section.

### DIFF
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -10,6 +10,7 @@
     </list>
     <list name="keywords">
       <item>input</item>
+      <item>type</item>
       <item>output</item>
       <item>defname</item>
       <item>parameter</item>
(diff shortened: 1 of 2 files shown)


include/firrtl.xml
```diff
@@ -10,6 +10,7 @@
     </list>
     <list name="keywords">
       <item>input</item>
+      <item>type</item>
```
seldridge, line 13: Good catch.

seldridge (approved):

PR #145, opened 2023-11-08
[patch] Rewrite Types Section

mmaloney-sf:
This PR is a moderate rewrite of the entire Types section.

This is best reviewed by building the PDF and reviewing paragraph by paragraph.

Here are some highlights:

- Uninferred types are mentioned in their own section. (Both width and reset inference).
- The section on zero-width integers is abbreviated. (It warrants mention, but doesn't need so much ink spilled)
- Analog type is scooched down.
- Small code formatting of FIRRTL (eg, spaces between braces in bundle types)
- Reword section on flips with bundle types.
- Rename "Reference Types" to "Probe Types"
- Move several examples of FIRRTL programs using probes out of the Probe Types section and into the probes section.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -593,7 +593,7 @@ SInt<32> ; 32-bit signed integer
 
 Both `UInt<0>`{.firrtl} and `SInt<0>`{.firrtl} are valid types.
 They are considered to have only a single value representing zero (written as either `UInt<0>(0)`{.firrtl} or `SInt<0>(0)`{.firrtl}).
-They are always zero extended, and thus, behave like zero when they are extended to a positive width.
+They behave like zero when they are extended to a positive width integer type.
 
 TODO: Talk about how zero-bit integers are lowered. And how they can be stripped out.
 TODO: There was an example here:
@@ -616,6 +616,8 @@ In FIRRTL, we have the option of using both synchronous or asynchronous resets.
 
 The synchronous reset type is simply a 1-bit unsigned integer: `UInt<1>`{.firrtl}.
 The asynchronous reset type is `AsyncReset`{.firrtl}.
+
+TODO: mention `regreset`.
 
 The reset types also have the uninferred form: `Reset`{.firrtl} (see [@sec:reset-inference]).
 
@@ -676,7 +678,7 @@ module Processor :
 ```
 
 This defines a module `Processor` with a single input port `enq` which enqueues data using a ready-valid interface.
-Because `enq` is a single port, we can connect to it with a single `connect` statement (see [@sec:connects]).
+Because `enq` is a single port, we can connect to it with a single `connect` statement (see [@sec:connections]).
 In the final hardware, however, while the `word` and `valid` fields point *into* the module, this port has a flipped field `ready` which points *out of* the module instead.
 
 The types of each field may be any type, including other aggregate types.
@@ -728,7 +730,6 @@ When analog signals are not given a concrete width, their widths are inferred ac
 ``` firrtl
 Analog<1>  ; 1-bit analog type
 Analog<32> ; 32-bit analog type
-Analog     ; analog type with inferred width
 ```
 
 ## Probe Types
@@ -742,11 +743,11 @@ Probe types are not synthesizable.
 There are two probe types, `Probe<T>`{.firrtl} is a read-only variant and `RWProbe<T>`{.firrtl} is a read-write variant.
 In both cases, `T`{.firrtl} must be a passive type (see [@sec:passive-types]).
 
-Both `Probe`{.firrtl} and `RWProbe`{.firrtl} may be read from using the `read` expression.
-`RWProbe`{.firrtl} may also be forced using the `force` and `force_initial` commands.
-However, when forcing is not needed, the `Probe`{.firrtl} allows more aggressive optimization.
-
 They are created using the `probe`{.firrtl} and `rwprobe`{.firrtl} expressions (see [@sec:probes]).
+
+Both `Probe`{.firrtl} and `RWProbe`{.firrtl} may be read from using the `read`{.firrtl} expression (see [@sec:reading-probe-references]).
+`RWProbe`{.firrtl} may also be forced using the `force`{.firrtl} and `force_initial`{.firrtl} commands (see [@sec:force-and-release]).
+However, when forcing is not needed, the `Probe`{.firrtl} allows more aggressive optimization.
 
 Probe references must always be able to be statically traced to their target, or to an external module's output reference.
 Reference-type ports are statically routed through the design using the `define`{.firrtl} statement.
@@ -757,9 +758,9 @@ When associated with an optional group, the reference type may only be driven fr
 Examples:
 
 ```firrtl
-Probe<UInt> ; readable reference to unsigned integer with inferred width
-RWProbe<{x: {y: UInt}}> ; readable and forceable reference to bundle
-Probe<UInt, A.B> ; readable reference associated with group A.B
+Probe<UInt<8>>                         ; readable probe to an 8-bit unsigned integer
+RWProbe<{ x : UInt<1>, y : UInt<1> }>  ; read-writable probe to bundle
+Probe<UInt, A.B>                       ; readable probe associated with group A.B
 ```
 
 For details of how to read and write through probe types, see [@sec:reading-probe-references;@sec:force-and-release].
@@ -773,7 +774,7 @@ The following example demonstrates some legal and illegal expressions:
 
 ```firrtl
 module NoSubAccessesWithProbes :
-  input x : {a : Probe<UInt[2]>, b : UInt}[3]
+  input x : { a : Probe<UInt[2]>, b : UInt }[3]
   input i : UInt
   input c : const UInt
   output p : Probe<UInt>
@@ -943,6 +944,12 @@ For multiplexing aggregate-typed expressions, the resulting widths of each leaf 
 The width of each primitive operation is detailed in [@sec:primitive-operations].
 
 The width of constant integer expressions is detailed in their respective sections.
+
+```firrtl
+UInt
+SInt
+Analog
+```
 
 ### Reset Inference
 
--- a/spec.md
+++ b/spec.md
@@ -934,30 +934,6 @@ Type aliases do not share the same namespace as modules; hence it is allowed for
 TODO: Is there a rationale for this design choice?
 TODO: Namespaces
 
-## Type Equivalence
-
-The type equivalence relation is used to determine whether a connection between two components is legal.
-See [@sec:connections] for further details about connect statements.
-
-An unsigned integer type is always equivalent to another unsigned integer type regardless of bit width, and is not equivalent to any other type.
-Similarly, a signed integer type is always equivalent to another signed integer type regardless of bit width, and is not equivalent to any other type.
-
-Clock types are equivalent to clock types, and are not equivalent to any other type.
-
-An uninferred `Reset`{.firrtl} can be connected to another `Reset`{.firrtl}, `UInt`{.firrtl} of unknown width, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}.
-It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
-
-The `AsyncReset`{.firrtl} type can be connected to another `AsyncReset`{.firrtl} or to a `Reset`{.firrtl}.
-
-Two enumeration types are equivalent if both have the same number of variants, and both the enumerations' i'th variants have matching names and equivalent types.
-
-Two vector types are equivalent if they have the same length, and if their element types are equivalent.
-
-Two bundle types are equivalent if they have the same number of fields, and both the bundles' i'th fields have matching names and orientations, as well as equivalent types.
-Consequently, `{a:UInt, b:UInt}`{.firrtl} is not equivalent to `{b:UInt, a:UInt}`{.firrtl}, and `{a: {flip b:UInt}}`{.firrtl} is not equivalent to `{flip a: {b: UInt}}`{.firrtl}.
-
-Two property types are equivalent if they are the same concrete property type.
-
 ## Type Inference
 
 FIRRTL has support for limited type inference.
@@ -1094,6 +1070,32 @@ Wires and registers may appear "on either side of a connect statement".
 Ports are always considered from the perspective of "inside the module".
 Moreover, input ports may only appear "on the left side" of a connect, while output ports may only appear "on the right side" of a connect.
 Finally, while submodules instances and memories are strictly sources, they interact with the sub-field rule below, allowing connections to their input ports
+
+## Type Equivalence
+
+TODO: Replace this with a relation like "S 'coerces to' T".
+
+The type equivalence relation determines whether one values of one type can be coerced into values of another.
+
+An unsigned integer type is always equivalent to another unsigned integer type regardless of bit width, and is not equivalent to any other type.
+Similarly, a signed integer type is always equivalent to another signed integer type regardless of bit width, and is not equivalent to any other type.
+
+Clock types are equivalent to clock types, and are not equivalent to any other type.
+
+An uninferred `Reset`{.firrtl} can be connected to another `Reset`{.firrtl}, `UInt`{.firrtl} of unknown width, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}.
+It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
+
+The `AsyncReset`{.firrtl} type can be connected to another `AsyncReset`{.firrtl} or to a `Reset`{.firrtl}.
+
+Two enumeration types are equivalent if both have the same number of variants, and both the enumerations' i'th variants have matching names and equivalent types.
+
+Two vector types are equivalent if they have the same length, and if their element types are equivalent.
+
+Two bundle types are equivalent if they have the same number of fields, and both the bundles' i'th fields have matching names and orientations, as well as equivalent types.
+Consequently, `{a:UInt, b:UInt}`{.firrtl} is not equivalent to `{b:UInt, a:UInt}`{.firrtl}, and `{a: {flip b:UInt}}`{.firrtl} is not equivalent to `{flip a: {b: UInt}}`{.firrtl}.
+
+Two property types are equivalent if they are the same concrete property type.
+
 
 ## The Connect Statement
 
--- a/spec.md
+++ b/spec.md
@@ -814,9 +814,10 @@ Property types are legal in the following constructs:
 
 * Port declarations on modules and external modules
 
-### Integer Property Types
+### Integer Type
 
-Integer property types represent arbitrary precision signed integer values.
+The `Integer` type represents an numeric property.
+It can represent arbitrary-precision signed integer values.
 
 ``` firrtl
 module Example:
--- a/spec.md
+++ b/spec.md
@@ -837,12 +837,12 @@ All integer literals are `const`{.firrtl}.
 For example, `UInt<8>(42)`{.firrtl} has type `const UInt<8>`{.firrtl}.
 
 Ports can have a `const`{.firrtl} type, and thus, a module may receive constant values from its parent module.
-This may even happen in `public`{.firrtl} modules, and so the value of a `const`{.firrtl} type need not be known statically (see TODO).
+This may even happen in `public`{.firrtl} modules, and so the value of a `const`{.firrtl} type need not be known statically (see [@sec:public-modules]).
 
-Typically, primitive operations will result in a `const`{.firrtl} type whenever each of its inputs are `const`{.firrtl} (see prim ops TODO).
+Typically, primitive operations will result in a `const`{.firrtl} type whenever each of its inputs are `const`{.firrtl} (see [@sec:primitive-operations]).
 For example, `add(x, y)`{.firrtl} will be `const`{.firrtl} if both `x`{.firrtl} and `y`{.firrtl} are `const`{.firrtl}.
 
-The resulting type of a multiplexer expression, `mux(s, a, b)`{.firrtl}, will be `const`{.firrtl} if all of `s`{.firrtl}, `a`{.firrtl}, and `b`{.firrtl} are `const`{.firrtl} (see TODO).
+The resulting type of a multiplexer expression, `mux(s, a, b)`{.firrtl}, will be `const`{.firrtl} if all of `s`{.firrtl}, `a`{.firrtl}, and `b`{.firrtl} are `const`{.firrtl} (see [@sec:multiplexers]).
 
 An expression of type `const T`{.firrtl} is implicitly upcast to type `T`{.firrtl} whenever it would be required to make a primitive operation, mux expression, or connect statement typecheck.
 
@@ -890,17 +890,8 @@ Type aliases have structural identity.
 In other words, when we compare two types, we expand all type aliases recursively until we are left with type expressions that have no aliases.
 For instance, in the above example, we can connect `in.w`{.firrtl} to `w`{.firrtl} since their types, `WordType` and `AnotherWordType` respectively, both expand to `UInt<32>`{.firrtl}.
 
-TODO: Type equivalence should mention this.
-TODO: Can type aliases be recursive? Are they in scope of each other?
-
-TODO: Mention that `type` is a top-level decl.
-TODO: How global? Global to the `circuit`?
-
 The `type`{.firrtl} declaration is globally defined and all named types exist in the same namespace and thus must all have a unique name.
 Type aliases do not share the same namespace as modules; hence it is allowed for type aliases to conflict with module names.
-
-TODO: Is there a rationale for this design choice?
-TODO: Namespaces
 
 ## Type Inference
 
@@ -911,7 +902,7 @@ Width inference and reset inference.
 ### Width Inference
 
 Normally, the three integer ground types are written with explicit bit widths.
-The bit widths are given in angel brackets, such as the 8 in `UInt<8>`{.firrtl}.
+The bit widths are given in angel brackets, such as the `8`{.firrtl} in `UInt<8>`{.firrtl}.
 This is called the **uninferred** variant of the type.
 
 FIRRTL also supports an **inferred** variant of these types.
@@ -926,16 +917,15 @@ Analog
 When an inferred variant of an integer ground type is used, FIRRTL will calculate the minimum width needed for it.
 It is an error if the minimum size cannot be calculated.
 
-The width of a mux expression is the maximum of its two corresponding input widths.
-For multiplexing aggregate-typed expressions, the resulting widths of each leaf sub-element is the maximum of its corresponding two input leaf sub-element widths.
-TODO: Are these anywhere near complete enough to bother mentioning them? It seems better to just remove this paragraph until we have the exact rules.
+For example, the width of a multiplexer expression is the maximum of its two corresponding input widths.
 
 The width of each primitive operation is detailed in [@sec:primitive-operations].
-TODO: What "width"? The minimum widths?
 
 ### Reset Inference
 
-The following example shows an uninferred reset that will get inferred to a synchronous reset.
+The uninferred `Reset`{.firrtl} type will be inferred to either a synchronous reset `UInt<1>`{.firrtl} or to an asynchronous reset `AsyncReset`{.firrtl}.
+
+The following example shows an inferred reset that will get inferred to a synchronous reset.
 
 ``` firrtl
 input a : UInt<1>
--- a/spec.md
+++ b/spec.md
@@ -763,7 +763,7 @@ Probe<UInt<8>, A.B>     ; A.B is an optional group
 RWProbe<UInt<8>, A.B>
 ```
 
-Probes are generally lowered to hierarchical expressions in Verilog.
+Probes are generally lowered to hierarchical names in Verilog.
 For details, see the FIRRTL ABI Specification.
 
 ## Property Types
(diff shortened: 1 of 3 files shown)


spec.md
```diff
@@ -547,212 +547,129 @@ See [@sec:mem] for more details.
...
-
-``` firrtl
-Clock
-```
+Both `UInt<0>`{.firrtl} and `SInt<0>`{.firrtl} are valid types.
+They are considered to have only a single value representing zero (written as either `UInt<0>(0)`{.firrtl} or `SInt<0>(0)`{.firrtl}).
+They are always zero extended, and thus, behave like zero when they are extended to a positive width.
 
-### Reset Type
+TODO: Talk about how zero-bit integers are lowered. And how they can be stripped out.
```
seldridge, line 573: The PR should be careful to avoid putting ABI information into the spec. E.g., this discussion could go in the direction of what happens to zero-width ports. However, that is part of the ABI guarantees of the port lowering ABIs.

There is something needed here, though!

spec.md
```diff
@@ -547,212 +547,129 @@ See [@sec:mem] for more details.
...
-input a : UInt<1>
-wire reset : UInt<1>
-connect reset, a
-```
+### Clock Type
 
-Asynchronous resets used in registers will be mapped to a hardware description language representation for asynchronous resets.
+Clocks require special physical considerations in hardware.
+FIRRTL defines the `Clock` type to track clocks throughout a design.
+All registers are linked to a clock (see [@sec:conditionals]).
```
seldridge, line 582: Is `conditionals` the correct section link?

spec.md
```diff
@@ -547,212 +547,129 @@ See [@sec:mem] for more details.
...
 ``` firrtl
 UInt<16>[10]
 ```
 
-The next example specifies a ten element vector of unsigned integers of omitted but identical bit widths.
-
-``` firrtl
-UInt[10]
-```
+Note that the element type of a vector can be any type, including another aggregate types.
```
seldridge, line 611:
```suggestion
Note that the element type of a vector can be any type, including another aggregate type.
```

spec.md
```diff
@@ -547,212 +547,129 @@ See [@sec:mem] for more details.
...
+
 ### Bundle Types
 
-A bundle type is used to express a collection of nested and named types.
-All fields in a bundle type must have a given name, and type.
-All names must be legal identifiers.
+A bundle type is used to represent a collection of values.
+They can also be used to facilitate bidirectional connections between circuit components.
+
+A bundle type consists of a set of fields.
```
seldridge, line 627: Mega-nit: I grok this is an intuitional sentence. However, `set` makes me ask if this is a mathematical set and whether it is ordered (it is from the port-lowering ABI). Is there a better way of stating this without making it sound like a vector? Collection, sequence, e.g.?

spec.md
```diff
@@ -547,212 +547,129 @@ See [@sec:mem] for more details.
...
 ```
 
-In a connection to the `a`{.firrtl} port, the data carried by the `word`{.firrtl} and `valid`{.firrtl} sub-fields will flow out of the module, while data carried by the `ready`{.firrtl} sub-field will flow into the module.
-More details about how the bundle field orientation affects connections are explained in [@sec:connects].
+This defines a module `Processor` with a single input port `enq` which enqueues data using a ready-valid interface.
+Because `enq` is a single port, we can connect to it with a single `connect` statement (see [@sec:connects]).
+In the final hardware, however, while the `word` and `valid` fields point *into* the module, this port has a flipped field `ready` which points *out of* the module instead.
 
-As in the case of vector types, a bundle field may be declared with any type, including other aggregate types.
+The types of each field may be any type, including other aggregate types.
```
seldridge, line 657: This is a bit of a hard transition from the previous paragraph. I thought this was continuing the example above.

Perhaps this would be better located at the start of the sub-section.

spec.md
```diff
@@ -547,212 +547,129 @@ See [@sec:mem] for more details.
...
-```
+### Enumeration Types
 
-In a connection to `myport`{.firrtl}, the `a`{.firrtl} sub-field flows out of the module.
-The `c`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field flows into the module, and the `d`{.firrtl} sub-field contained in the `b`{.firrtl} sub-field flows out of the module.
+Enumerations are disjoint union types.
 
-### Enumeration Types
+Each enumeration consists of a set of variants.
+Each variant is named with a tag and each has a passive type (see [@sec:passive-types]).
```
seldridge, line 670: Hemmingway speaking from the grave:

```suggestion
Each variant is named with a tag.
Each variants has a passive type (see [@sec:passive-types]).
```

spec.md
```diff
@@ -789,48 +706,29 @@ Analog<32> ; 32-bit analog type
...
+However, when forcing is not needed, the `Probe`{.firrtl} allows more aggressive optimization.
 
-Reference-type ports are statically routed through the design using the `define`{.firrtl} statement.
-
-There are two reference types, `Probe`{.firrtl} and `RWProbe`{.firrtl}, described below.
-These are used for indirect access to probes of the data underlying circuit constructs they originate from, captured using `probe`{.firrtl} expressions (see [@sec:probes]).
-
-`Probe`{.firrtl} types are read-only, and `RWProbe`{.firrtl} may be used with `force`{.firrtl} and related statements.
-Prefer the former as much as possible, as read-only probes impose fewer limitations and are more amenable to optimization.
+They are created using the `probe`{.firrtl} and `rwprobe`{.firrtl} expressions (see [@sec:probes]).
```
seldridge, line 724:
```suggestion
`Probe`{.firrtl} and `RWProbe`{.firrtl} are created using the `probe`{.firrtl} and `rwprobe`{.firrtl} expressions, respectively (see [@sec:probes]).
```

spec.md
```diff
@@ -1040,6 +919,83 @@ The width of each primitive operation is detailed in [@sec:primitive-operations]
...
+
+See [@sec:primitive-operations] for more details on casting.
+
+### Probes and Type Inference
+
+Probe types do participate in global width and reset inference, but only in the direction of the reference itself (no inference in the other direction, even with force statements).
+Both inner types of the references used in a `define`{.firrtl} statement must be identical or the same type with the destination uninferred (this is checked recursively).
+Additionally, any contained reset type is similarly only inferred in the direction of the reference, even if it eventually reaches a known reset type.
+
+In the following example, the FIRRTL compiler will produce an error constrasted with inferring the input port as `AsyncReset`{.firrtl} if a direct connection was used:
```
seldridge, line 977: I would avoid discussions of a FIRRTL compiler and opt instead for language around legal and illegal FIRRTL circuits. Also, while there is de facto one ("the") FIRRTL compiler, there may be many.

spec.md
```diff
@@ -547,212 +547,129 @@ See [@sec:mem] for more details.
...
-Aggregate types and reference types are composed of one or more aggregate or ground types.
-Reference types may not contain other reference types.
-Property types represent information about the circuit that is not hardware.
+FIRRTL has four classes of types: **ground** types, **aggregate** types, **probe** types, and **property** types.
 
 ## Ground Types
 
-There are five classes of ground types in FIRRTL: unsigned integer types, signed integer types, a clock type, reset types, and analog types.
+Ground types are the most fundamental types.
+They include integer types, clocks, and resets.
```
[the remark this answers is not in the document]
mmaloney-sf: > Ground types are types which are not composed of other types.
> They are the integer types, clocks, and resets.

Reworded to this. 

I made a note of the ground type terminology. We can revisit this.

spec.md
mmaloney-sf: Agreed. There was a whole paragraph on this, and I stripped it out. 

I'll put it in my notes to add to the ABI doc and remove this TODO.

spec.md
mmaloney-sf: > A bundle type consists of multiple fields.

Changed.

spec.md
```diff
@@ -547,212 +547,134 @@ See [@sec:mem] for more details.
...
-AsyncReset
+```firrtl
+UInt<10> ; a 10-bit unsigned integer
+SInt<32> ; 32-bit signed integer
 ```
 
-Synchronous resets used in registers will be mapped to a hardware description language representation for synchronous resets.
+Both `UInt<0>`{.firrtl} and `SInt<0>`{.firrtl} are valid types.
+They are considered to have only a single value representing zero (written as either `UInt<0>(0)`{.firrtl} or `SInt<0>(0)`{.firrtl}).
+They behave like zero when they are extended to a positive width integer type.
```
seldridge, line 571: Using `pad` with a reference to the `pad` primop may be more accurate than saying `extend` which could mean either zero-extend or sign-extend (and FIRRTL doesn't have both operations, just one operation that is sensitive to the type).

As commentary (either in the PR or future), interpreting sign-extension of a zero-width signed integer as adding leading zeros was something controversial when I tried to land this on upstream MLIR. There were discussions about whether this should _actually add ones_. I.e., the value of a zero-width signed integer is actually `-1`. The end result was that this was undefined and MLIR will error if you try to sign extend a zero-width signed integer.

This is again in the Bruce Hornsby bucket and it's fine that FIRRTL does it this way. A reader with experience (or a mathematical background) may do a double-take on these sentences.

Did you have a similar reaction given your math background?

spec.md
```diff
@@ -547,212 +547,134 @@ See [@sec:mem] for more details.
 
 # Types
 
-FIRRTL has four classes of types: _ground_ types, _aggregate_ types, _reference_ types, and _property_ types.
-Ground types are fundamental and are not composed of other types.
-Aggregate types and reference types are composed of one or more aggregate or ground types.
-Reference types may not contain other reference types.
-Property types represent information about the circuit that is not hardware.
+FIRRTL has four classes of types: **ground** types, **aggregate** types, **probe** types, and **property** types.
```
mwachs5, line 550: you seem to have removed analog from the ground types list below, but not added them to this list... where do they go in this classification?

mmaloney-sf: I need to ask some questions about analog types to decide where they should go.

Related to that is what "ground" type even means. The way we classify types is really lumpy, and I want to feel things out a bit.

mmaloney-sf:
@darthscsi I rewrote the Constant Types section as best as I could.

Being such a massively cross-cutting feature, it's hard to spec out in just one section. I think the rewritten version touches on all points from the original:

- include a (tentative) rationale for the feature
- remove the example of a bundle with a nested `const` -- it's theoretically interesting, but not really speced out, so let's not draw attention to it
- mention that `const` doesn't mean `static` -- these things might be unknown because of `input` ports
- explain the rules everyone will agree one (literals are `const`, prim ops of `const` are `const`, muxes of `const` are `const`).
- Gesture at how `const T` can be implicitly coerced/cast to `T`.
- Gesture at the connect rules for `const` (source must be `const`, any conditionals must have constant conditions)
-  Reword "indexing" into aggregates in terms of subcomponents (I haven't written the subcomponents section yet, but bear with me please!)
- Moved "A note on implementation" to the ABI doc.

Thank you @seldridge for letting me bounce my thoughts off of you.

mmaloney-sf:
One big addition at the last minute:

I define a **Connectable Type** as a type which may appear as the type of an expression which may participate in a `connect` statement.

Connectable types appear internally in the FIRRTL MLIR dialect. They affect how components are split up when lowering to Verilog. 

Currently, I only rewrote the enumerations section in terms of these. (Namely, the types of variants must be connectable and passive). But as I get to the section on Connections, we can apply this definition cleanly there.

This definition will also help us untangle some thorny issues in FIRRTL, which add a disproportionate amount of complexity to the spec:

- what it means for non-hardware to be packaged in aggregates (eg, bundles of probes or properties), and
- how we should think of the `Analog` type

mmaloney-sf:
I forgot -- I'm a bit stuck on the section on Probes and Type Inference.

@dtzSiFive Can you advise me on how to handle this section?

spec.md
mwachs5: sorry how was this resolved?

spec.md
```diff
@@ -792,133 +727,44 @@ In the following example, all variants have the type `UInt<0>`{.firrtl}.
...
-
-Using probe-type ports, modules may expose internals for reading and forcing without routing wires out of the design.
+## Probe Types
 
-This is often useful for testing and verification, where probe types allow reads of the entities to be explicitly exported without hard-coding their place in the design.
-Instead, by using probe-type references, a testbench module may express accesses to the internals which will resolve to the appropriate target language construct by the compiler (e.g., hierarchical reference).
+Probe types expose and provide access to circuit components contained inside a module.
 
-Reference ports are not expected to be synthesizable or representable in the target language and are omitted in the compiled design; they only exist at the FIRRTL level.
+Probe types are useful for testing and verification, since they allow a design to read to and write from circuit components without knowing the hierarchical path to that component in the design.
```
mwachs5, line 734: this seems like a possible mischaracterization... in firrtl there is no concept of hierarchical path to a component in a design (except in targets...), in verilog there is. We could have just implemented probes like that in the firrtl spec, the advantage isn't really that you do or do not know the hierarchical path to the component in the design.  I'm not sure that is what makes them useful for testing and verification, they are useful in verilog because there is a sense that the verilog hierarchy is "real" and XMRs are "not real" -- you can add XMRs without adding any "real" ports. Would love @seldridge, @darthscsi  to weigh in on this rewrite.

spec.md
```diff
@@ -792,133 +727,44 @@ In the following example, all variants have the type `UInt<0>`{.firrtl}.
...
-There are two probe types: `Probe`{.firrtl} and `RWProbe`{.firrtl}.
-`RWProbe`{.firrtl} is a `Probe`{.firrtl} type, but not the other way around.
-
-Probe types are parametric over the type of data that they refer to, which is always passive (as defined in [@sec:passive-types]) even when the probed target is not (see [@sec:probes-and-passive-types]).
-Probe types cannot contain reference types.
-
-Conceptually probe types are single-direction views of the probed data-flow point.
-They are references to the data accessed with the probe expression generating the reference.
+There are two probe types, `Probe<T>`{.firrtl} is a read-only variant and `RWProbe<T>`{.firrtl} is a read-write variant.
+In both cases, `T`{.firrtl} must be a ground type (see [@sec:ground-types]).
```
mwachs5, line 737: is this an intentional change to the spec? Because the example you have deleted is not a ground type. If you are trying to make a functional change to the spec that probes have to be ground types, please make a different PR for that.

mwachs5 (dismissed):
it seems this PR is (unintentionally?) making probes more restricted to be only on ground types. If that is the intention please break that change to its own, non-patch PR.

spec.md
```diff
@@ -572,217 +572,152 @@ See [@sec:mem] for more details.
...
-Aggregate types and reference types are composed of one or more aggregate or ground types.
-Reference types may not contain other reference types.
-Property types represent information about the circuit that is not hardware.
+FIRRTL has four classes of types: **ground** types, **aggregate** types, **probe** types, and **property** types.
 
 ## Ground Types
 
-There are five classes of ground types in FIRRTL: unsigned integer types, signed integer types, a clock type, reset types, and analog types.
+Ground types are types which are not composed of other types.
+They are the integer types, clocks, and resets.
```
seldridge, line 580: Also, analog?

spec.md
```diff
@@ -572,217 +572,152 @@ See [@sec:mem] for more details.
...
-``` firrtl
-input clock : Clock
-input reset : AsyncReset
-input x : UInt<8>
-regreset y : UInt<8>, clock, reset, UInt(123)
-; ...
-```
+The analog type specifies that a wire or port can be attached to multiple drivers.
+`Analog`{.firrtl} cannot be used as part of the type of a node or register, nor can it be used as part of the datatype of a memory.
+In this respect, it is similar to how `inout`{.firrtl} ports are used in Verilog, and FIRRTL analog signals are often used to interface with external Verilog or VHDL IP.
```
seldridge, line 623: This is fine as commentary here. However, there are a few things missing/unclear: (1) How is this represented in the ABI? (There is a missing ABI section on Analog lowering.) (2) Given the ABI, Analog does lower to `inout`, therefore, is it appropriate to say `Analog` is "similar to how inout ports are used in Verilog"?

spec.md
```diff
@@ -572,217 +572,152 @@ See [@sec:mem] for more details.
...
 
+Vectors with length 0 are permitted.
+For details on how 0-length vectors are lowered, see the FIRRTL ABI Specification.
+
 ### Bundle Types
 
-A bundle type is used to express a collection of nested and named types.
-All fields in a bundle type must have a given name, and type.
-All names must be legal identifiers.
+A bundle type is used to represent a collection of values.
```
seldridge, line 668: I'd opt for a "collection of types" as opposed to "collection of values".

spec.md
```diff
@@ -572,217 +572,152 @@ See [@sec:mem] for more details.
...
-UInt[10]
-```
-
-Note that any type, including other aggregate types, may be used as the element type of the vector.
-The following example specifies a twenty element vector, each of which is a ten element vector of 16-bit unsigned integers.
+Note that the element type of a vector can be any type, including another aggregate type.
+The following example specifies a 20-element vector, each of which is a 10-element vector of 16-bit unsigned integers.
 
-``` firrtl
+```firrtl
```
seldridge, line 659: Mega-nit: FWIW, the formatter is going to add a space after triple-backtick. It's fine to just handle this with a format PR.

spec.md
```diff
@@ -572,217 +572,152 @@ See [@sec:mem] for more details.
...
-When calculating the final direction of data flow, the orientation of a field is applied recursively to all nested types in the field.
-As an example, consider the following module port declared with a bundle type containing a nested bundle type.
+As an example of how `flip`{.firrtl} works in context, consider a module declared like this:
 
-``` firrtl
-output myport: {a: UInt, flip b: {c: UInt, flip d: UInt}}
+```firrtl
+module Processor :
+  input enq : { word : UInt<32>, valid : UInt<1>, flip ready : UInt<1> }
+  ; ...
```
seldridge, line 702: Nit: the ellipsis is unnecessary. You could make it clearly irrelevant by making `Processor` an `extmodule`.

spec.md
seldridge: A major use case of probes is that you can use them without knowing the hierarchical path which makes them work for separate compilation where annotations are intractable. They are highly amenable to compiler implementations because they work with hierarchy manipulations and optimizations (e.g., constant propagation) whereas annotations were a hug pain here.

That said, drawing a comparison to something which doesn't exist in FIRRTL (a hierarchical path) may not be the right way to explain this. I.e., they are useful for verification because they let you read from circuit components without creating ports. (This is restating Megan's penultimate sentence.)

spec.md
seldridge: I thought this is how these are being implemented from Chisel now. Probes of aggregates don't work well with the the Port Lowering ABI v1 (i.e., `LowerTypes`) on public modules. Hence, this is a restriction that I think is already in place.

We're at a `[major]` spec boundary given that the spec already has public modules. So, I'm fine if this is just a `[major]` change.

spec.md
```diff
@@ -792,133 +727,44 @@ In the following example, all variants have the type `UInt<0>`{.firrtl}.
...
-```firrtl
-circuit :
-  public module ResetInferGood :
-    input in : Reset
-    output out : Reset
-    output out2 : AsyncReset
-    connect out, read(probe(in))
-    connect out2, in
-```
+Probes are generally lowered to hierarchical expressions in Verilog.
```
seldridge, line 766: Nit:
```suggestion
Probes are generally lowered to hierarchical names in Verilog.
```

spec.md
```diff
@@ -937,140 +783,208 @@ Property types are legal in the following constructs:
...
+This comes in two flavors:
+Width inference and reset inference.
 
-An unsigned integer type is always equivalent to another unsigned integer type regardless of bit width, and is not equivalent to any other type.
-Similarly, a signed integer type is always equivalent to another signed integer type regardless of bit width, and is not equivalent to any other type.
+### Width Inference
 
-Clock types are equivalent to clock types, and are not equivalent to any other type.
+Normally, the three integer ground types are written with explicit bit widths.
+The bit widths are given in angel brackets, such as the `8`{.firrtl} in `UInt<8>`{.firrtl}.
```
seldridge, line 913:
```suggestion
The bit widths are given in angle brackets, such as the `8`{.firrtl} in `UInt<8>`{.firrtl}.
```

spec.md
```diff
@@ -937,140 +783,208 @@ Property types are legal in the following constructs:
...
+This is called the **uninferred** variant of the type.
 
-An uninferred `Reset`{.firrtl} can be connected to another `Reset`{.firrtl}, `UInt`{.firrtl} of unknown width, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}.
-It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
+FIRRTL also supports an **inferred** variant of these types.
+They are written as follows:
```
seldridge, line 917: The copy here introduces `uninferred` for the first time. However, `inferred` was introduced in the discussion of types. Does it make sense to introduce `uninferred` earlier as well?

spec.md
```diff
@@ -937,140 +783,208 @@ Property types are legal in the following constructs:
...
+For example, the width of a multiplexer expression is the maximum of its two corresponding input widths.
 
-Two bundle types are equivalent if they have the same number of fields, and both the bundles' i'th fields have matching names and orientations, as well as equivalent types.
-Consequently, `{a:UInt, b:UInt}`{.firrtl} is not equivalent to `{b:UInt, a:UInt}`{.firrtl}, and `{a: {flip b:UInt}}`{.firrtl} is not equivalent to `{flip a: {b: UInt}}`{.firrtl}.
+The width of each primitive operation is detailed in [@sec:primitive-operations].
 
-Two property types are equivalent if they are the same concrete property type.
+### Reset Inference
+
+The uninferred `Reset`{.firrtl} type will be inferred to either a synchronous reset `UInt<1>`{.firrtl} or to an asynchronous reset `AsyncReset`{.firrtl}.
```
seldridge, line 934: Supreme nit:
```suggestion
The uninferred `Reset`{.firrtl} type will be inferred to either a synchronous reset, `UInt<1>`{.firrtl}, or to an asynchronous reset, `AsyncReset`{.firrtl}.
```

spec.md
```diff
@@ -937,140 +783,208 @@ Property types are legal in the following constructs:
...
+``` firrtl
+input a : UInt<1>
+wire reset : Reset
+connect reset, a
+```
 
-For all circuit components declared with unspecified widths,
-the FIRRTL compiler will infer the minimum possible width that maintains the legality of all its incoming connections.
-If a circuit component has no incoming connections, and the width is unspecified, then an error is thrown to indicate that the width could not be inferred.
+After reset inference, `reset`{.firrtl} is inferred to the synchronous `UInt<1>`{.firrtl} type:
```
seldridge, line 944: This sentence and the previous one seem to restate the same point. I think this is actually incorrectly repeated twice.

spec.md
```diff
@@ -937,140 +783,208 @@ Property types are legal in the following constructs:
...
+The following example demonstrates usage of an asynchronous reset.
 
-The width of each primitive operation is detailed in [@sec:primitive-operations].
+``` firrtl
+input clock : Clock
+input reset : AsyncReset
+input x : UInt<8>
+regreset y : UInt<8>, clock, reset, UInt(123)
+; ...
+```
```
seldridge, line 960: This example seems like a non-sequitor?

spec.md
```diff
@@ -937,140 +783,208 @@ Property types are legal in the following constructs:
...
+
+Inference rules are as follows:
+
+1. An uninferred reset driven by and/or driving only asynchronous resets will be inferred as asynchronous reset.
+1. An uninferred reset driven by and/or driving both asynchronous and synchronous resets is an error.
+1. Otherwise, the reset is inferred as synchronous (i.e. the uninferred reset is only invalidated or is driven by or drives only synchronous resets).
```
seldridge, line 966: Mega-nit: 
```suggestion
1. Otherwise, the reset is inferred as synchronous. (I.e., the uninferred reset is only invalidated or is driven by or drives only synchronous resets.)
```

spec.md
```diff
@@ -937,140 +783,208 @@ Property types are legal in the following constructs:
...
+
+`Reset`{.firrtl}s, whether synchronous or asynchronous, can be cast to other types.
+Casting between reset types is also legal:
```
seldridge, line 969: Does the PR want to call out what are the casts in the example below?

spec.md
```diff
@@ -1096,6 +1010,30 @@ Ports are always considered from the perspective of "inside the module".
 Moreover, input ports may only appear "on the left side" of a connect, while output ports may only appear "on the right side" of a connect.
 Finally, while submodules instances and memories are strictly sources, they interact with the sub-field rule below, allowing connections to their input ports
 
+## Type Equivalence
+
+The type equivalence relation determines whether one values of one type can be coerced into values of another.
```
seldridge, line 1015: I'm not sure if type coercion is the right concept here. Type equivalence is really this concept that is used narrowly to check the legality of statements or expressions. Put differently, the type equivalence relation says that `UInt<8>` is equal to `UInt`. This really only matters for connects to allow:

```
wire a: UInt
wire b: UInt<8>
connect b, a
```

However, the _type_ of that connect is still `(UInt<8>, UInt) -> ()`. I.e., the type of `a` is not coerced in any way. `a` is still an uninferred `UInt`.

Put differently again, this entire section is basically, "How do I write a verifier for connect ops?" as opposed to anything type theoretic.

It might be possible to reframe all this as "connects require exact type equivalence and uninferred is coereced to its inferred version". However, that is practically not what is happening.

spec.md
```diff
@@ -1096,6 +1010,30 @@ Ports are always considered from the perspective of "inside the module".
...
+
+An unsigned integer type is always equivalent to another unsigned integer type regardless of bit width, and is not equivalent to any other type.
+Similarly, a signed integer type is always equivalent to another signed integer type regardless of bit width, and is not equivalent to any other type.
+
+Clock types are equivalent to clock types, and are not equivalent to any other type.
+
+An uninferred `Reset`{.firrtl} can be connected to another `Reset`{.firrtl}, `UInt`{.firrtl} of unknown width, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}.
+It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
```
seldridge, line 1023: This sentence doesn't seem appropriate for type equivalence.

spec.md
seldridge: Maybe this entire section of type equivalence should really be moved under connects and called "connect legality" or "connect verification". 🤔

spec.md
mmaloney-sf: `Analog` was simply moved back to the Ground Types subsection.

spec.md
mmaloney-sf: This is the kind of thing I'm looking to flush out with this spec work.

I'm fine with making this a `[major]` change. But let's be sure not to release a new version until the dust has all settled.

spec.md
mmaloney-sf: Wait what? Have I been writing markdown codeblocks wrong my whole life?

I'll submit a separate PR to run the formatter.

spec.md
mmaloney-sf: Fixed.

spec.md
mmaloney-sf: `{👼}`

spec.md
mwachs5: sure i am not objecting to a spec change, just seems like that should be done broken out from a big rewrite of things, given that its like a one sentence change and deletion of an example to say that probes don't work on aggregates.

I am not making a statement on how chisel is implementing them now/ near future, defer to @dtzSiFive on that.

mmaloney-sf: @mwachs5 I promoted the PR to `[major]`.

spec.md
mmaloney-sf: I intentionally dropped the commas, since otherwise it could be read a bit ambiguously.

spec.md
mmaloney-sf: This was the existing copy. It's in rough shape.

I want to make a separate PR at some point to rename "type equivalence" to something more sane.

spec.md
mmaloney-sf: This was the original copy.

The entire notion of type inference (width and reset) are borked in the spec.

I'm not even sure it's worth the time to try and rewrite it until we have a more principled approach to handling it.

spec.md
mmaloney-sf: This was the original copy.

spec.md
mmaloney-sf: I had originally considered rewriting this section, but it will need some dedicated attention.

I'll change this intro back to the original copy.

spec.md
mmaloney-sf: This was the original copy.

spec.md
mmaloney-sf: This was the original copy.

spec.md
mmaloney-sf: This section is the original copy.

I'll weigh in on the `SInt<0>` battle at some point. But not today :)

spec.md
mmaloney-sf: This is the original copy.

spec.md
mmaloney-sf: Rewrote this section after consulting @dtzSiFive.

It's still hand-wavy, but only due to the fact that all of type inference is one big hand-wave at this point. It should suffice for now.

spec.md
mmaloney-sf: I'm flipping this back to `[patch]` and removing this line.

This point represents a pretty big gap. But let me leave this without the restriction, and we can address it after all the low-hanging fruit is gone.

mwachs5: thanks :) nit I see you reverted the sentence but you didn't revert the example, which is now just sort of ambiguous and confusing 😂 If you think this is a noncontroversial change can just open a PR right now to remove it, no need to block on this larger (in lines of text) change

spec.md
mmaloney-sf: Removed the references to hierarchical paths.

I did slip one usage of the term "physical hardware". Please consider this an informal statement. We will work towards a formalization of what this means before too long.

PR #148, opened 2023-11-11
[patch] Allow multiple group defs with same decl

seldridge:
Add language to the spec that allows multiple group definitions in the same module to reference the same group declaration.

This is done as it is more straightforward to have Chisel create groups whenever a module's constructor runs as opposed to waiting until the module is "closed" before creating a group.  Modules may have arbitrarily complex levels of inheritance and there's no great reason to force Chisel to have to combine groups itself.  This also forces Chisel users of groups to not treat groups as arbitrary scope escape hatches where anything defined in a group is visible to anything in a later declared group.  This better matches Chisel's scoping rules where it tries to align with Scala scoping rules.

This change does not disallow _merging groups_ in a FIRRTL compiler.  This is support that will be added to CIRCT [1].  The ABI is intentionally silent about how this situation should be lowered as the modules that are created to implement the logic of a group are an implementation detail. E.g., in the code example in this commit, it is legal to either create two modules that will be bound in or one module that will be bound in which combines the two group definitions.

[1]: https://github.com/llvm/circt/issues/6228

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -215,7 +215,7 @@ The FIRRTL ABI specification defines supported lowering convention.  One such st
 The `group`{.firrtl} keyword defines optional functionality inside a module.
 An optional group may only be defined inside a module.
 An optional group must reference a group declared in the current circuit.
-Declarations of identifiers and references to existing identifiers following the same lexical scoping rules as FIRRTL conditional statements (see: [@sec:conditional-scopes])---identifiers declared in the group definition may not be used outside the group while groups may refer to identifiers declared outside the group.
+An optional group forms a lexical scope (as with [@sec:conditional-scopes]) for all identifiers declared inside it---a group may use identifiers declared outside the group, but identifiers declared in the group may not be used in parent lexical scopes.
 The statements in a group are restricted in what identifiers they are allowed to drive.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -233,6 +233,26 @@ circuit:
       node notA = not(a)
 ```
 
+Multiple optional group definitions may reference the same group declaration.
+However, optional group definitions may not reference signals in other definitions, even if their group definitions reference the same group declaration.
```
darthscsi, line 237: Maybe "a group definition forms a lexical scope for all identifiers, two groups referencing the same group declaration are separate lexical scopes".

darthscsi (approved):
LGTM.

spec.md
seldridge: There's some more precision language earlier that is not as terse as this. (I like this.)

> Declarations of identifiers and references to existing identifiers following the same lexical scoping rules as FIRRTL conditional statements (see: [@sec:conditional-scopes])---identifiers declared in the group definition may not be used outside the group while groups may refer to identifiers declared outside the group.

I was attempting to say, "I shouldn't have to say this because it says the same thing two paragraphs up. However, I want to make sure you know you can't escape lexical scoping this way."

I'll add something closer to your language here.

PR #149, opened 2023-11-13
[major] Rename groups to layers

seldridge:
Change optional groups to "layers".  This much better captures the underlying concept as well as the possibility of nesting.  Change `declgroup` to `layer`. Change `group` to `layerblock`.

I recognize this is a pretty disruptive change. However, being able to explain this with the correct terms is highly beneficial. I always had difficulty talking about "nested optional groups" while "nested layers" makes more sense. I'm trying to draw analogies to layers in graphic design or existing [software engineering concepts](https://en.wikipedia.org/wiki/Layer_(object-oriented_design)).

This also inverts the focus in a good way. Layers are really about a collection of optional functionality in "layer blocks" that exists across disparate modules. The previous description incorrectly inverted this: the "groups"/"layers" were inside the module and the "groupdecl" was outside. h/t @mmaloney-sf for pointing this out.

I do want to get feedback on this, specifically from @mmaloney-sf, @mwachs5, and @darthscsi.

### DIFF
--- a/abi.md
+++ b/abi.md
@@ -116,96 +116,96 @@ Passive bundles shall be lowered to Verilog packed structs.
 
 Reference types in ports shall be logically split out from aggregates and named as though "Aggregate Type Lowering" was used.
 
-## On Groups
+## On Layers
 
-The lowering convention of a declared optional group specifies how an optional group will be lowered.
+The lowering convention of a declared layer specifies how a layer and all its associated layer blocks will be lowered.
 Currently, the only lowering convention that must be supported is `"bind"`{.firrtl.}.
 FIRRTL compilers may implement other non-standard lowering conventions.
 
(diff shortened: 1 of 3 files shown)


mmaloney-sf:
I'll look over this closely.

I like the use of the word "layer" instead of "group" -- it's a more flavorful word.

However, I'm wondering if we can push it a hair further: Rather than using `layerdecl`, can we think of a more interesting name for this keyword?

image

Think of it like this: What *is* the layer `Bar`? Is it the block inside of a the `Foo` module? Or is it the collectivity of all the `layer` blocks?

mwachs5:
love the name. haven't read the PR

mwachs5 (approved):

dtzSiFive:
Love the name change!! :rocket:

PR #150, opened 2023-11-13
[patch] Define Circuit Subcomponents

mmaloney-sf:
Adds a section defining Circuit Subcomponents.

This notion is latent in the implementations, but is central to FIRRTL's semantics. 

Intuitively, subcomponents are what you "dot into" or "index into". Eg, `c[0].x` is a reference to a subcomponent of a circuit component `c`.

Circuit subcomponents unify the notion of references, both static (sub-index and sub-field) and dynamic (sub-access). They are the entities which are the targets of a connect (... I believe this is a categorical statement). They are also important for reasoning about the initialization coverage of a module definition.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -582,15 +582,15 @@ Circuit subcomponents have both a kind and a type.
 
 We define this tree of subcomponents recursively by defining the a **direct subcomponent** relation:
 
-A circuit component with a ground type, an enumeration type, a probe type, or property type (see TODO) has no direct subcomponents.
+A circuit component with a ground type, an enumeration type, a probe type, or property type (see [@sec:types]) has no direct subcomponents.
 For example, `wire w : UInt<8>`{.firrtl} has no direct subcomponents.
 
-When a circuit component has a vector type (see TODO), it has as many direct subcomponents as its length.
+When a circuit component has a vector type (see [@sec:vector-types]), it has as many direct subcomponents as its length.
 Each subcomponent will have the same kind as its parent and will have the element type.
 For example, if we declare `wire v : UInt<8>[3]`{.firrtl}, it will have three direct subcomponents: `v[0]`{.firrtl}, `v[1]`{.firrtl}, and `v[2]`{.firrtl}.
 All three are wires and all three have type `UInt<8>`{.firrtl}.
 
-When a circuit component has a bundle type (see TODO), it will end up with one direct subcomponent for each field.
+When a circuit component has a bundle type (see [@sec:bundle-types]), it will end up with one direct subcomponent for each field.
 The kind of the subcomponent depends on both the kind and the type of the parent:
 
 -   For nodes, wires, and registers, the kind of each direct subcomponent is the same.
@@ -615,7 +615,7 @@ A **leaf component** is a circuit component has no direct subcomponents.
 Leaf components are useful for checking initialization.
 
 Two components are **disjoint** if they have no common subcomponent.
-TODO: For example, given a register `r`{.firrtl} with ...
+For example, given `wire w : { x : UInt<1>, y : UInt<8>[2] }`{.firrtl}, `w.x`{.firrtl} and `w.y`{.firrtl} are disjoint, but `w.y[0]`{.firrtl} and `w.y`{.firrtl} are not.
 
 ### References
 
@@ -1243,8 +1243,7 @@ module MyModule :
   connect myport, portx
 ```
 
-TODO: Fix this reference.
-See ??? for more details about sub-field expressions.
+See [@sec:references] for more details about indexing.
 
 ## Invalidates
 
--- a/spec.md
+++ b/spec.md
@@ -593,13 +593,13 @@ All three are wires and all three have type `UInt<8>`{.firrtl}.
 When a circuit component has a bundle type (see TODO), it will end up with one direct subcomponent for each field.
 The kind of the subcomponent depends on both the kind and the type of the parent:
 
-- For nodes, wires, and registers, the kind of each direct subcomponent is the same.
-- For output and input ports, the kind of each direct subcomponent depends on whether or not the field is flipped.
-  When the field is not flipped, the kind remains the same.
-  When the field is flipped, it will change from output to input or vice versa.
-- For submodule instances and memories, the kind of each direct subcomponent depends on whether or not the field is flipped.
-  When the field is not flipped, the kind is an input port.
-  When the field is flipped, the kind is an output port.
+-   For nodes, wires, and registers, the kind of each direct subcomponent is the same.
+-   For output and input ports, the kind of each direct subcomponent depends on whether or not the field is flipped.
+    When the field is not flipped, the kind remains the same.
+    When the field is flipped, it will change from output to input or vice versa.
+-   For submodule instances and memories, the kind of each direct subcomponent depends on whether or not the field is flipped.
+    When the field is not flipped, the kind is an input port.
+    When the field is flipped, the kind is an output port.
 
 If the bundle is not `const`{.firrtl}, the type of each subcomponent will simply be the type of the corresponding field.
 However, if the bundle is `const`{.firrtl}, the type of each subcomponent will be the `const`{.firrtl} version of the type of the corresponding field.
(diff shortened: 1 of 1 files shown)


mmaloney-sf:
For clarity, the direct subcomponent relation needs to be defined on the *cross product* of the kind and type of components. 

There are only seven kinds. Since there are infinitely many types, we need to break them into a finite number of cases:

* Ground types, enums, probes, and properties are grouped together
* Vector types
* Bundle types

The first case is the grab bag of types which don't give rise to proper subcomponents. 

The second case, vectors, splits up in the "obvious" way: you get one direct subcomponent per element in the vector and each has the base type.

The third case, bundles, is trickier because of flips. It still splits up in the "obvious" way: each field gives rise to a direct subcomponent with the field type of the corresponding field. However, the flip can change the subcomponent's kind.

When indexing into wires, nodes, and registers the kind never changes.
When indexing into instances and memories, the subcomponents are as if you are "viewing" the port from the outside. The effect on kinds is a bit confusing at first glance, but it should check out when you work through the logic. (Or else be *exactly* backwards. As a matter of policy, I refuse to make guarantees I get my orientations correct. Ever.) Remember: `output` ports need to be driven and `input` ports drive. You need to ignore the fact that `output` and `input` suggest an absolute orientation. It's really hard to put into words -- work through this section carefully to see what I mean.

spec.md
```diff
@@ -570,6 +572,58 @@ It indicates the type of the data being stored inside of the memory.
...
+Equivalently, these are the circuit components which are declared in a module.
+
+A **leaf component** is a circuit component has no direct subcomponents.
+Leaf components are useful for checking initialization.
+
+One circuit component is a **subcomponent** of another if there is way to get from the first component to the second through the direct subcomponent relation.
+A circuit component is trivially considered to be a subcomponent of itself.
+If we need to speak of the subcomponents excluding the component itself, we call these the **proper subcomponents**.
+
+When a circuit component has a ground type, an enumeration time, a probe type, or property type (see TODO), has no direct subcomponents.
```
seldridge, line 600:
```suggestion
When a circuit component has a ground type, an enumeration type, a probe type, or property type (see TODO), has no direct subcomponents.
```

While obvious, does the PR also want to state that there are no proper subcomponents either?

seldridge (commented):
Roughly makes sense.

spec.md
mmaloney-sf: A direct subcomponent is always a proper subcomponent. It's so immediate from the definition it shouldn't warrant mention.

However, it seems like I do need to give examples for these terms. They are entirely standard in mathematical discussions, but I should be mindful they take a small bit of working with them to build up an intuition.

mmaloney-sf:
I took what I had written up and cleaned it up a bit (apologies to @seldridge and @dtzSiFive for having to read my initial draft!):

- Lead off with two motivating examples
- Lead off with the full definition of **direct subcomponent**
- Added a third section on references (the proper name for anything that looks like `c.x[0]`).
- Fold in the sections on sub-field, sub-index, and sub-access (still a WIP)

mmaloney-sf:
I also identified a few quirks:

- `firtool` actually allows field access into muxes and other arbitrary expressions. This is disallowed by the grammar.
- `const` parses really funny. When you write `const UInt<8>[3]` it's as if you had written `(const UInt)<8>[3]` (with the `const` is on the inner type), whereas `const { x : UInt<8> }` has the `const` on the outer type. This leads to a really ugly rule about the type of subcomponents when the thing has a bundle type.
- There is a *lot* of ink spilled over sub-field, sub-index, and sub-access. It's mostly examples. As far as I can tell, this was an attempt to specify how these things work, despite not having defined the underlying concept (subcomponents). I'm planning to delete most of this material.

spec.md
```diff
@@ -538,6 +540,129 @@ It indicates the type of the data being stored inside of the memory.
...
+    When the field is flipped, it will change from output to input or vice versa.
+-   For submodule instances and memories, the kind of each direct subcomponent depends on whether or not the field is flipped.
+    When the field is not flipped, the kind is an input port.
+    When the field is flipped, the kind is an output port.
+
+If the bundle is not `const`{.firrtl}, the type of each subcomponent will simply be the type of the corresponding field.
+However, if the bundle is `const`{.firrtl}, the type of each subcomponent will be the `const`{.firrtl} version of the type of the corresponding field.
+
+A circuit component is a **subcomponent** of another if there is way to get from the first component to the second through the direct subcomponent relation.
+A circuit component is trivially considered to be a subcomponent of itself.
```
seldridge, line 608: Consider packing this sentence and the previous into an enumeration as either of these can be true for this to be a subcomponent.

spec.md
```diff
@@ -538,6 +540,129 @@ It indicates the type of the data being stored inside of the memory.
...
+If we need to speak of the subcomponents excluding the component itself, we call these the **proper subcomponents**.
+
+A **root component** is a circuit component that is not the direct subcomponent of any other component.
+Equivalently, these are the circuit components which are declared inside a module.
+
+A **leaf component** is a circuit component has no direct subcomponents.
```
seldridge, line 614:
```suggestion
A **leaf component** is a circuit component that has no direct subcomponents.
```

spec.md
```diff
@@ -538,6 +540,129 @@ It indicates the type of the data being stored inside of the memory.
...
+Leaf components are useful for checking initialization.
+
+Two components are **disjoint** if they have no common subcomponent.
+For example, given `wire w : { x : UInt<1>, y : UInt<8>[2] }`{.firrtl}, `w.x`{.firrtl} and `w.y`{.firrtl} are disjoint, but `w.y[0]`{.firrtl} and `w.y`{.firrtl} are not.
```
seldridge, line 618: Great addition of "disjoint".

spec.md
```diff
@@ -538,6 +540,129 @@ It indicates the type of the data being stored inside of the memory.
...
+These are constructed using the names of root components, and optionally, through the dot and bracket operators.
+
+For instance, suppose we declare `wire w : { x : UInt<1>, v : UInt<8>[2] }`{.firrtl}.
+Then `w`{.firrtl}, `w.x`{.firrtl}, `w.y`{.firrtl}, `w.y[0]`{.firrtl}, and `w.y[1]`{.firrtl}, and `w.y[i]`{.firrtl} are all examples of references.
+
+When a circuit component has a vector type, you can perform both static and dynamic indexing.
+For example, given `wire v : UInt<8>[2]`{.firrtl}, both `v[0]`{.firrtl} and `v[1]`{.firrtl} are references using static indexing.
+On the other hand, `v[i]`{.firrtl} is a reference using dynamic indexing, given an expression `i`{.firrtl} with an unsigned integer type.
+
+The result of an out-of-bounds dynamic index is an indeterminate value (see [@sec:indeterminate-values]).
```
seldridge, line 632: Do you want to state that an out-of-bounds static index is illegal?

spec.md
```diff
@@ -538,6 +540,129 @@ It indicates the type of the data being stored inside of the memory.
...
+
+A **static reference** is a reference where all indexing is static.
+For instances, `v[0]`{.firrtl}, `v[1]`{.firrtl}, and `v[2]`{.firrtl} are static references.
+A static reference identify a concrete subcomponent within the module definition.
```
seldridge, line 636:
```suggestion
A static reference indetifies a concrete subcomponent within the module definition.
```

spec.md
```diff
@@ -538,6 +540,129 @@ It indicates the type of the data being stored inside of the memory.
...
+    connect v[0], UInt(0)
+  when eq(i, UInt(1)) :
+    connect v[1], UInt(0)
+  when eq(i, UInt(2)) :
+    connect v[2], UInt(0)
+  when eq(i, UInt(3)) :
+    connect v[3], UInt(0)
+  else :
+    invalidate v
+```
```
seldridge, line 664: This is not quite right. This either needs to have an `else : invalidate v` after every when or a single `invalidate v` needs to be moved before the first `when`. `v[0]`, `v[1]`, and `v[2]` are not properly initialized. This isn't normally important in examples like this. However, the intent is to show equivalence with the other code.

seldridge (approved):
LGTM

Mostly nits. Clean up as you choose and land it. 👍

PR #154, opened 2023-11-15
[patch] Probes Section cleanup

mmaloney-sf:
Organize and clean up the section on probes after they were disrupted by #143 and #145.


spec.md
```diff
@@ -2070,103 +2098,20 @@ module Foo:
...
-  ; ...
+Both `Probe`{.firrtl} and `RWProbe`{.firrtl} may be read from using the `read`{.firrtl} expression.
 
-module Forward :
-  output p : Probe<UInt>
-
-  inst f of Foo
-  define p = f.p[0][1]
-```
+Keep in mind that probes are intended for verification and isn't intended to be synthesizable.
```
mwachs5, line 2106:
```suggestion
Keep in mind that probes are intended for verification and aren't intended to be synthesizable.
```

spec.md
```diff
@@ -2070,103 +2098,20 @@ module Foo:
...
+A `read`{.firrtl} expression can reach arbitrarily far into the module hierarchy to read from a distant circuit component, which is arguably unphysical at worst and potentially expensive at best.
```
mwachs5, line 2107: all of this feels very... commentary-like. I'm not sure what it really has to do with the firrtl spec , especially without considering the one ABI.

spec.md
```diff
@@ -2070,103 +2098,20 @@ module Foo:
...
+The `release`{.firrtl} and `release_initial`{.firrtl} statements end the forcing, reconnecting the circuit component to its usual driver.
 
-## Force and Release
-
-To override existing drivers for a `RWProbe`{.firrtl}, the `force`{.firrtl} statement is used, and released with `release`{.firrtl}.
-Force statements are simulation-only constructs and may not be supported by all implementations.
-They are similar to the verification statements (e.g., `assert`{.firrtl}) in this regard.
-
-These are two variants of each, enumerated below:
+Like `read`{.firrtl}, the force statements are verification constructs.
```
mwachs5, line 2115: what is a "verification construct". Is this a term we've defined? It's rather confusing given we have a bunch of other verification like constructs such as "printf" and "assert" -- how are these things like eachother?

spec.md
```diff
@@ -2320,139 +2265,55 @@ module DUT :
...
+## Probes and Layers
 
-  connect y, read(u2.out) ; = x
-```
-
-In the above example, the probe of node `n`{.firrtl} is routed through two modules before its resolution.
-
-### Invalid Input Reference
+`Probe`{.firrtl} and `RWProbe`{.firrtl} types may be associated with a layer (see [@sec:layers]).
+When associated with a layer, the reference type may only be driven from tha layer.
```
mwachs5, line 2272:
```suggestion
When associated with a layer, the reference type may only be driven from that layer.
```

spec.md
mmaloney-sf: Agreed. However, I think it's good to have this as a placeholder until we can formalize the distinction between "synthesizable"/"hardware" and "verification-only"/"simulation-only".

spec.md
```diff
@@ -2320,139 +2265,55 @@ module DUT :
...
-
-    inst ud1 of UpDown
-    connect ud1.in, in
-    define ud1.in_ref = ud1.r1
-
-    inst ud2 of UpDown
-    connect ud2.in, in
-    define ud2.in_ref = ud2.r2
+Probes are generally lowered to hierarchical names in Verilog.
+For details, see the FIRRTL ABI Specification.
```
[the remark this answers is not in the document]
mwachs5: (i see the same sentence repeated just below)

spec.md
mmaloney-sf: I am moving towards a formal definition of these things.

We'll get there 👍

spec.md
```diff
@@ -2320,139 +2265,55 @@ module DUT :
...
 
-module Connect:
-  output out : {pref: Probe<UInt>, cref: Probe<UInt>}
+FIRRTL allows probes to be passed through ports.
+However, due to the limitations of Verilog, there is an important restriction on where they may be used.
 
-  inst a of Consumer
-  inst b of Producer
+A module which only sends probes up towards its parent via output ports have no restriction.
+However, when a module receives a probe from its parent through an input port, you may not use it in a `read`{.firrtl} expression nor may you force it if the circuit component it refences does not live in a module that is a descendent of the current module.
```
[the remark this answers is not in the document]
mwachs5: is this saying that it's ok for a module `M` to have an input probe port, as long as outside of `M` that input probe is actually driven by an output probe port of `M` itself (either directly or from some child `C` of `M` where C has the output probe port?)

spec.md
```diff
@@ -2320,139 +2265,55 @@ module DUT :
...
+### External Modules
 
-module Top:
-  inst c of Connect
+TODO
 
-  node producer_debug = read(c.out.pref); ; Producer-side signal
-  node consumer_debug = read(c.out.cref); ; Consumer-side signal
-```
+Probe types may be specified as part of an external module (see [@sec:externally-defined-modules]), with the resolved referent for each specified using `ref`{.firrtl} statements.
```
mwachs5:
```suggestion
Probe type ports may be specified on the interface of an external module (see [@sec:externally-defined-modules]), with the resolved referent for each specified using `ref`{.firrtl} statements.
```

spec.md
```diff
@@ -2023,18 +2023,47 @@ cover(clk, pred, en, "X equals Y when Z is valid") : optional_name
...
+## The `probe`{.firrtl} and `rwprobe`{.firrtl} Expressions
+
+Given a reference to a circuit component, you can capture it in a probe with `probe`{.firrtl} for a read-only probe or with `rwprobe`{.firrtl} for a read-writeable probe.
+
+## The `define`{.firrtl} Statement
+
+Since neither `Probe<T>`{.firrtl} nor `RWProbe<T>`{.firrtl} is a connectable type ([@sec:connectable-types]), circuit components with these types cannot be connected to.
+Instead, we use the `define`{.firrtl} statement to route probes through a FIRRTL circuit.
+
+The `define`{.firrtl} statements works similarly to `connect`{.firrtl}, but it has special interactions with conditional statements ([@sec:conditionals]).
```
dtzSiFive, line 2060:
```suggestion
The `define`{.firrtl} statement works similarly to `connect`{.firrtl}, but it has special interactions with conditional statements ([@sec:conditionals]).
```

spec.md
```diff
@@ -2023,18 +2023,47 @@ cover(clk, pred, en, "X equals Y when Z is valid") : optional_name
...
+`define`{.firrtl} statements may appear within conditional blocks, and when they do, they may connect circuit components defined in a local scope.
```
dtzSiFive, line 2061:
```suggestion
`define`{.firrtl} statements may appear within conditional blocks, and when they do, they may refer to circuit components defined in a local scope.
```
(mostly, probably should avoid saying `define` "connects" anything? And I think idea here is answering the question "why would you use define from within a conditional?" and the answer is "so the things you're referring to are in-scope" (right?))

spec.md
```diff
@@ -2023,18 +2023,47 @@ cover(clk, pred, en, "X equals Y when Z is valid") : optional_name
...
+However, `define`{.firrtl} are not subject to last connect semantics, and the `define`{.firrtl} statement is always and unconditionally in effect.
```
dtzSiFive, line 2062:
```suggestion
However, `define`{.firrtl} statements are not subject to last connect semantics, and the `define`{.firrtl} statement is always and unconditionally in effect.
```
(feels awkward to use "define" as plural?)

spec.md
```diff
@@ -2070,103 +2099,20 @@ module Foo:
...
-```
-
-## Force and Release
-
-To override existing drivers for a `RWProbe`{.firrtl}, the `force`{.firrtl} statement is used, and released with `release`{.firrtl}.
-Force statements are simulation-only constructs and may not be supported by all implementations.
-They are similar to the verification statements (e.g., `assert`{.firrtl}) in this regard.
-
-These are two variants of each, enumerated below:
+Like `read`{.firrtl}, the force statements are verification constructs.
```
dtzSiFive, line 2115: Just checking, "force" is not code-formatted because it's referring to the force "family" of statements?

spec.md
```diff
@@ -2320,139 +2266,52 @@ module DUT :
...
+## Probes and Layers
 
-  connect y, read(u2.out) ; = x
-```
-
-In the above example, the probe of node `n`{.firrtl} is routed through two modules before its resolution.
-
-### Invalid Input Reference
+`Probe`{.firrtl} and `RWProbe`{.firrtl} types may be associated with a layer (see [@sec:layers]).
+When associated with a layer, the reference type may only be driven from that layer.
```
dtzSiFive, line 2272:
```suggestion
When associated with a layer, the probe type may only be driven from that layer.
```
?

spec.md
```diff
@@ -2320,139 +2266,52 @@ module DUT :
...
+## Limitations
 
-The following demonstrates an example of this, where it's convenient to use the same bundle type as both output to one module and input to another, with references populated by both modules targeting signals of interest at each end.
-For this to be the same bundle type -- input on one and output on another -- the `Probe` references for each end should be output-oriented and accordingly are input-oriented at the other end.
-It would be inconvenient to generate this design so that each has output probe references only.
+### `RWProbe`{.firrtl} with Aggregate Types
 
-The `Connect` module instantiates a `Producer` and `Consumer` module, connects them using a bundle with references in both orientations, and forwards those references for inspection up the hierarchy.
-The probe targets are not significant, here they are the same data being sent between the two, as stored in each module.
+Probes of aggregates do not working under all ABI versions.
```
dtzSiFive, line 2291:
```suggestion
Probes of aggregates do not work under all ABI versions.
```

spec.md
```diff
@@ -2320,139 +2266,52 @@ module DUT :
...
-  wire x : UInt
-  define out.pref = probe(x)
-  ; ...
-  connect out.a, x
+### No `RWProbe`{.firrtl}s of Ports on Public Modules
 
-module Connect:
-  output out : {pref: Probe<UInt>, cref: Probe<UInt>}
+Conceptually, in order to force a value onto a circuit component, you need to be able to disconnect it from its former driver.
+However, the driver of a port is another port in some other module.
```
dtzSiFive, line 2299: This language is a bit off, LMK if not clear why or how to capture what you mean :+1: .

spec.md
```diff
@@ -2320,139 +2266,52 @@ module DUT :
...
+And so, to force a port, you must have control over both "ends" of a port.
 
-  inst a of Consumer
-  inst b of Producer
+Trouble arises with ports on a public module ([@sec:public-modules]).
+Since we don't control the remote side of ports (since they might live in another FIRRTL file), we cannot force these ports either.
+For this reason, it is an error to use `rwprobe`{.firrtl} on any port on a public module.
```
dtzSiFive, line 2304: Up to you, but at least to me this is same issue for public and extmodule's (and what used to be the main module/top)-- (the module definition, if FIRRTL-provided, must be public for it to be used as extmodule elsewhere).  Anyway I think this section could be tidied a little, I'll revisit with more concrete suggestions or we can work on it together if you'd like.

dtzSiFive: FWIW neither the other side of a public of extmodule necessarily is defined in FIRRTL.

dtzSiFive (commented):
Looking good! Made pass with some feedback/nits.

spec.md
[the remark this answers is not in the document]
dtzSiFive: Yes that's the current rule (and also that what the probe at-point-of-use-in-a-read-or-force resolves to must be same in all instantiation contexts...).  The n-turn part of the de-squiggling was implemented but not landed yet in favor of focusing on what we practically would use (u-turn de-squiggling), at least in theory.  This kept things simpler, at least until alternate choice re-introduces the need for proving they're "equivalent" in all possible bottom-up cases.

This sort of thing: https://github.com/dtzSiFive/reftype-inputs/blob/main/ref_in_eq.fir (or more complicated: https://github.com/dtzSiFive/reftype-inputs/blob/main/ref_in_eq_subs_parent.fir )

If this PR is meant to keep things the same but clean them up, that's the current rule yes.  We may want to change it however.

spec.md
mmaloney-sf: Yeah.

PR #155, opened 2023-11-28
[major] Define option groups and instance choices

nandor:
Add option and instance choice features to the FIRRTL spec. Define an ABI for these that uses files similar to how layers work.

There are more examples of syntax and possible ABIs in this repository: https://github.com/seldridge/instance-choice-abis

### More Information

This feature exists to serve two separate purposes both related to allowing for limited parameterization of a FIRRTL circuit. This parameterization is later removed via _specialization_ either: (1) by a FIRRTL compiler or (2) via Verilog elaboration.

_Targets_ describe available paramterization as well as legal _options_ for that target. (Names subject to change.) These target can then be used by FIRRTL constructs to describe behavior which is dependent on the value of a target.

Currently, only one construct can depend on the value of a target: _instance choice_. An instance choice is an instantiation of one of several modules where the choice of instantiation is controlled by the option selected for a given target. E.g., a likely target is one which captures "FPGA or ASIC or Generic" that, when set, can specialize the design for an FPGA, and ASIC, or give generic Verilog. E.g., a more "parameter-like" target would be one that sets the cache size.

All modules instantiated by an instance choice must have the exact same port-level interface including probes and properties. This is necessary to allow them to be substituted. The underlying hardware backing the probes and properties may vary from one instance choice to another. _This is intentional and critical for this to be useful._ Properties require no special treatment other than a specialization must be chosen before properties can be evaluated. Probes require tedious Verilog macros that allow the Verilog-level specialization to set the internal path.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -326,8 +326,8 @@ Specialization can occur either in the compiler or it can be materialized in the
 For details, consult the FIRRTL ABI specification.
 Specialization is not mandatory: options can be left unspecified, resorting to explicitly-defined default behaviour.
 
-A target may be delcared using the `target`{.firrtl} keyword.
-The availble options for which a target may take are listed using the `option`{.firrtl} keyword.
+A target may be declared using the `target`{.firrtl} keyword.
+The available options for which a target may take are listed using the `option`{.firrtl} keyword.
 An example FIRRTL circuit showing two targets, `Platform` and `Performance`, and their allowable options is shown below:
 
 ``` firrtl
@@ -534,7 +534,7 @@ By default, it will instantiate `DefaultClockGate` and when `Platform` is `FPGA`
 
 ``` firrtl
 circuit:
-  targe Platform:
+  target Platform:
     option FPGA
     option ASIC
 
@@ -4203,7 +4203,7 @@ decl =
   | decl_extmodule
   | decl_intmodule
   | decl_layer
-  | decl_option
+  | decl_target
   | decl_type_alias ;
 
 decl_module =
@@ -4232,9 +4232,9 @@ decl_layer =
     { decl_layer , newline } ,
   dedent ;
 
-decl_option =
-  "option" , id , ":" , [info] , newline, indent ,
-    { "case" , id , ":" , [ info ] , newline } ,
+decl_target =
+  "target" , id , ":" , [info] , newline, indent ,
+    { "option" , id , ":" , [ info ] , newline } ,
   dedent ;
 
 decl_type_alias = "type", id, "=", type ;
--- a/spec.md
+++ b/spec.md
@@ -152,32 +152,6 @@ Private modules have none of the restrictions of public modules.
 Private modules have no stable, defined interface and may not be used outside the current circuit.
 A private module may not be physically present in a compiled circuit.
 
-## Option Groups
-
-The option group mechanism declares configurable parameters with a pre-defined set of values to enable the specialization of designs during or after lowering.
-Options groups address the need for configurability.
-For example, designs may need to behave differently on ASIC and FPGA platforms, but at the time of FIRRTL elaboration it is not known which platform the design will be used on.
-This feature allows such choices to be expressed and embedded into the design.
-
-The `option`{.firrtl} keyword declares an option group, which contains `case`{.firrtl} declarations naming the settings allotted to that option.
-The circuit can be specialized for a single case of a given option at any time.
-Multiple option groups can be declared to capture orthogonal dimensions of configuration.
-
-Specialization can occur either in the compiler or it can be materialized in the lowering.
-For details, consult the FIRRTL ABI specification.
-Specialization is not mandatory: options can be left unspecified, resorting to explicitly-defined default behaviour.
-
-``` firrtl
-circuit:
-  option Platform:
-    case FPGA:
-    case ASIC:
-
-  option Performance:
-    case Slow:
-    case Fast
-```
-
 ## Externally Defined Modules
 
 Externally defined modules are modules whose implementation is not provided in the current circuit.
@@ -331,6 +305,40 @@ FIRRTL version 4.0.0
 circuit Foo :
   layer A, bind :
   public module Foo enablelayer A :
+```
+
+## Targets
+
+A `target`{.firrtl} describes one way that a FIRRTL circuit may be specialized for a certain use case.
+Here, specialization means choosing a specific option for a target.
+
+It is often desirable to have one FIRRTL circuit that has different logic when simulated, synthesized and mapped to a field-programmable gate array (FPGA), or synthesized to a given process technology and standard cell library.
+While this per-target customizability can be expressed and specialized in a frontend language that produces FIRRTL, it is often desirable to expose the target specialization in the FIRRTL.
+By delaying the specialization, the specialization can either be done by a FIRRTL compiler or exposed in the artifacts of a FIRRTL compiler for specialization by the consumer.
+
+Practically, targets describe a limited form of parameterization and, if not specialized by a FIRRTL compiler, allow for FIRRTL compilers to generate parametric artifacts (e.g., parametric Verilog).
+
+The `option`{.firrtl} keyword declares an option group, which contains `case`{.firrtl} declarations naming the settings allotted to that option.
+The circuit can be specialized for a single case of a given option at any time.
+Multiple option groups can be declared to capture orthogonal dimensions of configuration.
+
+Specialization can occur either in the compiler or it can be materialized in the lowering.
+For details, consult the FIRRTL ABI specification.
+Specialization is not mandatory: options can be left unspecified, resorting to explicitly-defined default behaviour.
+
+A target may be delcared using the `target`{.firrtl} keyword.
+The availble options for which a target may take are listed using the `option`{.firrtl} keyword.
+An example FIRRTL circuit showing two targets, `Platform` and `Performance`, and their allowable options is shown below:
+
+``` firrtl
+circuit:
+  target Platform:
+    option FPGA
+    option ASIC
+
+  target Performance:
+    option Slow
+    option Fast
 ```
 
 # Circuit Components
@@ -507,35 +515,46 @@ circuit Foo:
     ;; snippetend
 ```
 
-#### Instance Choices
+#### Instance Choice
 
-FIRRTL supports the specialization of designs through the `instchoice`{.firrtl} declaration, which selects the instantiated module based on an `option`{.firrtl}.
+An instance choice is a submodule instance where the choice of submodule is conditioned based on the value of a `target`{.firrtl}.
+This enables per-instance specialization for different targets.
+Additionally, this is a mechanism for module replacement.
 
-Example:
+An instance choice declaration specifies the instance name and names the option group based on which the choices are selected.
+A default module must be provided.
+The default module is instantiated by the instance choice when an option for its associated target is not specified.
+Subsequently, modules can be specified for the known choices of the selected option group.
+An instance choice does not need to specify modules for all cases.
+The instantiated modules must be either modules or external modules.
+
+An example of an instance choice is shown below.
+This instance choice is conditioned on the `Platform` target.
+By default, it will instantiate `DefaultClockGate` and when `Platform` is `FPGA` it will instantiate `FPGAClockGate`.
 
 ``` firrtl
 circuit:
-  option Platform:
-    case FPGA:
-    case ASIC:
+  targe Platform:
+    option FPGA
+    option ASIC
+
+  module DefaultClockGate:
+    input clock_in: Clock
+    output clock_out: Clock
+    input enable: UInt<1>
+
+  extmodule FPGAClockGate:
+    input clock_in: Clock
+    output clock_out: Clock
+    input enable: UInt<1>
 
   module InstanceChoice:
-    instchoice clock_gate of DefaultTarget, Platform:
-      FPGA => FPGATarget
-
-  module DefaultTarget:
-
-  module FPGATarget:
+    instchoice clock_gate of DefaultClockGate, Platform:
+      FPGA => FPGAClockGate
 ```
 
-The instance choice declaration specifies the instance name and names the option group based on which the choices are selected.
-A default module is provided, to be instantiated when the design is not specialized or it is not specialised for a known case.
-Subsequently, modules can be specified for the known choices of the selected option group.
-The operation does not need to specify modules for all cases.
-The references must be either modules or extmodules.
-
-The type of the instance bundle is determined identically to regular instances.
-The port lists of all modules must match and the ports are limited to ground and aggregate types.
+The type of an instance choice is the same as an instantiation of the default module.
+The ports of all module choices must be the same as the default module.
 
 ### Memories
 
--- a/spec.md
+++ b/spec.md
@@ -152,6 +152,32 @@ Private modules have none of the restrictions of public modules.
 Private modules have no stable, defined interface and may not be used outside the current circuit.
 A private module may not be physically present in a compiled circuit.
 
+## Option Groups
+
+The option group mechanism declares configurable parameters with a pre-defined set of values to enable the specialization of designs during or after lowering.
+Options groups address the need for configurability.
+For example, designs may need to behave differently on ASIC and FPGA platforms, but at the time of FIRRTL elaboration it is not known which platform the design will be used on.
+This feature allows such choices to be expressed and embedded into the design.
+
+The `option`{.firrtl} keyword declares an option group, which contains `case`{.firrtl} declarations naming the settings allotted to that option.
+The circuit can be specialized for a single case of a given option at any time.
+Multiple option groups can be declared to capture orthogonal dimensions of configuration.
+
+Specialization can occur either in the compiler or it can be materialized in the lowering.
+For details, consult the FIRRTL ABI specification.
+Specialization is not mandatory: options can be left unspecified, resorting to explicitly-defined default behaviour.
+
+``` firrtl
+circuit:
+  option Platform:
+    case FPGA:
+    case ASIC:
+
+  option Performance:
+    case Slow:
+    case Fast
+```
+
 ## Externally Defined Modules
 
 Externally defined modules are modules whose implementation is not provided in the current circuit.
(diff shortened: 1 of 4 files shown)


spec.md
```diff
@@ -516,6 +541,36 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+      when FPGA FPGATarget
+      default DefaultTarget
+
+  module DefaultTarget:
+
+  module FPGATarget:
+
+```
+
+The instance choice declaration specifies the instance name and names the option group based on which the choices are selected.
```
dtzSiFive, line 566: Is `inst_choice` of an `extmodule` or `intmodule` allowed? Do all cases need to map to the same?

spec.md
```diff
@@ -516,6 +541,36 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+It introduces mappings with the `when`{.firrtl} keyword, attaching a module to a specific case allowed by the option group.
+A default option is always present, to be implicitly used when no specialisation is requested at any stage.
+
+Probes or properties are not allowed on modules which are part of instance choices.
+The type of the instance bundle is determined identically to regular instances.
+The port lists of all modules must match.
+
```
dtzSiFive, line 573: Please update the grammar (at the end of this file).

spec.md
```diff
@@ -297,6 +297,31 @@ circuit:
 Layer blocks will be compiled to modules whose ports are derived from what they capture from their visible scope.
 For full details of the way layers are compiled, see the FIRRTL ABI specification.
 
+## Options
+
+The option mechanism declares configurable parameters with a pre-defined set of values.
+Options are primarily intended to enable target-specific (ASIC or FPGA) lowerings, but they can also be used specify parameters which are present in the output and
+can be selected after emission (for example, to tweak a design for size or performance).
```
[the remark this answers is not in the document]
mwachs5: i think we have a formatting script to handle this...? @mmaloney-sf is there a script one can run to just do the right thing here?

spec.md
nandor: Intmodules should not be allowed here, only full-fledged modules.

spec.md
```diff
@@ -516,6 +541,36 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+
+  module DefaultTarget:
+
+  module FPGATarget:
+
+```
+
+The instance choice declaration specifies the instance name and names the option group based on which the choices are selected.
+It introduces mappings with the `when`{.firrtl} keyword, attaching a module to a specific case allowed by the option group.
+A default option is always present, to be implicitly used when no specialisation is requested at any stage.
```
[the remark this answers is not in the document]
nandor: There is no need to specify all cases.

spec.md
```diff
@@ -516,6 +541,36 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+
+Example:
+``` firrtl
+circuit:
+  option Platform:
+    case FPGA:
+    case ASIC:
+
+  module InstanceChoice:
+    inst_choice clock_gate option Platform
```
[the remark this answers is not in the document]
nandor: `The port lists of all modules must match.`

spec.md
```diff
@@ -516,6 +539,37 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+  module FPGATarget:
+```
+
+The instance choice declaration specifies the instance name and names the option group based on which the choices are selected.
+It introduces mappings with the `when`{.firrtl} keyword, attaching a module to a specific case allowed by the option group.
+A default option must always be provided, to be implicitly used when no specialisation is requested at any stage.
+The mapping does not need to be exhaustive, cases can be left unspecified, resorting to the default mapping.
+Only modules or extmodules can be targeted.
+
+Probes or properties are not allowed on modules which are part of instance choices.
```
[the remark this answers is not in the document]
mwachs5: or whatever we call "ports" in this document

mmaloney-sf:
I am concerned about the orthogonality of this feature with external modules and intrinsic modules. It's hard for me to tell how this functionality is different in intent from those two features.

I also get a feeling this feature is overfitted to one particular usecase. Let's make sure we're being especially mindful of the complexity budget in FIRRTL. A feature like this feels to me like it might be better captured at the Chisel level. Are there arguments otherwise?

More detailed review is incoming.

spec.md
```diff
@@ -516,6 +539,37 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+  module DefaultTarget:
+
+  module FPGATarget:
+```
+
+The instance choice declaration specifies the instance name and names the option group based on which the choices are selected.
+It introduces mappings with the `when`{.firrtl} keyword, attaching a module to a specific case allowed by the option group.
+A default option must always be provided, to be implicitly used when no specialisation is requested at any stage.
+The mapping does not need to be exhaustive, cases can be left unspecified, resorting to the default mapping.
+Only modules or extmodules can be targeted.
```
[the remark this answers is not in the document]
nandor: Updated - simplified the syntax.

spec.md
```diff
@@ -516,6 +539,37 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+
+``` firrtl
+circuit:
+  option Platform:
+    case FPGA:
+    case ASIC:
+
+  module InstanceChoice:
+    inst_choice clock_gate of DefaultTarget, Platform:
+      FPGATarget => FPGA
```
seldridge, line 556:
```suggestion
      FPGA => FPGATarget
```

spec.md
```diff
@@ -297,6 +297,29 @@ circuit:
 Layer blocks will be compiled to modules whose ports are derived from what they capture from their visible scope.
 For full details of the way layers are compiled, see the FIRRTL ABI specification.
 
+## Options
+
+The option mechanism declares configurable parameters with a pre-defined set of values.
+Options are primarily intended to enable target-specific (ASIC or FPGA) lowerings, but they can also be used specify parameters which are present in the output and can be selected after emission (for example, to tweak a design for size or performance).
+Specialization can occur either in the compiler or in the output via the ABI. For full details, consult the FIRRTL ABI.
```
seldridge, line 304: Nit:
```suggestion
Specialization can occur either in the compiler or in the output via the ABI.
For full details, consult the FIRRTL ABI.
```

seldridge, line 304: You can collapse both of these sentences together. Options exist to enable specialization. Specialization may then happen before or after Verilog emission.

I am concerned about how to handle the specialization-during-compilation case as this seems to violate an ABI guarantee.

spec.md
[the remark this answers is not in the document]
seldridge: Text needs to be updated to remove references to `when`.

spec.md
```diff
@@ -297,6 +297,29 @@ circuit:
...
+
+The `option`{.firrtl} keyword declares an option group, which contains `case`{.firrtl} declarations naming the settings allotted to that option.
+The circuit can be specialized for a single case of a given option at any time. Multiple option groups can be declared to capture orthogonal dimensions of configuration.
+Constructs that rely on options assume an implicit default, which is selected when no specific case is specified.
+
+``` firrtl
+circuit:
+  option Platform:
+    case FPGA:
+    case ASIC:
```
seldridge, line 314: Thoughts on `option` and `case` @mmaloney-sf, @darthscsi?

spec.md
```diff
@@ -516,6 +539,37 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+  module FPGATarget:
+```
+
+The instance choice declaration specifies the instance name and names the option group based on which the choices are selected.
+It introduces mappings with the `when`{.firrtl} keyword, attaching a module to a specific case allowed by the option group.
+A default option must always be provided, to be implicitly used when no specialisation is requested at any stage.
+The mapping does not need to be exhaustive, cases can be left unspecified, resorting to the default mapping.
+Only modules or extmodules can be targeted.
+
+Probes or property ports are not allowed on modules which are part of instance choices.
```
seldridge, line 569: Why properties given that these are "evaluated" later?

Why no probes? It seems like this could be made to work if the `define` for a reference included another define that the optional instance must define?

nandor: They can be made to work, but it's not clear to me how at the moment and the feature is useful even without them.

seldridge:
@rwy7: FYI, give this some thought about composition with true Verilog width parameterization.

mmaloney-sf (commented):
I know we have an internal usecase for this. But let's make sure we have the abstraction really crisp before pushing it forward. Especially if we could "fake it" with external modules and annotations in the meantime.

spec.md
```diff
@@ -516,6 +539,37 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+Example:
+
+``` firrtl
+circuit:
+  option Platform:
+    case FPGA:
+    case ASIC:
+
+  module InstanceChoice:
+    inst_choice clock_gate of DefaultTarget, Platform:
```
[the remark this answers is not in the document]
nandor: https://en.cppreference.com/w/cpp/keyword, but `instchoice` is fine.

spec.md
```diff
@@ -297,6 +297,29 @@ circuit:
 Layer blocks will be compiled to modules whose ports are derived from what they capture from their visible scope.
 For full details of the way layers are compiled, see the FIRRTL ABI specification.
 
+## Options
+
+The option mechanism declares configurable parameters with a pre-defined set of values to enable the specialization of designs during or after lowering.
```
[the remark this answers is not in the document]
nandor: I do not see the parallel to extmodules and this point keeps popping up! This is pretty much a sum type, a fundamentally different language concept. Please clarify why do you see this connection so we can settle this.

spec.md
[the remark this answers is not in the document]
nandor: It is not 'silly' and please refrain from using such terms in written communication, as this was the result of somewhat careful consideration.

Each option defines an orthogonal dimension of configuration, implying the possibility of composition. We opted to make it explicit, limiting the operation to a choice along a single direction. This leaves composition to the user, who is now responsible for building the appropriate chain of instance choices if that is desired. Alternatively, we risk turning this into a full-blown pattern matching operation.

nandor:
I pushed a rewrite. I would want to focus on the syntax (as I am fairly open to changes) and I want to settle the debate about extmodules. Implementation-wise, an `op` in the FIRRTL dialect is necessary, but I can move forward with that.

spec.md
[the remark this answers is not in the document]
mmaloney-sf: Well sure enough. My C++ isn't so great :(

spec.md
```diff
@@ -297,6 +297,32 @@ circuit:
...
+For example, designs may need to behave differently on ASIC and FPGA platforms, but at the time of FIRRTL elaboration it is not known which platform the design will be used on.
+This feature allows such choices to be expressed and embedded into the design.
+
+The `option`{.firrtl} keyword declares an option group, which contains `case`{.firrtl} declarations naming the settings allotted to that option.
+The circuit can be specialized for a single case of a given option at any time.
+Multiple option groups can be declared to capture orthogonal dimensions of configuration.
+
+Specialization can occur either in the compiler or it can be materialized in the lowering.
+For details, consult the FIRRTL ABI specification.
+Specialization is not mandatory: options can be left unspecified, resorting to explicitly-defined default behaviour.
```
mwachs5, line 313:
```suggestion
Specialization is not mandatory: options can be left unspecified, resorting to explicitly-defined default behavior.
```

nit: american spelling

spec.md
```diff
@@ -516,6 +542,36 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+    instchoice clock_gate of DefaultTarget, Platform:
+      FPGA => FPGATarget
+
+  module DefaultTarget:
+
+  module FPGATarget:
+```
+
+The instance choice declaration specifies the instance name and names the option group based on which the choices are selected.
+A default module is provided, to be instantiated when the design is not specialized or it is not specialised for a known case.
```
mwachs5, line 567:
```suggestion
A default module is provided, to be instantiated when the design is not specialized or it is not specialized for a known case.
```

spec.md
```diff
@@ -516,6 +542,36 @@ The type of the submodule instance `passthrough`{.firrtl} above is thus:
...
+Subsequently, modules can be specified for the known choices of the selected option group.
+The operation does not need to specify modules for all cases.
+The references must be either modules or extmodules.
+
+The type of the instance bundle is determined identically to regular instances.
+The port lists of all modules must match and the ports are limited to ground and aggregate types.
```
mwachs5, line 573: is inference (Reset, width) allowed across these ports? If so must they resolve to the same thing for all choices?

spec.md
```diff
@@ -307,6 +307,40 @@ circuit Foo :
...
+
+It is often desirable to have one FIRRTL circuit that has different logic when simulated, synthesized and mapped to a field-programmable gate array (FPGA), or synthesized to a given process technology and standard cell library.
+While this per-target customizability can be expressed and specialized in a frontend language that produces FIRRTL, it is often desirable to expose the target specialization in the FIRRTL.
+By delaying the specialization, the specialization can either be done by a FIRRTL compiler or exposed in the artifacts of a FIRRTL compiler for specialization by the consumer.
+
+Practically, targets describe a limited form of parameterization and, if not specialized by a FIRRTL compiler, allow for FIRRTL compilers to generate parametric artifacts (e.g., parametric Verilog).
+
+The `option`{.firrtl} keyword declares an option group, which contains `case`{.firrtl} declarations naming the settings allotted to that option.
+The circuit can be specialized for a single case of a given option at any time.
+Multiple option groups can be declared to capture orthogonal dimensions of configuration.
```
dtzSiFive, line 323: Are targets design-wide? Should they be? Or can they be set per-FIRRTL component, or maybe even per (public?) module (doesn't seem to be broken apart in that way, so I guess per-translation-unit)?  If I specialize one component, can I use it with another component that has that option that hasn't been specialized?

As-is I think you could include a `target_Target_A.sv` for component X and `target_Target_B.sv` for component Y (different FIRRTL files) and since all macros are in terms of specific names within each, they wouldn't conflict.  Maybe they should? (or at least let's be sure we're clear on whether this is intentional/supported or explicitly not).

On that, why are the macros per-module-instance (appears to be lowercase of at least the module name?), if they're bulk-enabled all at once?

Thinking out loud a smidge -- so if these just set a single macro (or otherwise manipulate the global environment pre-elaboration) like `define Target A`... the SV for each instance choice would then have effectively its dispatch table locally (vs sunk into/across the Target headers, for better or for worse).

Might be interesting to look at how neat we could make that with some light parameter/enum/generate emission, to perhaps avoid some messier ifdef chain.  Offhand I think we'd have copies of the instantiation statements but I'm not sure that's a big loss and might be more readable when debugging / locally inspectable.

It also makes explicitly that this is closed and must be one of those options, and makes that locally visible.

Hmm.

spec.md
```diff
@@ -4156,6 +4232,11 @@ decl_layer =
     { decl_layer , newline } ,
   dedent ;
 
+decl_option =
```
dtzSiFive, line 4235: Grammar says "option" and "case", FWIW.  firrtl.xml too.

spec.md
```diff
@@ -307,6 +307,40 @@ circuit Foo :
   public module Foo enablelayer A :
 ```
 
+## Targets
```
dtzSiFive, line 310: Maybe it's bikeshedding too much, but could you explain the thinking regarding naming the keyword and word used to describe one dimension of optionality as a "target"?

Target seems like a great and compelling example of the optionality we want to capture, and we might even want to capture different dimensions of targets (maybe), but I kinda like calling these "options" better. "case" works for me too, although on the subject -- I'm not sure they need keywords really vs just listing them like in an enum or something.

"Target" seems to imply a more narrow use/purpose and does not itself imply any optionality, and just to be frank but not argumentative, I found it a bit confusing at first/as a result.

Commonly you specify a target as a known thing your compiler understands and MAYBE it'll produce different code as a result but also maybe not....

Anyway, I don't understand this change, would you mind sparing some words to explain?

spec.md
```diff
@@ -307,6 +307,40 @@ circuit Foo :
...
+
+A `target`{.firrtl} describes one way that a FIRRTL circuit may be specialized for a certain use case.
+Here, specialization means choosing a specific option for a target.
```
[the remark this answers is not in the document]
seldridge: This was inherited from the original proposal. However, it occurred to me that the default could be made explicit.

There's two ABI concerns, one current and one future, that motivate some notion of a default:

1. (Current) A user shouldn't need to do anything to "specialize" for the default. Ergo, the mechanism of enabling the default target should be "no action". With the file-based ABI this would naturally mean "no additional files are included in the build."
2. (Future) Verilog parameters must have a default value.

I do think it's entirely reasonable to roll the default into the target. Part of the changes here would be to decouple the target from instance choice and it therefore seems weird to handle the default (a target concern) with the instance choice.

I think this means either every target has a reserved "default" option or the target indicates its default:

```
target Platform:
  option Generic ; default is first
  option FPGA
  option ASIC

target Platform:
  option FPGA
  option ASIC
  ; There is an implicit option "default" that can be set for any Platform

target Platform = Generic:
  alternative FPGA
  alternative ASIC
```

WDYT syntax-wise?

For these, it would then be easiest to state that either instance choices must be fully specified or that the default is taken for any option not specified:

``` firrtl
instchoice foo, Platform:
  FPGA => Bar
  ASIC => Foo
  default => Foo

instchoice foo, Platform:
  default => Foo
  FPGA => Bar
```

spec.md
seldridge: They're design-wide/bulk-enabled just like layers. They can be set per-component, but only if a FIRRTL circuit is organized to make that possible. E.g., if you want to specialize every instance of a module, you are really describing different modules with different instance choices set by different targets. This is also motivated by the fact that I do not know of a way _in Verilog_ to achieve per-instance specialization. There may be ways to do this that are tool-specific.

It probably needs to be per-public module. However, there are some problems here for a module which has an instance choice which is instantiated by two public modules where one public module instantiates the other. In this situation, you need to give the option of enabling the instance choice from each public module (via two files). However, the two public modules have to agree on the choice.

This could be avoided if public modules were not allowed to instantiated any common submodules. It could also be avoided with generate and parameters at the cost of exposing all instance choices in the design and not allowing easy directory pruning specialization.

spec.md
seldridge: The motivation to call this "target" and not "option" is primarily by the use case. These are intended to allow for specializing the design for some target purpose, feature, technology, etc. FPGA vs. ASIC vs. Generic is closer to the more specific "target" than it is to a generic "option". You can counter argue that "target" doesn't make sense for parametric features that this would enable like cache size.

This is also try to permute the discussion with alternative language in response to confusion like this: https://github.com/chipsalliance/firrtl-spec/pull/155#discussion_r1408358567

I agree that "case" or "option" is unnecessary as these are really just describing a sum type. Any of the following would be fine and don't add a second superfluous keyword:

```
option Target = FPGA, ASIC, Generic
option Target:
  FPGA
  ASIC
  Generic
```

WDYT?

spec.md
seldridge: I updated this to put the module name in the file.

The issue of multiple public modules needing to agree is actually a red herring. While having multiple public modules is possible, a Verilog elaboration is always a tree rooted at a FIRRTL public module. Therefore, I think the only checking that needs to be done is if a macro is multiply defined.

E.g., an error is possible for a design like the following with public modules `Foo`, `Bar`, and `Baz`. Both `Bar` and `Baz` instantiate `Qux` which has an instance choice of `Qux` or `Corge`.

image

The problem here is that there are six specialization files (Cartesian product `[Foo, Bar, Baz] x [Quz, Corge]`). Some of these are compatible and some are incompatible. However, it is easiest to just make all of them incompatible in the ABI as it is likely wrong for such a circuit to ever be including specialization files for any module which is not the root, unless the user has guaranteed that they are using isolated instance graph trees. I.e., if the user wants to specialize `Qux` differently under the `Bar` and `Baz` modules, then `Qux` needs to be duplicated in the instance graph.

This is going to need to be handled carefully in the implementation of the ABI based on how the internal names of defines are chosen.

seldridge:
After some discussion with @rwy7, `option` seems like a better name to describe what is going on. I'll revise this and get this moving.

PR #156, opened 2023-11-29
Update Initialization Coverage spec for zero-width

albertchen-sifive:
Specify that non-zero-width components must be initialized, while zero-width components are implicitly initialized with value zero.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1616,8 +1616,10 @@ module MyModule :
 ```
 
 This is an illegal FIRRTL circuit and an error will be thrown during compilation.
-All wires, memory ports, instance ports, and module ports that can be connected to must be connected to under all conditions.
+All non-zero-width wires, memory ports, instance ports, and module ports that can be connected to must be connected to under all conditions.
 Registers do not need to be connected to under all conditions, as it will keep its previous value if unconnected.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1616,8 +1616,10 @@ module MyModule :
 ```
 
 This is an illegal FIRRTL circuit and an error will be thrown during compilation.
-All wires, memory ports, instance ports, and module ports that can be connected to must be connected to under all conditions.
+All non-zero-width wires, memory ports, instance ports, and module ports that can be connected to must be connected to under all conditions.
```
darthscsi, line 1619: This and the next are ambiguous as it's not clear (at least looking at each sentence in isolation) whether "non-zero-width" only applies to wires or to the entire list.

darthscsi (commented):
How does this rule interact with uninferred widths?  Are those assumed to be non-zero width until proven otherwise?

spec.md
mwachs5: i agree that there seems to be a difference in use case for 0-width elements (SInt, UInt, Analog) and 0-element Aggregates (Vec, bundle types). It seems like users should be responsible for the first one, while the latter does make sense to allow to not happen in the spec. I'm actually not sure what the proposed spec change is currently trying to cover --both of these cases?

PR #160, opened 2024-01-12
[major] Layer-associated probe semantics and ABI

seldridge:
Define semantics and an ABI for layer-associated probes.  ~This introduces a major change to the existing layer ABI where layers are now enabled with a Verilog text macro (`` `define ``) and not by including a file.  This is done to unify the enabling of all layer collateral which can include modules/logic _and_ new probe ports (which were not supported previously).~

This uses a basic ABI where all layer-colored probes are always emitted. We may want to change this in the future to use an ABI where Verilog macros are conditionally defined. However, it is much simpler to start without this. It is also not clear that such an ABI provides much over just requiring a user to enable a layer for the reference to be valid.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -233,6 +233,7 @@ The types of intrinsic module parameters may only be literal integers or string 
 ## Layers
 
 Layers are collections of functionality which will not be present in all executions of a circuit.
+When a layer is enabled, the FIRRTL circuit behaves *as-if* all the optional functionality of that layer was executed.
 Layers are intended to be used to keep verification, debugging, or other collateral, not relevant to the operation of the circuit, in a separate area.
 Each layer can then be optionally included in the resulting design.
 
--- a/spec.md
+++ b/spec.md
@@ -263,14 +263,14 @@ For more information on layer blocks see [@sec:layer-blocks].
 
 Probe types may be colored with a layer.
 Layer coloring indicates which layers must be enabled for a probe to be used.
-For more information on layer-colored probes see [@sec:probe-types] and [@sec:probes-and-layers].
+For more information on layer-colored probes see [@sec:probe-types] and [@sec:layer-coloring].
 
 ### Modules with Enabled Layers
 
 Modules may be declared with enabled layers.
 A module with enabled layers colors the body of the module with the color of all enabled layers.
 This affects the legality of operations which use probes.
-See [@sec:probes-and-layers] for more information on layer coloring.
+See [@sec:layer-coloring] for more information on layer coloring.
 
 To declare a module with layers enabled, use the `enablelayer`{.firrtl} keyword.
 The circuit below shows a module with one layer enabled:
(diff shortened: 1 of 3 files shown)


spec.md
```diff
@@ -260,6 +261,25 @@ circuit Foo :
 Functionality enabled by a layer is put in one or more layer blocks inside modules.
 For more information on layer blocks see [@sec:layer-blocks].
 
+Probe types may be colored with a layer to condition their existence on that layer being enabled.
+For more information on layer-colored probes see [@sec:probe-types] and [@sec:layer-coloring].
```
mmaloney-sf, line 265: The phrase "to condition their existence" sounds a bit existential. The very fabric of reality urges a slight rewording :) 

Perhaps: "Optionally, probe types may be colored with a layer. Layer coloring will indicate which layers must be enabled for a probe to be used."

Alternatively, if we want to resort to stronger language, we can do that too, but we would want to establish a deeper formalism for what that means. That is, we would need a technical term to describe when a probe is "not active" (or "disabled" or "erased" or whatever), and then we can cross-reference that term in the ABI to make that precise in terms of Verilog: that the probe does not exist when it's "not active".

seldridge: It's only problematic if the probes question their existence...

I switched to use a variant of your terse language in a fixup commit: https://github.com/chipsalliance/firrtl-spec/pull/160/commits/774cc117784c283c059426bf704c532a86de0f50

I like the simple "enabled"/"disabled" language. I'd almost prefer something even shorter ("on"/"off"), but we can bikeshed this in a cleanup over the language in the spec.

spec.md
```diff
@@ -2027,19 +2049,16 @@ Since they are intended for verification, ports with a probe type do not necessa
...
-```
-
-A `RWProbe<T>`{.firrtl} will be implicitly cast to a `Probe<T>`{.firrtl} whenever an expression of type `RWProbe<T>`{.firrtl} is assigned to a port with type `Probe<T>`{.firrtl}.
+Probe types may be colored with a layer (see [@sec:layers]).
+A layer-colored probe type places restrictions on the operations which use it.
 
-While `RWProbe<T>`{.firrtl} is strictly more powerful, `Probe<T>`{.firrtl} allows for more aggressive optimization.
-You should consider using `Probe`{.firrtl} whenever you know you don't need to ability to force (see [@sec:forcing]).
+An operation may only "read" from a probe whose layer color is enabled when the operation is enabled.
+An operation that "writes" to a probe must be in a region whose layer color is enabled when the probe's layer color is enabled.
```
seldridge, line 2061: These are the two most important sentences in this PR.

Give these a lot of scrutiny @mmaloney-sf.

mmaloney-sf: Is there a meaningful difference between a layer and a layer color?

My impression is the term "color" is just a colorful way to speak of the association between a probe and the layer it works against.

Both of these sentences also confuse me a bit, trying to interpret them. Trying to dig into why:

* What does it mean for "the operation [to be] enabled"?
* The quotes around "read" and "write" make me wonder if these are formal statements or if they should be taken intuitively.
* You use the word "region" -- Should this be "layer" instead? Eg, "must be in a layer whose color is enabled".
* The phrase "whose color is enabled" is also weird. Intuitively, colors can be mixed, but not enabled or disabled.
* I feel like there's some subtle compound condition in the "write" restriction.

seldridge:
Synced with @mmaloney-sf offline. This is good enough to land. @mmaloney-sf will take over driving this with subsequent PRs by providing more formal semantics around describing the layers as a lattice (with the "colorless" root layer, i.e., the design, defined).

PR #161, opened 2024-01-15
[Proposal] Stronger and more complete reset types

darthscsi:
Make all resets use a uniform type.  Expand that type to cover asynchronous set, synchronous release resets.  Expand that type to handle both active-high and active-low resets.

Provide a couple of ops that allow writing generic code on uninferred reset signals.

Tighten type conversion to make UInt<1> not equivalent to synchronous reset.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -680,15 +680,23 @@ All registers are linked to a clock (see [@sec:registers]).
 ### Reset Types
 
 Once a circuit is powered on, it may require an explicit reset in order to put it into a known state.
-For this, we use a reset type.
-In FIRRTL, we have the option of using both synchronous or asynchronous resets.
+For this, registers take a reset signal of a reset type.  The reset type
+determines the implementation of the registers.
 
-The synchronous reset type is simply a 1-bit unsigned integer: `UInt<1>`{.firrtl}.
-The asynchronous reset type is `AsyncReset`{.firrtl}.
+There are 3 reset behaviors captured by reset type:
+* Synchronous resets (`sync`{.firrtl}) - These are sampled for activation on the rising clock edge.
+* Asynchronous resets (`async`{.firrtl}) - These become active (and inactive) immediately without regard to a clock.
+* Asynchronous activation, synchronous deactivation (`asyncsync`{.firrtl}) - These become active immediate without regard to a clock, but deactivate on the clock edge after the reset is no longer asserted.
+
+There are 2 implementations of the reset signal:
+* Active high (`high`{.firrtl}) - a 1 indicates the reset is active
+* Active low (`low`{.firrtl}) - a 0 indicates the reset is active
+
+The entire cross product of behavior and implementation is valid.
 
 Registers may be declared linked to a reset (see [@sec:registers-with-reset]).
 
-The reset types also have the inferred form: `Reset`{.firrtl} (see [@sec:reset-inference]).
+There are explicit and inferred reset types.  The inferred form is `Reset`{.firrtl} (see [@sec:reset-inference]).  The explicit form is `Reset<Kind,Active>`{.firrtl} where `Kind`{.firrtl} is one of `sync`{.firrtl}, `async`{.firrtl}, or `asyncsync`{.firrtl} and `Active`{.firrtl} is one of `high`{.firrtl} or `low`{.firrtl}.
 
 ### Analog Type
 
@@ -1003,47 +1011,39 @@ The following example shows an inferred reset that will get inferred to a synchr
 ``` firrtl
 input a : UInt<1>
 wire reset : Reset
-connect reset, a
+connect reset, asReset(a, Reset<sync, high>)
 ```
 
-After reset inference, `reset`{.firrtl} is inferred to the synchronous `UInt<1>`{.firrtl} type:
+After reset inference, `reset`{.firrtl} is inferred to the synchronous `Reset<sync, high>`{.firrtl} type:
 
 ``` firrtl
 input a : UInt<1>
-wire reset : UInt<1>
-connect reset, a
-```
-
-The following example demonstrates usage of an asynchronous reset.
-
-``` firrtl
-input clock : Clock
-input reset : AsyncReset
-input x : UInt<8>
-regreset y : UInt<8>, clock, reset, UInt(123)
-; ...
+wire reset : Reset<sync, high>
+connect reset, asReset(a, Reset<sync, high>)
 ```
 
 Inference rules are as follows:
 
 1.  An uninferred reset driven by and/or driving only asynchronous resets will be inferred as asynchronous reset.
-2.  An uninferred reset driven by and/or driving both asynchronous and synchronous resets is an error.
+1.  An uninferred reset driven by and/or driving only asynchronous set, synchronous release resets will be inferred as asynchronous set, synchronous release reset.
+2.  An uninferred reset driven by and/or driving both synchronous resets and any other type of reset is an error.
 3.  Otherwise, the reset is inferred as synchronous (i.e. the uninferred reset is only invalidated or is driven by or drives only synchronous resets).
+4. A reset driven by and/or driving both `high` and `low` shall be an error.
 
-`Reset`{.firrtl}s, whether synchronous or asynchronous, can be cast to other types.
-Casting between reset types is also legal:
+`Reset`{.firrtl}s, whether synchronous or asynchronous, can be converted to other types.
+Converting between reset types is also legal:
 
 ``` firrtl
 input a : UInt<1>
 output y : AsyncReset
 output z : Reset
 wire r : Reset
-connect r, a
+connect r, asReset(a, Reset)
 connect y, asAsyncReset(r)
 connect z, asUInt(y)
 ```
 
-See [@sec:primitive-operations] for more details on casting.
+See [@sec:primitive-operations] for more details on conversion.
 
 ### Probes and Type Inference
 
@@ -1098,10 +1098,7 @@ Similarly, a signed integer type is always equivalent to another signed integer 
 
 Clock types are equivalent to clock types, and are not equivalent to any other type.
 
-An uninferred `Reset`{.firrtl} can be connected to another `Reset`{.firrtl}, `UInt`{.firrtl} of unknown width, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl} (see [@sec:reset-inference]).
-It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
-
-The `AsyncReset`{.firrtl} type can be connected to another `AsyncReset`{.firrtl} or to a `Reset`{.firrtl}.
+The `Reset<y,x>`{.firrtl} type can be connected to another `Reset<y,x>`{.firrtl} or to a `Reset`{.firrtl}.
 
 Two enumeration types are equivalent if both have the same number of variants, and both the enumerations' i'th variants have matching names and equivalent types.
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -680,15 +680,23 @@ All registers are linked to a clock (see [@sec:registers]).
...
-In FIRRTL, we have the option of using both synchronous or asynchronous resets.
+For this, registers take a reset signal of a reset type.  The reset type
+determines the implementation of the registers.
 
-The synchronous reset type is simply a 1-bit unsigned integer: `UInt<1>`{.firrtl}.
-The asynchronous reset type is `AsyncReset`{.firrtl}.
+There are 3 reset behaviors captured by reset type:
+* Synchronous resets (`sync`{.firrtl}) - These are sampled for activation on the rising clock edge.
+* Asynchronous resets (`async`{.firrtl}) - These become active (and inactive) immediately without regard to a clock.
+* Asynchronous activation, synchronous deactivation (`asyncsync`{.firrtl}) - These become active immediate without regard to a clock, but deactivate on the clock edge after the reset is no longer asserted.
```
seldridge, line 689: It may be useful to be very precise here. There's four possibilities:

1. sync assert, sync deassert
2. async assert, sync deassert
3. sync assert, async deassert
4. async assert, async deassert

Only (1) and (2) are useful from a digital design perspective. (4) is useful when modeling the real world (this is a button press) and (3) seems very odd, but might come up in the real world.

Currently, this PR is defining (1), (4), and (2), respectively. Should the PR skip (4) here and only talk about (2)?

tl;dr: I would suggest defining `async` as `asyncsync` and then not mentioning what is called `async` today. My reasoning here is that this is not particularly useful.

The caveat for doing this, which I expect is your motivation here, is that you need what the PR is calling `async` in order to write the actual reset synchronizer circuit (e.g., `rst_n_in` input in this circuit: https://github.com/seldridge/verilog/blob/69b1a206b1ccd8f1ca8cdb6482c59292d6700b9c/src/reset.v).

Given that, this is likely exactly what we want.

spec.md
```diff
@@ -2948,38 +2941,52 @@ The interpret as SInt operation reinterprets e's bits as a signed integer accord
...
+## Interpret as Reset
+
+|  Name    | Arguments  | Parameters  | Arg Types     | Result Type  | Result Width |
+| -------- | ---------- | ----------- | ------------- | ------------ | -------------|
+| asReset  | \(e\)      | \(t\)       | (Reset)       | t            | n/a          |
+|          |            |             | (UInt<1>)     | t            | n/a          |
+|          |            |             | (SInt<1>)     | t            | n/a          |
+|          |            |             | (Clock)       | t            | n/a          |
+
+The result of the interpret as reset operation is an reset typed signal.  This operation doesn't interpret the data based on type, it is a simple cast.  To write general reset operations, use the `setReset`{.firrtl} operation.
```
mwachs5, line 2964: An reset -> a reset

mwachs5 (commented):
This looks great!

One thing to discuss is whether the abi can allow us to put a _n on an active low reset signal once it's lowered to verilog

spec.md
```diff
@@ -680,15 +680,23 @@ All registers are linked to a clock (see [@sec:registers]).
...
 Once a circuit is powered on, it may require an explicit reset in order to put it into a known state.
-For this, we use a reset type.
-In FIRRTL, we have the option of using both synchronous or asynchronous resets.
+For this, registers take a reset signal of a reset type.  The reset type
+determines the implementation of the registers.
 
-The synchronous reset type is simply a 1-bit unsigned integer: `UInt<1>`{.firrtl}.
-The asynchronous reset type is `AsyncReset`{.firrtl}.
+There are 3 reset behaviors captured by reset type:
+* Synchronous resets (`sync`{.firrtl}) - These are sampled for activation on the rising clock edge.
```
mmaloney-sf, line 687: > These are sampled for activation [...].

I've never heard this term "activation" before. Is this a term our general audience would be expected to understand? I'm assuming this refers to when the register samples its input. Is that right?

spec.md
```diff
@@ -680,15 +680,23 @@ All registers are linked to a clock (see [@sec:registers]).
...
+* Asynchronous resets (`async`{.firrtl}) - These become active (and inactive) immediately without regard to a clock.
+* Asynchronous activation, synchronous deactivation (`asyncsync`{.firrtl}) - These become active immediate without regard to a clock, but deactivate on the clock edge after the reset is no longer asserted.
+
+There are 2 implementations of the reset signal:
+* Active high (`high`{.firrtl}) - a 1 indicates the reset is active
+* Active low (`low`{.firrtl}) - a 0 indicates the reset is active
+
+The entire cross product of behavior and implementation is valid.
```
mmaloney-sf, line 695: "Reset is active" is a phrase worth giving more of a definition to. 

> The entire cross product of behavior and implementation is valid.

Phasing. I would avoid the technical term "cross product" in favor of something more pedestrian: "all possible combinations". Given an example of the first few combinations might also help, since I can't quite tell, at a glance, what I'm taking the cross product over.

spec.md
```diff
@@ -1003,47 +1011,39 @@ The following example shows an inferred reset that will get inferred to a synchr
 ``` firrtl
 input a : UInt<1>
 wire reset : Reset
-connect reset, a
+connect reset, asReset(a, Reset<sync, high>)
```
mmaloney-sf, line 1014: This change is fine. But if we move forward with this change, remind me to make a pass against all of our examples to clean them up using a native `Reset<_, _>` type rather than casting `UInt<1>`s.

spec.md
```diff
@@ -1098,10 +1098,7 @@ Similarly, a signed integer type is always equivalent to another signed integer
 
 Clock types are equivalent to clock types, and are not equivalent to any other type.
 
-An uninferred `Reset`{.firrtl} can be connected to another `Reset`{.firrtl}, `UInt`{.firrtl} of unknown width, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl} (see [@sec:reset-inference]).
-It cannot be connected to both a `UInt`{.firrtl} and an `AsyncReset`{.firrtl}.
-
-The `AsyncReset`{.firrtl} type can be connected to another `AsyncReset`{.firrtl} or to a `Reset`{.firrtl}.
+The `Reset<y,x>`{.firrtl} type can be connected to another `Reset<y,x>`{.firrtl} or to a `Reset`{.firrtl}.
```
mmaloney-sf, line 1101: I'm sure you had a reason, but why is it `y, x` instead of `x, y`?

spec.md
```diff
@@ -1409,17 +1406,17 @@ reg myreg: SInt, myclock
 A register with a reset is declared using `regreset`{.firrtl}.
 A `regreset`{.firrtl} adds two expressions after the type and clock arguments: a reset signal and a reset value.
 The register's value is updated with the reset value when the reset is asserted.
-The reset signal must be a `Reset`{.firrtl}, `UInt<1>`{.firrtl}, or `AsyncReset`{.firrtl}, and the type of initialization value must be equivalent to the declared type of the register (see [@sec:type-equivalence] for details).
-If the reset signal is an `AsyncReset`{.firrtl}, then the reset value must be a constant type.
+The reset signal must be a `Reset`{.firrtl} or `Reset<_,_>`{.firrtl}, and the type of initialization value must be equivalent to the declared type of the register (see [@sec:type-equivalence] for details).
```
mmaloney-sf, line 1409: I think this is the first time the `_` notation has been used.

Are we treating this as a metanotation? There's no underscore syntax intended here, right?

We may want to smooth this over at some point in the prose.

spec.md
```diff
@@ -1409,17 +1406,17 @@ reg myreg: SInt, myclock
...
+If the reset signal is an `Reset<async,_>`{.firrtl} or an `Reset<asyncsync,_>`{.firrtl}, then the reset value must be a constant type.
```
mmaloney-sf, line 1410: It would be good to include a one-sentence justification for why this is required.

spec.md
```diff
@@ -3572,10 +3581,13 @@ linecol = digit_dec , { digit_dec } , ":" , digit_dec , { digit_dec } ;
 
 (* Tokens: PrimOp Keywords *)
 primop_1expr_keyword =
-    "asUInt" | "asSInt" | "asClock" | "asAsyncReset" | "cvt"
+    "testRest" | "asUInt" | "asSInt" | "asClock" | "cvt"
```
mmaloney-sf, line 3584: Where does `testRest` come from? Is that supposed to be `testReset`?

spec.md
```diff
@@ -2948,38 +2941,52 @@ The interpret as SInt operation reinterprets e's bits as a signed integer accord
...
+|  Name     | Arguments  | Arg Types  | Result Type  | Result Width |
+| --------- | ---------- | ---------- | ------------ | -------------|
+| testReset | \(e\)      | (Reset)    | (UInt<1>)    | 1            |
+
+This operation returns a 1 if a reset is active and a 0 if a reset is inactive regardless of the type of the reset signal.
+
+This allows writing generic reset handling code by decoupling the reset implementation from activation logic.  This operation is different from `asUInt`{.firrtl} in that it always returns `1`{.firrtl} for active resets regardless of the underlying implmentation values.
+
+The behavior of a circuit testing an asynchronous reset are unspecified due to the lack of
+an execution model for FIRRTL in the presence of signal transitions between clock
```
mmaloney-sf, line 2988: I fully agree with this. But it's also kind of telling on ourselves :)

I think it's sufficient for us to just say, "the behavior is unspecified". We can then leave the execution model as an aspiration.

spec.md
```diff
@@ -3489,17 +3496,18 @@ type_hardware =
...
 type_ground_width =
     "UInt" , [ width ]
   | "SInt" , [ width ]
-  | "Analog" , [ width ] ;
+  | "Analog" , [ width ] 
+  | "Reset" , [ resetspec ] ;
 
 width = "<" , int , ">" ;
 
+resetspec = "<" , ( "sync" | "async" | "syncasync" ) ,  "," , ( "high | "low" ) , ">" ;
```
mmaloney-sf, line 3509: Could we name the categories for these parameters and introduce two simple nonterminals?

For instance:

```
resetspec = "<" , reset_synchronicity ,  "," , reset_activation , ">" ;

reset_synchronicity = "sync" | "async" | "syncasync" ;

reset_activation = "high | "low" ;
```

mmaloney-sf:
I love this.

I added a few comments, mostly in terms of the presentation.

spec.md
jackkoenig: I agree with Schuyler here as well.

spec.md
```diff
@@ -1003,47 +1011,39 @@ The following example shows an inferred reset that will get inferred to a synchr
...
 ```
 
 Inference rules are as follows:
 
 1.  An uninferred reset driven by and/or driving only asynchronous resets will be inferred as asynchronous reset.
-2.  An uninferred reset driven by and/or driving both asynchronous and synchronous resets is an error.
+1.  An uninferred reset driven by and/or driving only asynchronous set, synchronous release resets will be inferred as asynchronous set, synchronous release reset.
+2.  An uninferred reset driven by and/or driving both synchronous resets and any other type of reset is an error.
 3.  Otherwise, the reset is inferred as synchronous (i.e. the uninferred reset is only invalidated or is driven by or drives only synchronous resets).
+4. A reset driven by and/or driving both `high` and `low` shall be an error.
```
jackkoenig, line 1031: I think we need a rule for what happens when you have a reset driven with an explicit `Kind` but no explicit `Active`. I think it should default to active high.

spec.md
```diff
@@ -1003,47 +1011,39 @@ The following example shows an inferred reset that will get inferred to a synchr
...
 output z : Reset
 wire r : Reset
-connect r, a
+connect r, asReset(a, Reset)
 connect y, asAsyncReset(r)
 connect z, asUInt(y)
 ```
 
-See [@sec:primitive-operations] for more details on casting.
+See [@sec:primitive-operations] for more details on conversion.
```
jackkoenig, line 1046: This terminology was changed to "convert" but `asReset` below is called a "cast". I think we should keep the `cast` terminology here and add an example of converting with `setReset`.

jackkoenig (commented):
This looks great!

It would be really useful to have an example or two of generic reset handling code. In particular we need ways to invert and do some simple boolean operations (eg. OR) resets generically where inverting a reset should change the `Active` while keeping the kind--I'd just like to understand how to build these operations using the existing primitives. In particular, if I have a `Reset<_, _>` I need to invert it and get back effectively `Reset<_, !_>` in a way that reset inference can still fill in the holes.

spec.md
darthscsi: simplified.

spec.md
darthscsi: While this would need to happen, there is a transition period where we will likely want to codegen casts different (or accept them for a while).

spec.md
darthscsi: I was intending underscore syntax.  But I missed that in the grammar.

PR #162, opened 2024-01-19
[patch] Define Storable types

mmaloney-sf:
# Storable Types
This PR defines a new term "Storable Type" which captures the class of FIRRTL types which are admissible inside of registers and memories.

## Isn't this just like Passive?
Yes.

The definition has a large overlap with the term Passive Type. And that is not a coincidence.

The term Passive Type has been used to designate 'types without flips'. However, with the introduction of Probe Types and also Property types, more nuance is required to handle types. Arguments have been made both for and against the above types being allowed be passive.

In order to help resolve this tension, the new term Storable Type is intended to separate out one particular usage of Passive Types.

## What is different between Passive and Storable?

Some notable differences between Passive and Storable:

1. Clocks and Resets are Passive but not Storable
1. Analogs are Passive but not Storable.
1. Properties are not Storable
1. Probes are not Storable 
1. Any structural part of a type being `const` disqualifies it from being Storable.

No. 1 acknowledges the special nature of Clocks and Resets.

I would insist on No. 2 on the grounds that `Analog` isn't very well-specified. We can always add it in later once we have a better foundation for it.

No. 3 and No. 4 are the primary motivation. Properties are easy to exclude, as we have stipulated they are not to affect the RTL, but rather serve as metadata to it. For probes, we use a similar appeal to the non-"hardware-y-ness" of probes, combined with the fact that storing a probe in a register requires a nontrivial encoding (and ultimately is likely to wind up meaningless).

No. 5 is a lesson from @darthscsi (feel free to forward attribution if needed) who pointed out to me that you can use `const` registers to create a mutating `const` value. This PR removes the non-`const` restriction from registers, as it is subsumed into this definition. (The spec seems to have forgotten to include a `const` restriction on memories).

## Why doesn't this PR remove the term Passive?

The term Passive still exists after this PR goes through. It still requires a bit of investigation to work out what role passivity plays in FIRRTL. There are only a handful of places in the spec Passive Types are referenced:

* Inside the definition itself (11 times out of 18)
* Rule 3 of the Connection Rules
  * Either the flow of the right-hand side expression is source or duplex, or the right-hand side expression has a **passive type**.
* One incidental time in an example about probes (can probably be removed safely)
* Talking about non-passive force targets (2 times)
* Talking about multiplexers
* Once talking about the types of probes
* Once in the definition of enumeration types

This gives us only a handful of places where we need to apply scrutiny. But none of these cases is so clear cut I feel I could change or remove them without changing the interpretation of the spec.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -785,7 +785,7 @@ Enumerations are disjoint union types.
 
 Each enumeration consists of a set of variants.
 Each variant is named with a tag.
-Each variant also has a type associated with it which must be connectable (see [@sec:connectable-types]), storable (see [@sec:storable-types]), passive (see [@sec:passive-types]).
+Each variant also has a type associated with it which must be connectable (see [@sec:connectable-types]) and passive (see [@sec:passive-types]).
 
 In the following example, the first variant has the tag `a`{.firrtl} with type `UInt<8>`{.firrtl}, and the second variant has the tag `b`{.firrtl} with type `UInt<16>`{.firrtl}.
 
@@ -878,7 +878,7 @@ A **storable type** is a type which may appear as the type of a register or the 
 A storable type is defined recursively:
 
 -   All non-`const`{.firrtl} integer types are storable.
--   All non-`const`{.firrtl} enumeration types are storable.
+-   A non-`const`{.firrtl} enumeration types are storable if and only if the type of each of its variants is storable.
 -   A non-`const`{.firrtl} vector type is storable if and only if the element type is storable.
 -   A non-`const`{.firrtl} bundle type is storable if and only if it contains no field which is marked `flip`{.firrtl} and the type of each field is storable.
 
--- a/spec.md
+++ b/spec.md
@@ -870,26 +870,31 @@ A connectable type is defined recursively:
 -   bundles where each field type is a connectable type, or
 -   an enumeration type
 
+## Storable Types
+
+Stateful elements, such as registers and memories, are restricted in what types they can store.
+A **storable type** is a type which may appear as the type of a register or the element type of a memory.
+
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -785,7 +785,7 @@ Enumerations are disjoint union types.
 
 Each enumeration consists of a set of variants.
 Each variant is named with a tag.
-Each variant also has a type associated with it which must be connectable (see [@sec:connectable-types]) and passive (see [@sec:passive-types]).
+Each variant also has a type associated with it which must be connectable (see [@sec:connectable-types]), storable (see [@sec:storable-types]), passive (see [@sec:passive-types]).
```
mwachs5, line 788: I am not sure about this one. I'm not sure what the intended restrictions on enums are. Are we trying to keep Properties and Probes out?

mwachs5: Well, I guess they are already excluded by the requirement on `connectable`. I am not sure why enums would care about being storable? Unless we would want to change the definition below to say say "an Enum is storable if all of its variants are storable"?

mmaloney-sf: Good point. I could relax this to say that enums are storable when the type of each variant is storable.

mmaloney-sf: Changed.

mwachs5:
Can you please add to the version notes and also note what you think this is for the spec. Is this patch or minor? I think patch since the intention is not to actually change any allowed behavior ?

azidar (approved):
This distinction makes sense to me, thanks! I agree with @mwachs5 that this should probably be included in the release notes.

mmaloney-sf:
Updated the release notes.

Marked this as `[patch]`.

PR #164, opened 2024-01-23
[major] Change the minimum width of the result of "Shift Right" for UInt to 0-bit

SpriteOvO:
Change the minimum width of the result of "Shift Right" to 0-bit to align the behavior between Chisel and CIRCT.

Context: https://github.com/chipsalliance/chisel/pull/3752#issuecomment-1905086931, https://github.com/chipsalliance/chisel/pull/3752#issuecomment-1906225248 and chipsalliance/chisel#3735.

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -13,6 +13,7 @@ revisionHistory:
       - Reorganized statements section.
       - Rewrite of the Types section.
       - Fix mistake in layer-associated probe grammar.
+      - Change the minimum width of the result of "Shift Right" on UInt to 0-bit.
     abi:
       - Add ABI for public modules and filelist output.
       - Changed ABI for group and ref generated files.
--- a/spec.md
+++ b/spec.md
@@ -2999,7 +2999,7 @@ n must be non-negative.
   --------------------------------------------------------------------------
   Name   Arguments   Parameters   Arg Types   Result Type   Result Width
   ------ ----------- ------------ ----------- ------------- ----------------
-  shr    \(e\)       \(n\)        (UInt)      UInt          max(w~e~-n, 1)
+  shr    \(e\)       \(n\)        (UInt)      UInt          max(w~e~-n, 0)
 
                                   (SInt)      SInt          max(w~e~-n, 1)
   --------------------------------------------------------------------------


dtzSiFive:
The text that follows says:

> If n is greater than or equal to the bit-width of e, the resulting value will be zero for unsigned types and the sign bit for signed types.
n must be non-negative.

Which suggests maybe the table should still have `1` for the `SInt` case?

Consider:

```
circuit SIntShr:
  module SIntShr:
    input in: SInt<5>
    output out: SInt
    out <= shr(in, 8)
```

Which feeding through `firtool` produces:

```
// Generated by CIRCT firtool-1.63.0g20240122_ce27439
module SIntShr(
  input  [4:0] in,
  output       out
);

  assign out = in[4];
endmodule
```

jackkoenig:
Good catch @dtzSiFive, I agree. The analogy with `tail` is good, but `tail` on an `SInt` returns `UInt` whereas `shr` on an `SInt` returns `SInt` and as such should preserve the sign bit.

revision-history.yaml
```diff
@@ -13,6 +13,7 @@ revisionHistory:
       - Reorganized statements section.
       - Rewrite of the Types section.
       - Fix mistake in layer-associated probe grammar.
+      - Change the minimum width of the result of "Shift Right" to 0-bit.
```
jackkoenig, line 16:
```suggestion
      - Change the minimum width of the result of "Shift Right" on UInt to 0-bit.
```

spec.md
```diff
@@ -2999,9 +2999,9 @@ n must be non-negative.
   --------------------------------------------------------------------------
   Name   Arguments   Parameters   Arg Types   Result Type   Result Width
   ------ ----------- ------------ ----------- ------------- ----------------
-  shr    \(e\)       \(n\)        (UInt)      UInt          max(w~e~-n, 1)
+  shr    \(e\)       \(n\)        (UInt)      UInt          max(w~e~-n, 0)
 
-                                  (SInt)      SInt          max(w~e~-n, 1)
+                                  (SInt)      SInt          max(w~e~-n, 0)
```
jackkoenig, line 3004:
```suggestion
                                  (SInt)      SInt          max(w~e~-n, 1)
```

jackkoenig (changes requested):
After offline discussion with @darthscsi, @seldridge, and @dtzSiFive, we think the best course of action is to change just UInt to have a minimum width of 0 and keep SInt with a minimum width of 1 to maintain mathematical soundness with what it means to do an arithmetic shift to a signed number.

SpriteOvO:
> we think the best course of action is to change just UInt to have a minimum width of 0 and keep SInt with a minimum width of 1 to maintain mathematical soundness with what it means to do an arithmetic shift to a signed number.

Makes sense to me, changes applied.

jackkoenig (approved):

seldridge (commented):
Thanks for working with us to figure out the right thing to do here. 👍

PR #166, opened 2024-01-30
[major] Allow assert message to be a format string

uenoku:
This commit allow assert statement's message string to be a format string which is corrently expressed as printf-encoded assertion. CIRCT side PR is https://github.com/llvm/circt/pull/6621 

This also adds missing grammar for verif statements.

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -15,6 +15,7 @@ revisionHistory:
       - Fix mistake in layer-associated probe grammar.
       - Define "Storable Type".
       - Change the minimum width of the result of "Shift Right" on UInt to 0-bit.
+      - Allow assert and assume statements to have a format string
     abi:
       - Add ABI for public modules and filelist output.
       - Changed ABI for group and ref generated files.
--- a/spec.md
+++ b/spec.md
@@ -1957,7 +1957,7 @@ Format strings support the following escape characters:
 
 To facilitate simulation, model checking and formal methods, there are three non-synthesizable verification statements available: assert, assume and cover.
 Assert and assume statements require a clock signal, a predicate signal, an enable signal, a format string and a variable list of argument signals.
-Cover statement requires a clock signal, a predicate signal, an enable signal, a format string and a string literal.
+Cover statement requires a clock signal, a predicate signal, an enable signal, and a string literal.
 The predicate and enable signals must have single bit unsigned integer type.
 Assert and assume statement may print the format string as an explanatory message where its argument placeholders are substituted.
 See [@sec:format-strings] for information about format strings.
--- a/spec.md
+++ b/spec.md
@@ -1956,9 +1956,11 @@ Format strings support the following escape characters:
 ## Verification
 
 To facilitate simulation, model checking and formal methods, there are three non-synthesizable verification statements available: assert, assume and cover.
-Each type of verification statement requires a clock signal, a predicate signal, an enable signal and a string literal.
+Assert and assume statements require a clock signal, a predicate signal, an enable signal, a format string and a variable list of argument signals.
+Cover statement requires a clock signal, a predicate signal, an enable signal, a format string and a string literal.
 The predicate and enable signals must have single bit unsigned integer type.
-Assert and assume use the string as an explanatory message.
+Assert and assume statement may print the format string as an explanatory message where its argument placeholders are substituted.
+See [@sec:format-strings] for information about format strings.
 For cover statements the string indicates a suggested comment.
 When an assert or assume is violated the explanatory message may be issued as guidance.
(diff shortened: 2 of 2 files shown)


revision-history.yaml
```diff
@@ -15,6 +15,7 @@ revisionHistory:
       - Fix mistake in layer-associated probe grammar.
       - Define "Storable Type".
       - Change the minimum width of the result of "Shift Right" on UInt to 0-bit.
+      - Allow assert stmt to have a format string
```
mwachs5, line 18:
```suggestion
      - Allow assert, assume, cover statements to have a format string
```

spec.md
```diff
@@ -3341,6 +3341,25 @@ command =
...
+        expr , "," ,
+        expr , "," ,
+        string_dq
+    , ")" ,
+  | "cover" , "(" ,
+        expr , "," ,
+        expr , "," ,
+        expr , "," ,
+        string_dq
+    , ")" ,
```
jackkoenig, line 3362: What does a format String on a cover mean? What is a typical or lowering to SystemVerilog?

uenoku: `string_dq` doesn't indicate a format string. A format string is not currently specially handled in the grammar. `{ "," , expr }` represents operands for substitutions.

seldridge (approved):
LGTM

spec.md
```diff
@@ -3341,6 +3343,25 @@ command =
...
+        expr , "," ,
+        string_dq ,
+        { "," , expr }
+    , ")" ,
+  | "assume" , "(" ,
+        expr , "," ,
+        expr , "," ,
+        expr , "," ,
+        string_dq
+    , ")" ,
```
seldridge, line 3358: I'm good with adding assume. The grammar needs an update to reflect this.

spec.md
```diff
@@ -1977,14 +1977,15 @@ However it can never be used in a reference since it is not of any valid type.
 
 The assert statement verifies that the predicate is true on the rising edge of any clock cycle when the enable is true.
 In other words, it verifies that enable implies predicate.
+When the predicate is false, the assert statement prints out the format string where its argument placeholders are substituted.
```
dtzSiFive, line 1980: L1959 describes the operands of all verification statements and refers to this as "string literal" and does not mention the new additional operands, please align.

(it does mention the message (without format-string behavior naturally) may be printed for the assert/assume cases already, FYI -- but address however you see fit just please keep these sections on same page please).

The language currently is guarded by "may".

jackkoenig:
Unless someone more clever than me has an idea on how this can be backwards compatible, I think this is a major change since assert and assume messages containing `%` will now be rejected

dtzSiFive:
> Unless someone more clever than me has an idea on how this can be backwards compatible, I think this is a major change since assert and assume messages containing `%` will now be rejected

If they contain invalid format strings (or missing arguments for the otherwise valid string) there will likely be problems and possibly rejected by tools.

This already is the case and has been for some time, FWIW, arguably this is updating the spec to reflect reality and to allow specifying operands for what's already a format string in practice.
..or implementation has been wrong for some time but even with that framing FWIW this is unlikely to break anything as a result.

But focusing solely on the FIRRTL spec what you say makes sense, but I don't think there's any utility in ensuring or worrying about backwards compatibility since in practice this isn't something folks are able to rely on today.  So probably can rest easy on that front at least :+1: .

Example:

```
circuit Assert:
  module Assert:
    input clock: Clock
    input cond: UInt<1>
    input enable: UInt<1>

    assert(clock, cond, enable, "Testing: %d")
```
->
```
// Generated by CIRCT firtool-1.63.0g20240129_8e3d29f
module Assert(
  input clock,
        cond,
        enable
);

  always @(posedge clock) begin
    if (enable)
      assert(cond) else $error("Testing: %d");
  end // always @(posedge)
endmodule
```

Which if you consult your SV spec or feed through, e.g., verilator, will complain.

spec.md
```diff
@@ -1956,9 +1956,11 @@ Format strings support the following escape characters:
 ## Verification
 
 To facilitate simulation, model checking and formal methods, there are three non-synthesizable verification statements available: assert, assume and cover.
-Each type of verification statement requires a clock signal, a predicate signal, an enable signal and a string literal.
+Assert and assume statements require a clock signal, a predicate signal, an enable signal, a format string and a variable list of argument signals.
+Cover statement requires a clock signal, a predicate signal, an enable signal, a format string and a string literal.
```
dtzSiFive, line 1960:
```suggestion
Cover statement requires a clock signal, a predicate signal, an enable signal, and a string literal.
```

uenoku:
Could someone merge this? Thanks!

PR #168, opened 2024-02-09
[minor] Add Property primitive operations, starting with integer addition.

mikeurbach:
This adds prose to the Expressions section introducing the concept of
primitive Property operations, adds a new section defining the
primitive Property operations, and adds primitive Property operations
to the grammar.

As a concrete example, integer addition is included.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2450,6 +2450,26 @@ asClock(x)
 
 [@sec:primitive-operations] will describe the format and semantics of each primitive operation.
 
+## Primitive Property Operations {#sec:expressions:primitive-property-operations}
+
+All fundamental operations on property types are expressed as a FIRRTL primitive operation.
+In general, each operation takes some number of property type expressions as arguments.
+
+The general form of a primitive property operation is expressed as follows:
+
+```firrtl
+op(arg0, arg1, ..., argn)
+```
+
+The following examples of primitive property operations demonstrate adding two property expressions, `a`{.firrtl} and `b`{.firrtl}, and adding an integer property literal to expression `a`{.firrtl}.
+
+```firrtl
+integer_add(a, b)
+integer_add(a, Integer(2))
+```
+
+[@sec:primitive-property-operations] will describe the format and semantics of each primitive property operation.
+
 ## Reading Probe References
 
 Probes are read using the `read`{.firrtl} operation.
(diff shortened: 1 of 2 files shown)


mikeurbach:
I tried to follow very closely the language, structure, and grammar that we are using when describing the other primitive operations on ground types. Please let me know what you think.

spec.md
```diff
@@ -3459,6 +3498,9 @@ primop_1expr     = primop_1expr_keyword , "(" , expr , ")" ;
 primop_1expr1int = primop_1expr1int_keyword , "(", expr , "," , int , ")" ;
 primop_1expr2int = primop_1expr2int_keyword , "(" , expr , "," , int , "," , int , ")" ;
 
+(* Primitive Property Operations *)
+property_primop_2expr = property_primop_2expr_keyword , "(" , property_expr , "," , property_expr ")" ;
```
mikeurbach, line 3502: Hmm in the rendered PDF this overflows... I will try breaking this differently.

seldridge: Yes... the grammar is verbatim and you have to do some manual reflowing to get things to fit. There are other examples of manual breaking in the grammar where this happens.

mwachs5 (approved):
lgtm!

spec.md
```diff
@@ -2450,6 +2450,26 @@ asClock(x)
...
+All fundamental operations on property types are expressed as a FIRRTL primitive operation.
+In general, each operation takes some number of property type expressions as arguments.
+
+The general form of a primitive property operation is expressed as follows:
+
+```firrtl
+op(arg0, arg1, ..., argn)
+```
+
+The following examples of primitive property operations demonstrate adding two property expressions, `a`{.firrtl} and `b`{.firrtl}, and adding an integer property literal to expression `a`{.firrtl}.
```
seldridge, line 2464: Nit: adding the `{.firrtl}` hint to the syntax highlighter doesn't change the format. It's usually cleaner to drop it. This is _not_ the case for `Integer` later, though. That does add useful highlighting.

spec.md
```diff
@@ -2450,6 +2450,26 @@ asClock(x)
...
+## Primitive Property Operations {#sec:expressions:primitive-property-operations}
+
+All fundamental operations on property types are expressed as a FIRRTL primitive operation.
+In general, each operation takes some number of property type expressions as arguments.
+
+The general form of a primitive property operation is expressed as follows:
+
+```firrtl
+op(arg0, arg1, ..., argn)
+```
```
seldridge, line 2462: Nit: Pandoc should want to reformat this to include a space:

```suggestion
``` firrtl
op(arg0, arg1, ..., argn)
```
```

File: spec.md [hunk 4]
```diff
@@ -2450,6 +2450,26 @@ asClock(x)
 
 [@sec:primitive-operations] will describe the format and semantics of each primitive operation.
 
+## Primitive Property Operations {#sec:expressions:primitive-property-operations}
```
seldridge, line 2453: It's kind of unfortunate that the PR has to go through this just because two different subsections have the same name. I should see if there is a pandoc option to do this automatically.

seldridge (approved):
LGTM

This is beginning to motivate prefixes on the hardware operations, e.g., `hw_add`. 🧐

spec.md
mikeurbach: Updated slightly, this is what I went with.
Screen Shot 2024-02-08 at 5 19 07 PM

Open to other indentation ideas.

spec.md
mikeurbach: Yeah, I just sort of followed how HW primitives do it--both naming the subsection the same as the section, and using these anchors to disambiguate. If there is a better option with pandoc, happy to use that.

mikeurbach:
> This is beginning to motivate prefixes on the hardware operations, e.g., hw_add. 🧐

Hmm, yeah maybe. I don't know if we need to update the original primitives, but I think the type prefix is good for the new ones (e.g., integer_add, list_create, etc.).

PR #172, opened 2024-02-12
[nfc] Add ability to test code snippets

seldridge:
This demonstrates a flow that can be used to validate that all code
snippets used in the FIRRTL specification correctly parse.  This is
implemented using a simple pandoc filter to extract the bodies of all code
blocks with known languages (FIRRTL and Verilog), write them to the build
directory, and then run them all through an appropriate tool based on file
extension (firtool or verilator).

This may be a little cumbersome as every code example has to be completely
legal.  To alleviate this, this PR includes the ability to add special
delimiters, indicated by lines with the text `snippetbegin`/`snippetend` which
will cause the filter to extract only the lines between the delimiters.

CC: @dtzSiFive

### DIFF
--- a/scripts/extract-firrtl-code.lua
+++ b/scripts/extract-firrtl-code.lua
@@ -1,0 +1,76 @@
+index = 0
+
+-- Return the indices of all strings in `text` that are between `snippetbegin`
+-- and `snippetend` strings.
+function getSnippets (text)
+   local foundSnippet = false
+   local i = 0
+   local start
+   local snippets = {}
+   while true do
+      if start == nil then
+         _, i = string.find(text, "snippetbegin.-\n", i + 1)
+         if i == nil then
+            break
+         end
+         -- Increment by one to drop the newline from the snippet.
+         start = i + 1
+         -- Decrement 'i' so that there is _always_ a newline for the match in
+         -- the else block to use.  This makes it possible to match emtpy
+         -- snippets easily as they look exactly like non-empty snippets.
+         i = i - 1
+      else
+         e, i = string.find(text, "\n[^\n]-snippetend", i + 1)
+         if i == nil then
+            error("missing 'snippetend'")
+         end
+         table.insert(snippets, {i = start, e = e - 1})
+         start = nil
+      end
+   end
+   return snippets
+end
+
+function CodeBlock(elem)
+   local extMap = {
+      firrtl = "fir",
+      verilog = "v",
+      systemverilog = "sv"
+   }
+   local ext = extMap[elem.classes[1]]
+   local skip = elem.classes[2] == "notest"
+
+   -- Early exit if we don't know a file extension for this code block.
+   if (not ext) or skip then
+      return
+   end
+
+   -- Write the entire code block to a file.
+   local filename = string.format("build/%s-code-example-%03d.%s", PANDOC_STATE.input_files[1], index, ext)
+   index = index + 1
+
+   local f = io.open(filename, 'w')
+   f:write(elem.text)
+   f:write("\n")
+   f:close()
+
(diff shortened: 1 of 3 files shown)


seldridge:
This now includes basic delimited support. The magic delimiters `snippetbegin`/`snippetend` can be used to extract out any number of code regions. E.g., this allows you to do things like:

```
FIRRTL version 4.0.0
circuit Foo:
  module Foo:
    input a: UInt<1>
    output b: UInt<1>

    // snippetbegin
    connect b, a
    // snippetend
```

This will extract the entire code block to a file. It will replace the original code block with:

```
    connect b, a
```

scripts/extract-firrtl-code.lua
```diff
@@ -0,0 +1,68 @@
...
+            if string.len(newText) == 0 then
+               newText = new
+            else
+               newText = newText .. "\n" .. new
+            end
+            text = string.sub(text, jj + 1, -1)
+         else
+            foundSnippet = false
+         end
+      end
```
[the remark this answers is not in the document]
seldridge: I rewrote this after some better understanding of lua. Good suggestion on moving to a function. Changes pushed that has the rewrite and conversion to a function.

mmaloney-sf:
Just to be sure, the line comment syntax is `; ...` in FIRRTL and not `// ...` right? (Your example in the PR looks like it will error out. I don't know if that's material to the logic, though).

mwachs5 (commented):
could we add a few sentences to CONTRIBUTING.md that explains this a little bit?

seldridge:
> Just to be sure, the line comment syntax is ; ... in FIRRTL and not // ... right? (Your example in the PR looks like it will error out. I don't know if that's material to the logic, though).

Yes! That's a mistake. The filter will still handle that, but that is illegal FIRRTL. The intent is that the code should put the `snippetbegin`/`snippetend` delimiters inside a comment.

scripts/extract-firrtl-code.lua
```diff
@@ -0,0 +1,76 @@
...
+      return
+   end
+
+   -- Write the entire code block to a file.
+   local filename = string.format("build/%s-code-example-%03d.%s", PANDOC_STATE.input_files[1], index, ext)
+   index = index + 1
+
+   local f = io.open(filename, 'w')
+   f:write(elem.text)
+   f:write("\n")
```
seldridge, line 54: This newline helps for readability and avoids a Verilator lint warning that we don't have to waive.

scripts/extract-firrtl-code.lua
```diff
@@ -0,0 +1,76 @@
...
+end
+
+function CodeBlock(elem)
+   local extMap = {
+      firrtl = "fir",
+      verilog = "v",
+      systemverilog = "sv"
+   }
+   local ext = extMap[elem.classes[1]]
+   local skip = extMap[elem.classes[2]] == notest
```
seldridge, line 41: This allows opting out from testing for a specific snippet using syntax like `{.firrtl .notest}` on the code block. This is very helpful for things that are really annoying to test like Verilog that includes input files or FIRRTL code blocks that are intentionally not legal through the use of pseudo-type parameters (see memories).

seldridge (commented):
This is good for a final review. I will open a second PR which fixes all the code examples. We can then turn this on in CI.

seldridge:
@mwachs5 wrote:
> could we add a few sentences to CONTRIBUTING.md that explains this a little bit?

Added in the force push!

mikeurbach (approved):
This is my first exposure to pandoc filters, so I don't have any strong comments on the lua implementation. The usage and goal makes sense to me in general.

PR #181, opened 2024-02-21
[major] No abstract reset on extmodule.

dtzSiFive:
Align extmodule port requirements re:abstract reset to match those on "public".

### DIFF
--- a/revision-history.yaml
+++ b/revision-history.yaml
@@ -32,6 +32,7 @@ revisionHistory:
       - Remove intrinsic modules.
       - Allow layerbocks anywhere in a module.
       - Add language clarifying behavior statements affected by conditionals.
+      - No abstract reset on externally-defined modules.
     abi:
       - Add ABI for public modules and filelist output.
       - Changed ABI for group and ref generated files.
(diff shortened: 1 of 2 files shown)


seldridge (approved):
LGTM

Please add tests/issues to Chisel or x-fail tests to CIRCT as I don't think this is currently rejected.

Chisel doesn't currently enforce and CIRCT doesn't check it.

CC: @jackkoenig

seldridge:
This one has users and, after talking with @dtzSiFive, it would be better to hold off on this for 4.0.0 so we can figure out how to get Chisel to stop emitting this.

darthscsi:
> so we can figure out how to get Chisel to stop emitting this.

Is it Chisel or code bases using it which do this?  If it's just code-bases and there is nothing in Chisel forcing this, then that's an easier transition.

jackkoenig:
> Is it Chisel or code bases using it which do this? If it's just code-bases and there is nothing in Chisel forcing this, then that's an easier transition.

Chisel does **not** do this, the only things that get abstract reset automatically in Chisel are normal Modules (i.e. not extmodules) that are not the top module.

mmaloney-sf (approved):

PR #182, opened 2024-02-22
[patch] Add enablelayer to grammar.

dtzSiFive:


dtzSiFive:
Any suggestions for org welcome!

spec.md
```diff
@@ -4358,6 +4359,9 @@ property_primop_2expr = property_primop_2expr_keyword ,
 (* Tokens: Annotations *)
 annotations = "%" , "[" , json_array , "]" ;
 
+(* Enable Layer *)
+enablelayer = "enablelayer" , id , { "." , id };
```
mmaloney-sf, line 4363: Could we introduce a new production rule for colorings?

`coloring := id , { "." , id } ;`

Also, to be consistent with the other rules, let's put a space before the semicolon.

spec.md
```diff
@@ -4358,6 +4359,9 @@ property_primop_2expr = property_primop_2expr_keyword ,
...
+
```
seldridge, line 4364: If this is inlined above, does that fit on one line still? (Or, can it be made to split with line splitting?)

spec.md
dtzSiFive: Oops dropped the space when reworking, good catch ty!

spec.md
dtzSiFive: I'll check.  I broke it out to not have to wrap the line from what it had already, but since I did I'll give that a go :+1:.

mmaloney-sf: Inlining it would lead to doubly-nested `{` `}`'s. 

It might be good to leave this split out for that reason.

seldridge (approved):
LGTM

spec.md
mmaloney-sf: More practically, some parser generators would not accept the inlined version because of the doubly-nested `{` `}`.

seldridge: I'm fine with splitting it out. @mmaloney-sf: Do you know where this should live? It was previously under the "token" section.  Is that appropriate or could you propose an alternative?

mmaloney-sf: For nonterminals which are only used once, it's probably good to keep them near their usage point.

But it's easier for us to shuffle definitions than to refactor the grammar, so anywhere is fine by me.

dtzSiFive: Great, wasn't sure how to balance against the idea to sink less-important things later (e.g., `annotations`).

But yes easy to rearrange as we see fit at any point.

Thanks for your input all!

PR #183, opened 2024-02-22
[major] Drop input probes from spec.

dtzSiFive:
Not used in practice and have restrictions that they must be basically equivalent to a design that doesn't use them at all.


seldridge (approved):
LGTM

In case you didn't notice, `make format` will renumber the itemized lists for you.

mmaloney-sf:
Simple as. LGTM.

mmaloney-sf (approved):

PR #185, opened 2024-02-23
[major] Main module is public.

dtzSiFive:
Clarify main module is implicitly/always public.

Update examples to specify widths (and concrete reset type).

Some of these are invalid regardless (e.g., "input foo : UInt" on main module) but recent CIRCT checks this now while parsing.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -77,6 +77,7 @@ Each module is a hardware "unit" that has ports, registers, wires, and may insta
 
 Each FIRRTL circuit must have exactly one *main module*.
 The main module is a module that has the same name as the FIRRTL circuit.
+The main module must be public (see [@sec:public-modules]).
 
 The following circuit contains a FIRRTL circuit with two modules.
 `Foo` is the main module:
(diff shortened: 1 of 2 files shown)


dtzSiFive:
TODO: Make "public" keyword required on main modules.

spec.md
```diff
@@ -121,6 +121,8 @@ A public module has a stable interface allowing it to be instantiated by *other*
 A public module will always be physically present in a compiled circuit.
 In effect, public modules are the exported identifiers of a circuit.
 
+The main module of a circuit must be public.
```
seldridge, line 124: This language would be better placed where the main module is discussed, after line 79:

``` diff
Each FIRRTL circuit must have exactly one *main module*.
The main module is a module that has the same name as the FIRRTL circuit.
+ The main module must be public (see [@sec:public-modules]).
```

seldridge (approved):
LGTM

spec.md
dtzSiFive: Good call, updated thanks!

PR #192, opened 2024-03-11
Fix some bugs introduced in #136

darthscsi:
LHS and RHS are swapped in the flow "intuition" section.  Further, the flow section no longer describes how flow propagates.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1211,7 +1211,7 @@ If the field is flipped, then its flow is the reverse of the flow of the bundle-
 The reverse of source is sink, and vice-versa.
 The reverse of duplex remains duplex.
 
-The flow of all other expressions are source, including mux and cast. 
+The flow of all other expressions are source, including mux and cast.
 
 ## Type Equivalence
 
--- a/spec.md
+++ b/spec.md
@@ -1212,7 +1212,14 @@ Ports are always considered from the perspective of "inside the module".
 Moreover, input ports may only appear "on the right side" of a connect, while output ports may only appear on either side of a connect.
 Finally, while submodules instances and memories are strictly sources, they interact with the sub-field rule in such a way that their input ports are sinks.
 
-Subindex, subaccess, and subfield preserve flow from argument to result.  All other primitive operations produce source flows, including mux and cast. 
+The flow of a sub-index or sub-access expression is the flow of the vector-typed expression it indexes or accesses.
+The flow of a sub-field expression depends upon the orientation of the field.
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -1197,21 +1197,23 @@ A duplex expression is an expression that is both a source and sink.
...
 -   Memories are sources.
 
 Here are a few comments to help with intuition:
-Nodes may only appear "on the left side" of a connect.
-Wires and registers may appear "on either side of a connect statement".
+Nodes may only appear on the right side of a connect.
+Wires and registers may appear on either side of a connect statement.
 Ports are always considered from the perspective of "inside the module".
-Moreover, input ports may only appear "on the left side" of a connect, while output ports may only appear "on the right side" of a connect.
+Moreover, input ports may only appear "on the right side" of a connect, while output ports may only appear on either side of a connect.
```
seldridge, line 1212: This is more correct:

```suggestion
Moreover, input ports may only appear "on the right side" of a connect, while output ports may appear on either side of a connect.
```

However, it's not really correct because an input port _can_ appear on the either side of a connect depending on the flow. If this is redefining input port to mean "input scalar ports", then this is true.

seldridge (approved):
This was pretty busted. Thanks for fixing it.

One problem with the port section is that this section is conflating "input port" with "port with source flow". I don't know what to do with it. However, I'll accept this PR as-is because it is more correct with the patch. 🙃

Please add an entry to `revision-history.yaml` and update the commit/PR title to indicate `[patch]`.

seldridge:
Pandoc formatting is unhappy with two-space after period. 😉

spec.md
```diff
@@ -1195,22 +1195,24 @@ A source expression supplies a signal and can be used to drive a circuit compone
...
-The rules for the flow of an expression are as follows:
+The rules for the flow of an expression are as follows.
 
-If the expression is a reference, we look at the kind of the circuit component:
+If the expression is an identifier, we look at the kind of the circuit component the identifier refers to:
 
 -   Nodes are sources.
 -   Wires and registers are duplex.
--   For ports, `input` ports are sources and `output` ports are sinks.
+-   For ports, `input` ports are sources and `output` ports are duplex.
```
seldridge, line 1204: This may not be correct as it conflates direction with flow. An output port swaps the flow, but an input port doesn't. 

What I think this means is that the list of rules later should indicate that "if a port has output direction, then its flow is reversed" and "if a port has input direction, then its flow is not reversed."

Issue #196, opened 2024-03-26
`eq`/`neq` comparison operators missing from table


programmerjake:
https://github.com/chipsalliance/firrtl-spec/blob/e28204d2176788909efc8405696e68efd78f54e3/spec.md?plain=1#L3657-L3661

seldridge:
👍 Thanks for the report!

seldridge:
Fixed in https://github.com/chipsalliance/firrtl-spec/commit/0e15e532d2c47d84f00daba5a15b0fc4b94b8711.

PR #198, opened 2024-03-26
Add eq, neq to operations table.

mmaloney-sf:
Responding to #196.

Looking closer at the table, I'm not sure that the data is represented in an entirely consistent way. But I can't imagine this will cause confusion.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -3656,8 +3656,9 @@ Together with the divide operator, the modulus operator satisfies the relationsh
 
 | Name   | Arguments | Parameters | Arg Types   | Result Type | Result Width |
 |--------|-----------|------------|-------------|-------------|--------------|
-| lt,leq |           |            | (UInt,UInt) | UInt        | 1            |
-| gt,geq | (e1,e2)   | ()         | (SInt,SInt) | UInt        | 1            |
+| lt,leq |  (e1,e2)  | ()         | (UInt,UInt) | UInt        | 1            |
+| gt,geq |           |            | (SInt,SInt) | UInt        | 1            |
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -3656,8 +3656,9 @@ Together with the divide operator, the modulus operator satisfies the relationsh
 
 | Name   | Arguments | Parameters | Arg Types   | Result Type | Result Width |
 |--------|-----------|------------|-------------|-------------|--------------|
-| lt,leq |           |            | (UInt,UInt) | UInt        | 1            |
-| gt,geq | (e1,e2)   | ()         | (SInt,SInt) | UInt        | 1            |
+| lt,leq | (e1,e2)   | ()         | (UInt,UInt) | UInt        | 1            |
+| gt,geq |           |            | (SInt,SInt) | UInt        | 1            |
+| eq,neq |           |            |             |             |              |
```
seldridge, line 3661: This might be more consistent (and would require formatting):

```suggestion
| lt,leq,gt,geq,eq,neq | (e1,e2)   | ()         | (UInt,UInt) | UInt        | 1            |
|                      |           |            | (SInt,SInt) | UInt        | 1            |
```

seldridge (approved):
I was confused reading this with it on multiple lines. Maybe we just clean it up and put all the operations on one line?

spec.md
mmaloney-sf: Unfortunately, this doesn't fit in the rendered version.

mmaloney-sf: I made a compromise, grouping them in pairs. It looks passable:
image

Issue #197, opened 2024-03-26
Missing specification of how bit numbers are determined


programmerjake:
the FIRRTL specification doesn't specify how to determine if `bits(UInt<8>(0hF0), 3, 0)` returns `0hF` or `0h0` -- specifying this is needed because some significant specifications (the [PowerISA specification](https://files.openpower.foundation/s/dAYSdGzTfW4j2r2) in particular) counts bits starting from 0 at the MSB end and counting up to the LSB end, and most others count bits starting from 0 at the LSB end, though I expect some start counting at 1.

mmaloney-sf:
Thanks for the report.

To be sure, we label the LSB as bit `0` and the MSB as `width - 1`. The `bits` primop is inclusive on both sides.

That's interesting to hear that PowerISA does it as it does. We can add verbiage to make the convention we use explicit.

mmaloney-sf:
Just as a note to myself (or anyone who snipes the issue from me), this note is *probably* best given as a language-wide convention, rather than a footnote on `bits` (although we should crossreference it wherever we make use of it).

darthscsi:
> the [PowerISA specification](https://files.openpower.foundation/s/dAYSdGzTfW4j2r2) in particular

And to be clear, this is a separate issue from endian, which refers to the storage (in memory) ordering of Bytes of a word.  In-register operations do not depend on the endian.

darthscsi:
Which is to say, please don't use "endian' to describe the bit ordering in firrtl.  Also, 1.3.2 of the Power spec for those curious to check and it's weird when numerically equivalent words of different sizes have their common bits in different numerical positions.

mmaloney-sf:
I was under the impression that the One True meaning of "endian" was from which end you ate your boiled eggs.

PR #200, opened 2024-03-27
Bits index convention

mmaloney-sf:
Responding to #197, this PR adds a sentence explaining that we use a big-endian representation for values, with the index of the LSB as 0 and the MSB as one less than the length.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -3832,6 +3832,7 @@ The result of the concatenate operation is the bits of e1 concatenated to the mo
 The result of the bit extraction operation are the bits of e between lo (inclusive) and hi (inclusive).
 hi must be greater than or equal to lo.
 Both hi and lo must be non-negative and strictly less than the bit width of e.
+The index of the least significant bit is 0 and the index of the most significant bit is one less than the width of the argument.
 
 ## Head
 


mwachs5:
this seems to include changes from another PR?

seldridge (approved):
Approving the last commit.

This appears to be stacked on https://github.com/chipsalliance/firrtl-spec/pull/198.

mmaloney-sf:
Oops.

Fixed and force-pushed.

programmerjake:
> Responding to #197, this PR adds a sentence explaining that we use a big-endian representation

I would call the LSB being bit zero little endian, unrelated to byte endianness.

also, iirc you wanted to have this be a language-wide convention, not just on the bits function.

mmaloney-sf:
> I would call the LSB being bit zero little-endian or LSB0, unrelated to byte endianness.

I meant it more generally as bit endianness. The language I used in the PR should hopefully suffice.

> also, iirc you wanted to have this be a language-wide convention, not just on the bits function.

I considered this, but it looked fairly specific to `bits`. (Although let me know if I overlooked another primop that used this). I'll keep this in the back of my mind and hoist it up if needed.

PR #202, opened 2024-04-10
[minor] Add ABI to public module, extmodule

### LINKED ISSUE #171: Add Lowering Convention to External, Public Modules
Both external and public modules need to specify their port lowering convention. Currently, this is underspecified and left up to a compiler to implement (with annotations or a global option).

Some constraints:

- The syntax should be similar for both external and public modules.
- This should by optional in the syntax with a default convention (v1) assumed if not defined. This provides backwards compatibility with the current behavior.
- The syntax shouldn't cause problems or confusion when trying to implement parameters later.
- `enablelayer` currently goes after the module name

The following are some syntax ideas:

### String after "public"

```
public v2 module Foo:
```

This is slightly problematic as it is incongruent with external modules which don't have the public keyword. We could add the public keyword to them:

```
public v1 extmodule Bar:
```

### Use a string that starts with `v` after the module name:

```
public module Foo v2 enablelayer A:
extmodule Bar v1:
```

### Use brackets, parentheses, or something like that

```
public module<v2> Foo:
extmodule<v1> Bar:
```

I expect that parameters will eventually show up as `Foo<x: Int>` type syntax, so this may not conflict too badly.

### Make the convention port-like

```
public module Foo:
  convention v2
extmodule Bar:
  convention v1
```

### Use an attribute dictionary-like syntax

```
public module Foo enablelayer A {convention = v2}:
extmodule Bar {convention = v1}:

### PR DESCRIPTION
**Pull request by seldridge**
Add a way to specify the port lowering ABI of a public module or an external module.  This uses syntax like the following:

    public module<v1> Foo:

    extmodule<v2> Bar:

This includes language that indicates that the "v1" ABI is used if no ABI is specified.  This makes this change fully backwards compatible with older FIRRTL versions (and thereby a "minor" change).

Fixes #171.

### CODE REVIEW

**Comment by seldridge** [comment]
Also, note that this is intended to _not_ conflict with a future syntax for parameters or type parameters which would look similar, but would be located after the module name and not placed on the module keyword. Spitballing:

``` firrtl
public module<v2> Foo<x: Int, y: String>:
```

darthscsi (approved):
Do instances need to specify this?  I don't think so, the module declarations should be enough.

seldridge:
> Do instances need to specify this?

They do not. It would also create some complications if they did as a single public module could then have different conventions requiring using the convention as part of the resulting module's name (this would affect the ABI of the module name and not just expectations around its ports).

mmaloney-sf:
Absolute bikeshed: For the syntax, could we add a Java/Scala-style annotation syntax?

Eg:

```firrtl
@abi("v1")
public module Foo:

@abi("v2")
extmodule Bar:
```

PR #207, opened 2024-04-24
[nfc] fix: overflowing text, highlighting, typos

mbty:
This PR fixes several minor presentation issues and don't impact the contents of the specification:
* Fix broken syntax highlighting for "data-type" on p.17 and extmodule Foo's parameters on p.13
* Fix several instances of overflowing lines (literal code)
* Fix some typos: "Arthmetic", missing commas in grammar

However, changes to the syntax highlighting have some consequences that reach beyond their immediate objectives:
* "-" is now classified as a weakDeliminator, i.e. it doesn't act as a word delimiter anymore (for p.13). See https://docs.kde.org/stable5/en/kate/katepart/highlight.html
* "`" is never detected as a string delimiter anymore (for p.17)

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -506,12 +506,10 @@ circuit Foo:
     ;; snippetend
 ```
 
-The type of a memory is a bundle type derived from the declaration (see [@sec:memories]).
+The type of a memory is a bundle type derived from the declaration.
 
 The type named in `data-type`{.firrtl} must be storable (see [@sec:storable-types]).
 It indicates the type of the data being stored inside of the memory.
-
-See [@sec:memories] for more details.
 
 ## Subcomponents
 
@@ -2150,7 +2148,8 @@ The port is driven by the clock signal in the `clk`{.firrtl} field.
 Finally, the readwrite ports have type:
 
 ``` {.firrtl .notest}
-{addr: UInt<N>, en: UInt<1>, clk: Clock, flip rdata: T, wmode: UInt<1>, wdata: T, wmask: M}
+{addr: UInt<N>, en: UInt<1>, clk: Clock, flip rdata: T, wmode: UInt<1>,
+ wdata: T, wmask: M}
 ```
 
 A readwrite port is a single port that, on a given cycle, can be used either as a read or a write port.
@@ -2265,7 +2264,9 @@ circuit Foo:
     wire cond: UInt<1>
     wire a: UInt
     wire b: UInt
-    printf(clk, cond, "a in hex: %x, b in decimal:%d.\n", a, b) : optional_name
+    printf(
+      clk, cond, "a in hex: %x, b in decimal:%d.\n", a, b
+    ) : optional_name
     ;; snippetend
 ```
 
@@ -2336,7 +2337,9 @@ circuit Foo:
     wire en: UInt<1>
     connect pred, eq(X, Y)
     connect en, Z_valid
-    assert(clk, pred, en, "X equals Y when Z is valid but got X=%d Y=%d", X, Y) : optional_name
+    assert(
+      clk, pred, en, "X equals Y when Z is valid but got X=%d Y=%d", X, Y
+    ) : optional_name
     ;; snippetend
 ```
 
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -2412,12 +2418,15 @@ Module `Foo` contains a layer block that creates a node computed from a port def
...
-  layer Bar, bind:       ; Declaration of layer Bar with convention "bind"
+  ; Declaration of layer Bar with convention "bind"
+  layer Bar, bind:
 
   public module Foo:
     input a: UInt<1>
 
-    layerblock Bar:      ; Declaration of a layer block associated with layer Bar inside module Foo
+    ; Declaration of a layer block associated with layer Bar inside module
+    ; Foo
```
mmaloney-sf, line 2428: Could you manually wrap at "inside module Foo" instead of at "Foo" to avoid a [typographic orphan](https://en.wikipedia.org/wiki/Widows_and_orphans)?

spec.md
```diff
@@ -3927,7 +3938,7 @@ The arguments of all primitive property operations must be expressions with prop
 Each specific operation can place additional restrictions on the number and types of their arguments.
 In general, primitive property operations are named with a prefix indicating the property type on which they operate.
 
-## Integer Arthmetic
+## Integer Arithmetic
```
mmaloney-sf, line 3941: Yikes! Thanks for this catch :)

spec.md
```diff
@@ -4215,7 +4226,8 @@ decl_layer =
 
 decl_type_alias = "type", id, "=", type ;
 
-port = ( "input" | "output" ) , id , ":" , (type | type_property) , [ info ] ;
+port = ( "input" | "output" ) , id , ":" , (type | type_property) ,
+       [ info ] ;
 type_param = int | string_dq | string_sq ;
```
mmaloney-sf, line 4231: Please revert this and all other changes to the grammar.

These do not fall off the page in the PDF. But I laid the grammar out manually so that the newlines in the grammar section are suggestive of where the `newline` tokens appear.

If you think I missed something regarding these changes, please let me know.

spec.md
```diff
@@ -506,13 +506,11 @@ circuit Foo:
     ;; snippetend
 ```
 
-The type of a memory is a bundle type derived from the declaration (see [@sec:memories]).
+The type of a memory is a bundle type derived from the declaration.
```
mmaloney-sf, line 509: Good catch.

spec.md
```diff
@@ -2361,7 +2364,10 @@ circuit Foo:
     wire en: UInt<1>
     connect pred, eq(X, Y)
     connect en, Z_valid
-    assume(clk, pred, en, "X equals Y when Z is valid but got X=%d Y=%d", X, Y) : optional_name
+    assume(
+      clk, pred, en, "X equals Y when Z is valid but got X=%d Y=%d", X,
```
mmaloney-sf, line 2368: While I'd rather the example just fit on one line, this is better than before.

Thanks!

mmaloney-sf (commented):
Thanks a lot.

My only ask if that we revert the changes to the Grammar section.

spec.md
mbty: I reverted most changes as per your request, with the exception of the addition of a missing literal comma in the definition of reg.

The very end of `primop_1expr2int`'s definition does fall entirely off the page, however.

To be clear, I originally formatted this section so that everything fit in the standard width of text, as indicated by the horizontal rule under the pages' headers (and many definitions did indeed not quite fit). However, I can fully understand why making the position of newline tokens suggestive takes precedence over this concern.

spec.md
```diff
@@ -4241,7 +4252,7 @@ circuit_component_wire = "wire" , id , ":" , type , [ info ] ;
 circuit_component_inst = "inst" , id , "of" , id , [ info ] ;
 
 circuit_component_reg =
-    "reg" , id , ":" , type , expr , [ info ]
+    "reg" , id , ":" , type , "," , expr , [ info ]
```
mmaloney-sf, line 4255: Also good catch.

spec.md
mmaloney-sf: Thank you!

PR #215, opened 2024-06-21
[patch] Clarify existing when behavior

seldridge:
Add language that defines what conditionals are (they are a region where the operations in the region change their behavior based on the condition) and how different operations are affected by the condition.  This fills a hole where the behavior of commands was not explicitly specified.  Add an aside that clarifies that this means that a _trivial_ inlining will not work as a user might expect.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1623,7 +1623,6 @@ Conditional statements define one or more regions, each with an associated condi
 These regions contain other statements (including nested conditional statements).
 The behavior of the contained statements is affected by the conditions associated to the regions containing them.
 
-
 ## Conditional Statements
 
 FIRRTL provides several kinds of conditional statements.
@@ -1828,7 +1827,6 @@ circuit Foo:
     ;; snippetend
 ```
 
-
 ## Conditional Execution
 
 Statements that appear in a conditional region behave differently based on the type of statement and on the conditions of the blocks containing it.
@@ -1847,7 +1845,6 @@ For hardware component declarations, conditional regions have no effect.
 > A register in a conditional region and an instantiation in a conditional region where the instantiated module contains a register will execute the same after a trivial inlining.
 > A command in a conditional region and a module instantiation in a conditional region where the instantiated module contains a command will not execute the same after a trivial inlining.
 > In this latter case, the command is not conditional before trivial inlining and conditional after trivial inlining.
-
 
 ## Initialization Coverage
 
@@ -1880,7 +1877,6 @@ Circuit components declared within condition regions may only be referred to wit
 > The behavior is also a bit non-intuitive at first.
 > This differs from the behavior in most programming languages, where variables in a local scope can shadow variables declared outside that scope.
 > Additionally, while the names *exist* in the module's namespace, those names cannot be *used* outside of the region which they are declared.
-
 
 ## Conditional Last Connect Semantics
 
--- a/spec.md
+++ b/spec.md
@@ -228,7 +228,8 @@ endmodule
 ## Layers
 
 Layers are collections of functionality which will not be present in all executions of a circuit.
-When a layer is enabled, the FIRRTL circuit behaves *as-if* all the optional functionality of that layer was executed.
+When a layer is enabled, the FIRRTL circuit behaves *as-if* all the optional functionality of that layer was included in the normal execution of the circuit.
+When a layer is disabled, the circuit behaves *as-if* all the optional functionality of that layer was removed from the execution of the circuit.
 Layers are intended to be used to keep verification, debugging, or other collateral, not relevant to the operation of the circuit, in a separate area.
 Each layer can then be optionally included in the resulting design.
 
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -1616,23 +1617,34 @@ A register is initialized with an indeterminate value (see [@sec:indeterminate-v
...
--   Connections (see [@sec:connections]) only drive their sink values when the product (logical and) of all conditions that are parents of the connect, but not parents of the sink declaration evaluates to true.
--   Commands (see [@sec:commands]) only evaluate when the product (logical and) of all parent conditions evaluates to true.
--   The evaluation of all other operations in conditional regions is unaffected by the condition.
-    E.g., register, wire, and node declarations as well as module instantiation statements and primitive operation expressions may occur in conditional regions but never evaluate conditionally.
+## Conditional Execution
+
+The *conditional execution* of operations in the region of a conditional statement is operation dependent:
+
+-   Connections (see [@sec:connections]) only execute when the product (logical and) of all conditions that are parents of the connect, but not parents of the sink declaration evaluates to true.
+-   Commands (see [@sec:commands]) only execute when the product (logical and) of all parent conditions evaluates to true.
```
mmaloney-sf, line 1628: I'm not sure what you mean by "all" conditions, since `when` statements only have a single condition.

Your use of the term "parents" here suggests in my mind you're trying to define a new notion on the spot: "a **parent** (`when`/condition?) of a statement is any (`when`/condition of a `when` statement) which appears as the ancestor in the abstract syntax tree of the given statement." -- And even this is too cramped a definition.

spec.md
```diff
@@ -1616,23 +1617,34 @@ A register is initialized with an indeterminate value (see [@sec:indeterminate-v
...
-Conditional statements define one or more regions where the execution of operations (declarations, statements, and expressions) inside the regions is augmented by a *condition*.
-The evaluation behavior of operations is operation-dependent:
+Conditional statements define one or more regions where the operations (declarations, statements, and expressions) inside a region are *conditionally executed*.
+I.e., the operations within the region execute depending on the value of the condition and the kind of operation.
 
--   Connections (see [@sec:connections]) only drive their sink values when the product (logical and) of all conditions that are parents of the connect, but not parents of the sink declaration evaluates to true.
--   Commands (see [@sec:commands]) only evaluate when the product (logical and) of all parent conditions evaluates to true.
--   The evaluation of all other operations in conditional regions is unaffected by the condition.
-    E.g., register, wire, and node declarations as well as module instantiation statements and primitive operation expressions may occur in conditional regions but never evaluate conditionally.
+## Conditional Execution
```
mmaloney-sf, line 1623: This section is probably good to have. I'm not sure that the initial version quite captures it.

If I have a module:

```firrtl
circuit Top :
  public module Top :
    output out : UInt<1>
    when UInt(1) :
      out := UInt(0) ; first connect
    else :
      out := UInt(1) ; second connect
```

Let's look at the first connect. The first connect is a conditional statement. There's only one parent of the connect (`UInt(1)`) and its product evaluates to true. The sink is `out` and it has no parents, so the product is vacuously true.

Combining these two, we see that the connect "only executes when `UInt(1)` is true, (which it always is), so this connect is always executing.

But what does that mean? 

* Are we assuming an event model, and the event constantly fires?
* Are we assuming an event model, and the event fires once, since it true at the start of the simulation but is a constant with "no sensitivity list"?
* Are we assuming this happens every clock cycle? (There's no clock, so this isn't it).

spec.md
```diff
@@ -1616,23 +1617,34 @@ A register is initialized with an indeterminate value (see [@sec:indeterminate-v
 
 # Conditionals
 
-Conditional statements define one or more regions where the execution of operations (declarations, statements, and expressions) inside the regions is augmented by a *condition*.
-The evaluation behavior of operations is operation-dependent:
+Conditional statements define one or more regions where the operations (declarations, statements, and expressions) inside a region are *conditionally executed*.
```
mmaloney-sf, line 1620: Bikeshed, but I'm still firmly in the camp that what we're doing is better named "simulation" and not "execution."

Concurrency notwithstanding, "execution" rings to my ears as something inherently sequential. (Maybe it's just because the phrase "to execute a program" is burned into my brain).

Execution may play a part in the simulation (eg, a multi-step task is scheduled, like a procedural block in Verilog). But if I take a circuit diagram, plugged a number on each of its inputs, and computed the resulting outputs, I don't think of that as "executing the circuit."

spec.md
```diff
@@ -1616,23 +1617,34 @@ A register is initialized with an indeterminate value (see [@sec:indeterminate-v
...
-## When Statements
+## Conditional Statements
+
+FIRRTL provides several types of conditional statements.
+
+### When Statements
 
-Connections within a when statement that connect to previously declared components hold only when the given condition is high.
-The condition must have a 1-bit unsigned integer type.
+When statements define a condition and two conditional regions: an "if" region and an "else" region.
```
mmaloney-sf, line 1643: The term "if" region seems a bit weird, since we call them `when` statements.

Also, you're calling them "regions", but wouldn't "branches" be more conventional?

mmaloney-sf:
Also, I realized I missed the implication that "trivial inlining is not possible."

What's an example of this? (Just in the comments here is fine).

I can imagine inlining a module definition's body wouldn't work. But I just chalk that up to declarations-under-whens is just a bad idea.

spec.md
seldridge: This is trying to define how the following works. It may conflate the notion of a parent region from a parent condition. The notion of regions and parents is pretty standard in LLVM and MLIR. (I'm not attempting to define something non-standard on the spot!):

``` firrtl
when a:
  wire x: UInt<1>
  when b:
    when c:
      connect x, y
```

This has three regions and associated conditions: one for a, one for b, and one for c. a is a parent of b and b is a parent of c. This is then trying to explain that the connection to `x` is conditioned on `b` and `c`, but not `a`. It may be more reasonable to not define it this way and instead say that the connection to `x` is _not_ conditional with respect to any parent conditions of `x`.

WDYT?

How would you describe this?

seldridge:
> What's an example of this? (Just in the comments here is fine).

Wires and registers are fine. However, any command is not. Consider:

``` firrtl
FIRRTL version 4.0.0
circuit Foo:
  module Bar:
    input clock: Clock
    printf(clock, UInt<1>, "hello")
  public module Foo:
    input clock: Clock
    input cond: UInt<1>

    when cond:
      inst bar of Bar
      connect bar.clock, clock
```

If this is trivially inlined, you get:

``` firrtl
FIRRTL version 4.0.0
circuit Foo:
  public module Foo:
    input clock: Clock
    input cond: UInt<1>

    when cond:
      printf(clock, UInt<1>, "hello")
```

The problem is that before the trivial inlining, the printf was unconditional. However, after the trivial inlining, then the printf is conditional.

This doesn't seem really problematic. Instead, it just means that inlining (or outlining) needs to be handled correctly and the above trivial inlining is not correct.

spec.md
seldridge: This is a good question, though I'm not sure how much that this can be rectified on this PR as this seems to want a complete simulation / execution model of FIRRTL. I do think this is needed, though.

I think I'd try to define this as falling back to the definition of what connect means. The connect "is conditionally executed" (or evaluated, or simulated). What this conditional execution means should be left to the definition of the connect (which is not defined).

Taking a stab at this... what this means is that the connect changes the value of `out`. What "changes" mean and when it happens is a function of the destination. Wires are instantaneous and continuous. Registers are only updated on a clock. This pedantically _shouldn't_ be defined in terms of sensitivity lists, though it is is similar to how those work.

spec.md
seldridge: The FIRRTL Dialect term here is "then" and "else". How does that sound?

seldridge: I'm leaning on "region" here as it's the more general term. E.g., match statements have a number of regions that are all conditioned on a condition. Layer blocks also have regions. Modules really have a single region, too.

There's prior art here where MLIR's SCF (structured control flow) actually calls them "regions" for it's `if` operation: https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfif-scfifop

seldridge: Changed to "then" and "else" region language here: https://github.com/chipsalliance/firrtl-spec/pull/215/commits/f0f62dfa1cd9a6fb6be83dab9c202a8dc25f3021

spec.md
seldridge: I don't know. The reason why I've been partial to "execute" is that "simulate" seems to narrowly be limited to "simulation" (which seems to pedantically not be what happens if you go and put this FIRRTL on an FPGA). Another alternative could be "evaluate".

That said, this matches the existing language of "execute" which is already in the layers and command sections. Is it fine to punt on this to fix in a follow-on PR on-mass if we decide to change?

spec.md
seldridge: I've added clarifications of what a "parent condition" is in https://github.com/chipsalliance/firrtl-spec/pull/215/commits/052f1ebf01f19f3efcedfcec229d5256af646a40.

spec.md
```diff
@@ -1616,12 +1617,37 @@ A register is initialized with an indeterminate value (see [@sec:indeterminate-v
 
 # Conditionals
 
-Several statements provide branching in the data-flow and conditional control of verification constructs.
+Conditional statements define one or more regions where the operations (declarations, statements, and expressions) inside these regions *conditionally execute* based on the value of a condition.
```
mmaloney-sf, line 1620: > Each conditional statement defines a regions. Statements and expressions inside this region execute conditionally based on the value of an expression. Declarations are also allowed inside of conditional statements, but the resulting hardware component always exists.

Trying to figure out how to suggest an edit, but I fail. This verbiage is a little more exact for this sentence.

mmaloney-sf (approved):
I pushed a few changes. Especially to the section on Conditional Execution.

PR #216, opened 2024-06-27
[patch] Allow layerblocks anywhere in a module.

dtzSiFive:
Support added for CIRCT here: https://github.com/llvm/circt/pull/7234 .


mmaloney-sf:
For context, is this hinting at layer blocks under whens? Are there other situations we have our attention on as well?

dtzSiFive:
> For context, is this hinting at layer blocks under whens? Are there other situations we have our attention on as well?

Yep, that's what it's doing, sorry for not specifying that more clearly!  Also, similarly, within `match` statements.

I don't know of other situations, attention or otherwise.  LMK if have any concerns, please! 👍

PR #218, opened 2024-06-28
[nfc] Replace mentions of region with block for consistency.

dtzSiFive:
These mean same thing presently, and are used interchangeably. Standardize on the more commonly used term.


mmaloney-sf:
As I've mentioned in DMs:

I don't make the distinction in my head. I find "block" slightly more natural in the FIRRTL spec. But ultimately, I can go either way.

seldridge (commented):
LGTM

PR #226, opened 2024-08-09
[minor] Add Property type and operations for List.

mikeurbach:
This defines the List property type, and primitive operations to construct a List and concatenate Lists.

### DIFF
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -36,6 +36,7 @@
       <item>Reset</item>
       <item>AsyncReset</item>
       <item>Integer</item>
+      <item>List</item>
     </list>
     <list name="outertype">
       <item>Probe</item>
--- a/include/firrtl.xml
+++ b/include/firrtl.xml
@@ -36,11 +36,11 @@
       <item>Reset</item>
       <item>AsyncReset</item>
       <item>Integer</item>
-      <item>List</item>
     </list>
     <list name="outertype">
       <item>Probe</item>
       <item>RWProbe</item>
+      <item>List</item>
     </list>
     <list name="typemodifiers">
       <item>const</item>
--- a/spec.md
+++ b/spec.md
@@ -900,6 +900,20 @@ circuit Example:
   ;; snippetend
 ```
 
+### List Type
+
+The `List` type represents an ordered sequence of properties.
+It is parameterized by the type of its elements, which can be any legal Property type.
+
+``` firrtl
+FIRRTL version 4.0.0
+circuit Example:
+  ;; snippetbegin
+  public module Example:
+    input listProp : List<Integer> ; an input port of List property type
+  ;; snippetend
+```
+
 ## Connectable Types
 
 A **connectable type** is one which may be the type of expressions which may participate in the `connect`{.firrtl} statement.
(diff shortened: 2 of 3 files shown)


spec.md
```diff
@@ -3938,6 +3952,27 @@ The shift right operation result is the arbitrary precision signed integer arith
 e2 sign bits from e1 are shifted into the most significant bits, and the e2 least significant bits of e1 are truncated.
 e2 must be non-negative.
 
+## List Operations
+
+List operations create `List`{.firrtl} property type expressions from other property expressions.
```
mmaloney-sf, line 3957: Formally speaking, what is a "property expression"?

spec.md
```diff
@@ -3938,6 +3952,27 @@ The shift right operation result is the arbitrary precision signed integer arith
...
+
+### List Construction Operation
+
+  Name        Arguments   Arg Types   Result Type
+  ----------- ----------- ----------- -------------
+  List\<t\>   (e\*)       (t\*)       List\<t\>
```
mmaloney-sf, line 3963: This section breaks a lot of invariants previously held by primops. Is it even appropriate to categorize it as one?

For example, you gave `List<t>` special syntax, thanks to the generic. Every other primop is signified with a simple keyword. Is `List` the primop? Or are the primops `List<Integer>`, `List<List<Integer>>`, `List<String>`, etc.

No other previous primops in FIRRTL have been variadic. In fact, it's so *un-variadic* that is encodes the arity in the grammar. If we're going to be adding variadics (which there's an argument to be made we shouldn't, given the kind of language FIRRTL is), it warrants some commentary.

mmaloney-sf (commented):
Thank you for writing this up.

I don't know what you might do about my comments. It's better to have this information in the spec than not.

spec.md
mikeurbach: These are introduced in the section Primitive Property Operations.

spec.md
mikeurbach: I think I'd consider `List<T>(...)` the primop.

This is just writing down the spec as-is for a feature that has been supported in the CIRCT FIRRTL compiler for nearly a year. If we want to change it we can change it, but I think it's better to have something that has stabilized in the spec than not.

include/firrtl.xml
```diff
@@ -36,6 +36,7 @@
       <item>Reset</item>
       <item>AsyncReset</item>
       <item>Integer</item>
+      <item>List</item>
```
dtzSiFive, line 39: Since it's parameterized, and looking at my draft List spec commits (apparently never submitted oops), I think this might render better if you put this down in `outertype`.

IIRC making this work with, e.g., `List<List<Integer>>` was a bear and lead to a bunch of fixes since unlike probes these can nest.

spec.md
```diff
@@ -3938,6 +3952,27 @@ The shift right operation result is the arbitrary precision signed integer arith
...
+
+The list construction operation constructs a List property type expression of a given element type.
```
dtzSiFive, line 3965:
```suggestion
The list construction operation constructs a `List`{.firrtl} property type expression of a given element type.
```
for consistency?

spec.md
```diff
@@ -4472,6 +4509,9 @@ primop_1expr2int_keyword = "bits" ;
 
 property_primop_2expr_keyword =
     "integer_add" | "integer_mul" | "integer_shr" ;
+
+property_primop_varexpr_keyword =
+    "List", "<", type_property, ">" | "list_concat" ;
```
dtzSiFive, line 4514:
```suggestion
    "List" , "<" , type_property , ">" | "list_concat" ;
```

spec.md
```diff
@@ -4190,7 +4225,7 @@ decl_type_alias = "type", id, "=", type ;
 
 port = ( "input" | "output" ) , id , ":" , (type | type_property) , [ info ] ;
 type_param = int | string_dq | string_sq ;
-type_property = "Integer" ;
+type_property = "Integer" | "List", "<", type_property, ">";
```
dtzSiFive, line 4228:
```suggestion
type_property = "Integer" | "List" , "<" , type_property , ">" ;
```
(Just advocating for consistency with style surrounding 👍 )

spec.md
```diff
@@ -4399,6 +4434,8 @@ primop_1expr2int = primop_1expr2int_keyword , "(" , expr , "," , int , "," , int
 (* Primitive Property Operations *)
 property_primop_2expr = property_primop_2expr_keyword ,
                           "(" , property_expr , "," , property_expr ")" ;
+property_primop_varexpr = property_primop_varexpr_keyword ,
+                            "(" , property_expr , [ "," , property_expr ]* ")" ;
```
dtzSiFive, line 4438: This should support zero arguments (in spec text above and as implemented), but as written requires at least one property argument.

Also, I think `{`s are used for zero-or-more, not kleene star.

dtzSiFive (approved):
LGTM, left some suggestions.  Thanks for getting this into the spec!

include/firrtl.xml
dtzSiFive: Testing with the above nested List indeed this renders better -- in particular the angle brackets and types are consistently highlighted.

spec.md
mikeurbach: Thanks, I missed the `{` `}` instead of kleene star. I think this should now indicate zero arguments the way we normally specify zero or more.

include/firrtl.xml
mikeurbach: Thanks, yep that looks good.

PR #229, opened 2024-08-12
[minor] Formal Test definitions and test harnesses

dobios:
This PR proposes a new `formal` construct that allows for test harness modules to be marked as formal tests to be run using bounded model checking with a given bound. This is a first step towards enabling inline formal tests and already has support in CIRCT through the btor2 and circt-bmc backends. It also has the following approved PR that adds support for the new op in CIRCT (https://github.com/llvm/circt/pull/7374)

Example:
```firrtl
FIRRTL version 4.0.0

circuit Foo:
  public module Foo:
    input data : UInt<32>
    input c : UInt<1>
    output out : UInt<32>
    ;; Foo body

  module FooTest:
      ;; example test
      inst foo of Foo
      ;; symbolic input -- maps to input in btor2
      input s_foo_c : UInt<1>
      input s_foo_data : UInt<32> 
      ;; feed the symbolic inputs to the instance
      connect foo.c, s_foo_c 
      connect foo.data, s_foo_data
      ;; example assertion that uses the symbolic inputs and outputs
      intrinsic(circt_verif_assert, intrinsic(
        circt_ltl_implication : UInt<1>, s_foo_c, eq(foo.out, s_foo_data))
      )

  formal testFormal of FooTest, bound = 20
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -3442,7 +3442,7 @@ These are for demonstration and their meaning or validity is determined by the i
 
 The following shows an intrinsic expression for the intrinsic named "circt\_ltl\_delay" with two parameters, returns `UInt<1>`{.firrtl}, and has one operand.
 
-``` {.{..firrtl}}
+``` {.{.{..firrtl}}}
 FIRRTL version 4.0.0
 circuit Foo :
   ;; snippetbegin
@@ -3457,7 +3457,7 @@ The following has an intrinsic statement with an intrinsic expression as its ope
 The statement is for the intrinsic named "circt\_verif\_assert".
 The expression is for the intrinsic named "circt\_isX" which returns a `UInt<1>`{.firrtl} and takes an operand.
 
-``` {.{..firrtl}}
+``` {.{.{..firrtl}}}
 FIRRTL version 4.0.0
 circuit Foo :
   ;; snippetbegin
--- a/spec.md
+++ b/spec.md
@@ -3442,7 +3442,7 @@ These are for demonstration and their meaning or validity is determined by the i
 
 The following shows an intrinsic expression for the intrinsic named "circt\_ltl\_delay" with two parameters, returns `UInt<1>`{.firrtl}, and has one operand.
 
-``` {..firrtl}
+``` {.{..firrtl}}
 FIRRTL version 4.0.0
 circuit Foo :
   ;; snippetbegin
@@ -3457,7 +3457,7 @@ The following has an intrinsic statement with an intrinsic expression as its ope
 The statement is for the intrinsic named "circt\_verif\_assert".
 The expression is for the intrinsic named "circt\_isX" which returns a `UInt<1>`{.firrtl} and takes an operand.
 
-``` {..firrtl}
+``` {.{..firrtl}}
 FIRRTL version 4.0.0
 circuit Foo :
   ;; snippetbegin
--- a/spec.md
+++ b/spec.md
@@ -286,6 +286,59 @@ circuit Foo :
   public module Foo enablelayer A :
 ```
 
+## Formal Tests
+
+Formal tests are names top-level constructs which allow for test-harnesses to be definied 
+as bounded model checking problems. Formal tests have 3 operands: 
+- A symbol that the test can be referred to with.  
+- The name of the test harness module.  
+- A bound for the bounded model checking problem, which refers to the number of cycles that 
+the module will be unrolled for in the formal test. This is a strictly positive integer.  
+
+```firrtl
+FIRRTL version 4.0.0
+circuit Foo :
+  public module Foo :
+    ; ...
+  formal testFoo of Foo, bound = 20
+```
+ 
+### Test Harness  
+
+A test harness is a public module that is typically only used for defining a test-bench.
+These are not instanciated and their inputs function as free (or symbolic) variables.
+Test harnesses typically instantiate a module, know as the Device Under Test (DUT), 
+connect a set of free variables to the DUT's inputs and then use those free variables to 
+check various properties of these DUT using assertions and assumptions.  
+
+Example:
+```firrtl
+FIRRTL version 4.0.0
+
+circuit Foo:
+  public module Foo:
+    input data : UInt<32>
+    input c : UInt<1>
+    output out : UInt<32>
+    ;; Foo body
+
+  module FooTest:
+      ;; example test
+      inst foo of Foo
+      ;; symbolic input -- maps to input in btor2
+      input s_foo_c : UInt<1>
+      input s_foo_data : UInt<32> 
+      ;; feed the symbolic inputs to the instance
+      connect foo.c, s_foo_c 
+      connect foo.data, s_foo_data
+      ;; example assertion that uses the symbolic inputs and outputs
+      intrinsic(circt_verif_assert, intrinsic(
+        circt_ltl_implication : UInt<1>, s_foo_c, eq(foo.out, s_foo_data))
+      )
+
+  formal testFormal of FooTest, bound = 20
+```
+
 # Circuit Components
 
 Circuit components are the named parts of a module corresponding to hardware.
@@ -4202,6 +4255,7 @@ decl =
     decl_module
   | decl_extmodule
   | decl_layer
+  | decl_formal
   | decl_type_alias ;
 
 decl_module =
@@ -4220,8 +4274,10 @@ decl_extmodule =
 
 decl_layer =
   "layer" , id , string , ":" , [ info ] , newline , indent ,
-    { decl_layer , newline } ,
   dedent ;
+
+decl_formal =
+  "formal" , id , "of" , id , "," , "bound" , "=" , int;
 
 decl_type_alias = "type", id, "=", type ;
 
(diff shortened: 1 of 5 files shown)


mwachs5 (commented):
from an administrative standpoint, if this is just adding a new feature it can be `minor`

spec.md
```diff
@@ -286,6 +286,59 @@ circuit Foo :
   public module Foo enablelayer A :
 ```
 
+## Formal Tests
+
+Formal tests are names top-level constructs which allow for test-harnesses to be definied
```
mwachs5, line 291:
```suggestion
Formal tests are named top-level constructs which allow for test-harnesses to be definied 
```

spec.md
```diff
@@ -286,6 +286,59 @@ circuit Foo :
   public module Foo enablelayer A :
 ```
 
+## Formal Tests
+
+Formal tests are names top-level constructs which allow for test-harnesses to be definied 
+as bounded model checking problems. Formal tests have 3 operands: 
+- A symbol that the test can be referred to with.  
+- The name of the test harness module.  
+- A bound for the bounded model checking problem, which refers to the number of cycles that
```
mwachs5, line 295: what is a 'cycle'? Is there an assumed clock for a formal test?

spec.md
```diff
@@ -286,6 +286,59 @@ circuit Foo :
...
+circuit Foo :
+  public module Foo :
+    ; ...
+  formal testFoo of Foo, bound = 20
+```
+ 
+### Test Harness  
+
+A test harness is a public module that is typically only used for defining a test-bench.
+These are not instanciated and their inputs function as free (or symbolic) variables.
```
mwachs5, line 309:
```suggestion
These are not instantiated and their inputs function as free (or symbolic) variables.
```

spec.md
```diff
@@ -286,6 +286,59 @@ circuit Foo :
...
+
+```firrtl
+FIRRTL version 4.0.0
+circuit Foo :
+  public module Foo :
+    ; ...
+  formal testFoo of Foo, bound = 20
+```
+ 
+### Test Harness
```
mwachs5, line 306:
```suggestion
### Formal Test Harness  
```

spec.md
dobios: Formal tests are converted into state-transition systems, so I guess 'time-step' is a better term. The clock used for the time step is whatever clocks are tied to the assertions or stateful elements used through the harness, so a bound of 20 defines a state transition system that is 20 nodes deep.

spec.md
```diff
@@ -286,6 +286,59 @@ circuit Foo :
...
+  formal testFoo of Foo, bound = 20
+```
+ 
+### Test Harness  
+
+A test harness is a public module that is typically only used for defining a test-bench.
+These are not instanciated and their inputs function as free (or symbolic) variables.
+Test harnesses typically instantiate a module, know as the Device Under Test (DUT), 
+connect a set of free variables to the DUT's inputs and then use those free variables to 
+check various properties of these DUT using assertions and assumptions.
```
mwachs5, line 312: Can you mention the keyword that is used to indicate the formal testharness? Is it `formal`?

Also can we standardize in this doc on Test harness vs testharness vs test-harness, same for testbench?

dobios: test-harness + `formal` = formal test-bench. I'll standardize that!

spec.md
```diff
@@ -286,6 +286,67 @@ circuit Foo :
...
+  public module Foo :
+    ; ...
+  formal testFoo of Foo, bound = 20
+```
+More details about how bounded model checking works and what the bound refers to can be found in 
+[Biere et al. 2003](https://www.cs.cmu.edu/~emc/papers/Books%20and%20Edited%20Volumes/Bounded%20Model%20Checking.pdf).
+ 
+### Formal Test Harness  
+
+A formal test harness is a public module that is typically only used for defining the
```
uenoku, line 315: I think  requiring test harness to be a public seems reasonable. Does this mean `FooTest` below must have `public` quantifier?

spec.md
```diff
@@ -286,6 +286,67 @@ circuit Foo :
...
 ```
 
+## Formal Tests
+
+Formal tests are named top-level constructs which allow for formal test harnesses to be definied 
+as bounded model checking (BMC) problems. The body of the formal test harness is converted 
+into a BMC formula which encodes a state-transition system version of the design using one 
+of the formal backends (targetting either BTOR2 or SMTLib). 
+Formal tests have 3 operands: 
+- A symbol that the test can be referred to with.
```
mmaloney-sf, line 296: Formatting breaks here. You need an extra newline.

mmaloney-sf, line 296: We should speak of the "name of the test" rather than a "symbol *for* the test".

spec.md
```diff
@@ -286,6 +286,67 @@ circuit Foo :
...
+as bounded model checking (BMC) problems. The body of the formal test harness is converted 
+into a BMC formula which encodes a state-transition system version of the design using one 
+of the formal backends (targetting either BTOR2 or SMTLib). 
+Formal tests have 3 operands: 
+- A symbol that the test can be referred to with.  
+- The name of the formal test harness module.  
+- A bound for the bounded model checking problem, which refers to the maximum bound k used
+to check the bounded model checking formula. This means that given a bound k, the BMC formula 
+will be checked for all unrollings of the resulting state-transition system of depths 1 through k.
+This bound is a strictly positive integer.
```
mmaloney-sf, line 301: This probably needs to be further layicized, since it introduces a number of concepts beyond the scope of the FIRRTL spec.

Could I suggest something like:

> "A positive integer bound for test. This number indicates the depth to which the model will be checked."

spec.md
```diff
@@ -286,6 +286,67 @@ circuit Foo :
...
+
+```firrtl
+FIRRTL version 4.0.0
+circuit Foo :
+  public module Foo :
+    ; ...
+  formal testFoo of Foo, bound = 20
+```
+More details about how bounded model checking works and what the bound refers to can be found in 
+[Biere et al. 2003](https://www.cs.cmu.edu/~emc/papers/Books%20and%20Edited%20Volumes/Bounded%20Model%20Checking.pdf).
```
mmaloney-sf, line 311: I think this is the first citation we have in the spec. Should we put a proper citation section at the bottom?

I guess I feel weird just linking to a university homepage like this. In my experience, they have a tendency to break.

seldridge: It's reasonable to stand up proper bibtex-backed citations. I can see about doing this.

spec.md
mmaloney-sf, line 315: I find the wording a little confusing. It is phrased as if we are defining a new notion ("formal test harness"), but reading on, I'm pretty sure you are just associating a module definition with the `formal` block.

spec.md
```diff
@@ -286,6 +286,67 @@ circuit Foo :
...
+ 
+### Formal Test Harness  
+
+A formal test harness is a public module that is typically only used for defining the 
+body of a formal test-bench. These are modules that are not intended to be instantiated and 
+their inputs function as free (or symbolic) variables. Formal test harnesses typically
```
mmaloney-sf, line 317: With this sentence and the one above, you use the word "typically".

We should be careful when we use this word. It conveys a norm rather than a rule.

spec.md
```diff
@@ -4220,9 +4282,11 @@ decl_extmodule =
 
 decl_layer =
   "layer" , id , string , ":" , [ info ] , newline , indent ,
-    { decl_layer , newline } ,
```
mmaloney-sf, line 4223: Is this intentional? Or did this clip another branch?

spec.md
dobios: I updated it to a doi link, those should never break

spec.md
dobios: I just want to describe what the expected input is. Rather than giving the formal block a region, we instead reuse the notion of a module to just wrap whatever logic the test will contain. In the module given to a formal test, inputs are considered symbolic/free, which allows us to avoid introducing a new construct call `symbolic` that does exactly that. This is just conceptual reuse but I wanted to describe how the body of the module given to a formal op differs from a regular module body.

spec.md
dobios: I just wanted to clarify that this isn't something that was made-up for firrtl and that it's rather a widely used concept already that can be used however you like but is expected to be used in a certain way -- not sure if that makes sense, I can remove the word typically to avoid any confusion.

spec.md
dobios: woops merge artifact probably

spec.md
mmaloney-sf: Sounds good to me.

spec.md
mmaloney-sf: Could I recommend this simpler variation?:

> A public module definition may be used as a formal test harness. Their `input` ports act as symbolic variables for the test.

Again, this sticks to just the mechanics of the feature, and avoids commenting on how it should be used. (If we want to include suggestions like "a test harness should instantiate a DUT" etc, we should split it out into its own paragraph).


By the way, does this mean the `symbolic` statements have been removed?

dobios: Yes @darthscsi pointed out that if these behave like undriven input then they should look like undriven inputs and that I should not add a new op that simply reproduces the behavior of another op but in a different context.

mmaloney-sf (approved):

PR #238, opened 2024-10-03
[minor] Add Property primitive operation for integer shift left.

maerhart:


mikeurbach (approved):
Looks good to me, thanks for adding it!

PR #248, opened 2024-11-13
Fix intrinsic grammar to match CIRCT implementation

fabianschuiki:


fabianschuiki:
I think this is part of "Add intrinsic expressions and statements.". It's basically a fixup for something that hasn't been released yet :slightly_smiling_face:

PR #252, opened 2024-11-14
[major] Adjust syntax of formal unit tests

fabianschuiki:
Adjust the syntax of the `formal` declaration to no longer expect a single fixed `bound = <N>` parameter, but instead accept a list of user-defined parameters similar to an `extmodule`. The parameters can be integers, strings, arrays, and dictionaries. They are passed through to the tooling running the formal test. As we build out that tooling, we may standardize some of the parameters.

Marking this as "major" since it technically is a breaking change, but in practice there are no users of `formal` as of today and the original syntax has never been released as part of the spec.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -286,61 +286,63 @@ circuit Foo :
   public module Foo enablelayer A :
 ```
 
-## Formal Tests
+## Formal Unit Tests
 
-Formal tests are named top-level constructs which allow for formal test harnesses to be definied
-as bounded model checking (BMC) problems. The body of the formal test harness is converted
-into a BMC formula which encodes a state-transition system version of the design using one
-of the formal backends (targetting either BTOR2 or SMTLib).
-Formal tests have 3 operands:
+The `formal`{.firrtl} keyword declares a formal unit test.
+Tools may look for these constructs and automatically run them as part of a regression test suite.
 
-1.  A name that the test can be referred to with.
-2.  The name of the formal test harness module.
-3.  A positive integer bound for the test. This number indicates the depth to which the model will be checked.
+A formal unit test has a unique name and refers to a module declaration that contains the design to be verified and any necessary test harness.
+Additionally, the test may also declare a list of user-defined named parameters which can be integers, strings, arrays, or dictionaries.
+These parameters are passed on to any tools that execute the tests.
+No parameters with well-known semantics are defined yet.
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -286,61 +286,63 @@ circuit Foo :
...
+A formal unit test has a unique name and refers to a module declaration that contains the design to be verified and any necessary test harness.
+Additionally, the test may also declare a list of user-defined named parameters which can be integers, strings, arrays, or dictionaries.
+These parameters are passed on to any tools that execute the tests.
+No parameters with well-known semantics are defined yet.
+During execution of the formal test, all input ports of the target module are treated as symbolic values that may change in every cycle.
+All output ports are ignored.
+
+The circuit below shows a module run as a formal unit test:
 
 ``` firrtl
```
seldridge, line 303: You can use the following to cause the build to _not_ test this.

```suggestion
``` {.firrtl .notest}
```

seldridge: Ditto for other places where you are running ahead of the latest CIRCT release.

spec.md
```diff
@@ -4292,7 +4294,15 @@ decl_layer =
...
-  "formal" , id , "of" , id , "," , "bound" , "=" , int;
+  "formal" , id , "of" , id , ":" , [ info ] , newline , indent ,
+    { id , "=" , decl_formal_param , newline } ,
+  dedent ;
+decl_formal_param =
+    int
+  | string_dq
+  | string_sq
+  | "[" , [ decl_formal_param , { "," , decl_formal_param } ] , "]"
+  | "{" , [ id , "=" , decl_formal_param , { "," , id , "=" , decl_formal_param } ] , "}"
```
seldridge, line 4305: Got it. So, numbers, strings, and lists or dictionaries of the same types. Seems reasonable and flexible.

spec.md
fabianschuiki: Awesome thanks, let me tweak that.

fabianschuiki:
I like the `<...>` syntax for the parameters. I was wondering where to put the test parameters as opposed to the module parameters in this case, though. Something like the following maybe?
```
formal myTest<mode="bmc", bound=42> of Foo<dataWidth=9001, fallthrough=0, otherModuleParam=42>:
```
Might be worth revisiting once we have module parameters?

PR #254, opened 2024-11-15
[nfc] Bump Runners and Actions

seldridge:
Bump GitHub actions and runners to latest available versions.


seldridge:
These changes will not be fully tested until CD and other actions trigger. I will deal with the fallout if they break at those times.

PR #265, opened 2024-12-09
[minor] Add Instance Choice

seldridge:
Add instance choice, a new feature that allows for one of several possible instances to be instantiated at Verilog elaboration time.

### DIFF
--- a/abi.md
+++ b/abi.md
@@ -311,6 +311,136 @@ bind Bar Bar_Layer1_Layer2 layer1_layer2(.notA(Bar.layer1.notA));
 The `` `ifdef ``{.verilog} guards enable any combination of the bind files to be included while still producing legal SystemVerilog.
 I.e., the end user may safely include none, either, or both of the bindings files.
 
+## On Targets
+
+A target will result in the creation of specialization files.
+One specialization file per public module per option for a given target will be created.
+
+Including a specialization file in the elaboration of Verilog produced by a FIRRTL compiler specializes the instantiation hierarchy under the associated public module with that option.
+
+Each specialization file must have the following filename where `module` is the name of the public module, `target` is the name of the target, and `option` is the name of the option:
+
+``` ebnf
+filename = "targets_" , module , "_" , target , "_", option , ".vh" ;
+```
+
+Each specialization file will guard for multiple inclusion.
+Namely, if two files that specialize the same target are included, this must produce an error.
+If the file is included after a module which relies on the specialization, this must produce an error.
+
+### Example
+
+The following circuit is an example implementation of how a target can be lowered to align with the ABI defined above.
+This circuit has one target named `Target` with two options, `A` and `B`.
+Module `Foo` may be specialized using an instance choice that will instantiate `Bar` by default and `Baz` if `Target` is set to `A`.
+If `Target` is set to `B`, then the default instantiation, `Bar`, will occur.
+Module `Foo` includes a probe whose define is inside the instance choice.
+
+``` firrtl
+FIRRTL version 4.1.0
+circuit Foo:
+
+  option Target:
+    case A
+    case B
+
+  module Baz:
+    output a: Probe<UInt<1>>
+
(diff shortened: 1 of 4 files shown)


abi.md
```diff
@@ -311,6 +311,136 @@ bind Bar Bar_Layer1_Layer2 layer1_layer2(.notA(Bar.layer1.notA));
...
+endmodule
+
+module Bar(output a);
+  assign a = 1'h0;
+endmodule
+
+// Defines for the instance choices
+`ifndef __target_Target_foo_x
+ `define __target_Target_foo_x Bar
+`endif
```
uenoku: comment for link.

abi.md
```diff
@@ -311,6 +311,136 @@ bind Bar Bar_Layer1_Layer2 layer1_layer2(.notA(Bar.layer1.notA));
...
+
+module Foo();
+
+  `__target_Target_foo_x x();
```
uenoku: Do we need this macro replacement for module name (which requires a additional very specialized operation to HW)? Can we do that with a plain if-def chains?

abi.md
```diff
@@ -311,6 +311,136 @@ bind Bar Bar_Layer1_Layer2 layer1_layer2(.notA(Bar.layer1.notA));
...
+`ifdef __target_Target_foo_x
+ `ERROR__target_Target_foo_x__must__not__be__set
+`else
+ `define __target_Target_foo_x Baz
+`endif
+
+`ifdef __targetref_Foo_x_a a
+ `ERROR__targetref_Foo_x_a__must__not__be__set
+`else
+ `define __targetref_Foo_x_a a
```
uenoku: Note that we cannot do emit `__targetref` in a header if we want to support nested instance choices. 

```scala

FIRRTL version 5.1.0
circuit Top :
  option Platform :
    FPGA
  option Optimization : 
    Fast
  

  module DefaultTarget : 
    output probe : Probe<UInt<8>> 

    wire r : UInt<8> 
    connect r, UInt<8>(0) 
    define probe = probe(r) 

  module FastTarget : 
    output probe : Probe<UInt<8>> 
    wire w : UInt<8> 
    connect w, UInt<8>(1) 
    define probe = probe(w) 
   
  module FPGATarget : 
    output probe : Probe<UInt<8>> 

    instchoice opt_inst of DefaultTarget, Optimization:
      Fast => FastTarget
    define probe = opt_inst.probe


  public module Top : 
    output probe: Probe<UInt<8>> 
    instchoice platform_inst of DefaultTarget, Platform:
      FPGA => FPGATarget
    define probe = platform_inst.probe

```

```verilog
// in targets_Platform_FPGA.svh
`define __targetref_Top_platform_inst_probe `__target_Platform_Top_platform_inst.`__targetref_FPGATarget_opt_inst_probe
```
`__targetref_FPGATarget_opt_inst_probe` is defined in `targets_Optimization_Fast.svh`, and cyclic dependency could be easily introduced if there is another nested instance choice op that went through `Optimization` -> `Platform`.

seldridge: Nice find! Can we break the file inclusion problem by requiring the ordering of the files to be:

1. Individual instance choice headers which only set instances
2. and a unified ref header which sets all refs based on instance choices.

Since we've backtracked on the need to hide all instance choices, and instead can handle that with a dedicated tool, this might work. It does mean that the ref header is going to be complex and likely have to mirror the instantiation hierarchy and the options along the way.

Alternatively, (1) could include partial ref paths which are assembled into the final paths in (2).

uenoku: > Individual instance choice headers which only set instances
and a unified ref header which sets all refs based on instance choices.

That is close to what I've implemented so far (= simply inlining paths in `ref_*_*.sv` files with if-def guards https://gist.github.com/uenoku/d47bd68320dc2cd424ff42df27ed91a4). 

The issue with this approach it's necessary to hide specific options/probes from ref files, but yes, it's possible to do with post-process.

PR #289, opened 2025-03-25
[minor] Add simulation unit test construct

fabianschuiki:
Add a `simulation` unit test declaration analogous to the existing `formal` construct. This allows simulation test harness modules to be marked as entry points for simulation, which downstream tools can detect and run. This is more or less a copy of `formal`, with a few added constraints on the target module's ports and the exact simulation schedule.


spec.md
```diff
@@ -350,6 +350,95 @@ circuit Foo :
...
 
+## Simulation Unit Tests
+
+The `simulation`{.firrtl} keyword declares a simulation unit test.
+Tools may look for these constructs and automatically run them as part of a regression test suite.
+
+A simulation unit test has a unique name and refers to a module declaration that contains the design to be verified and any necessary test harness.
+Additionally, the test may also declare a list of user-defined named parameters which can be integers, strings, arrays, or dictionaries.
+These parameters are passed on to any tools that execute the tests.
+No parameters with well-known semantics are defined yet.
```
seldridge, line 361: I'm a little worried about how underspecified this part is.

There are a couple of ways to take it: 

1. Leave it as is and basically say that the simulation can have an arbitrary key-value mapping. I'd be hesitant to include anything beyond integers and strings first. (Also, if strings, what is the syntax? Double quoting strings like everything else?)
2. Try to define this more strongly with an eye towards this being the eventual FIRRTL parameterization system.
3. Go in a janky direction and say that there is a single string where you can put arguments and this is all implementation defined. (This is basically the same as (1), but is trying to go harder in the direction that this anything is allowed.)

I think I'm okay with (1), as it is. If you want dictionaries and lists, then that needs to be defined. Strings and ints only needs a little more text, but not much.

seldridge (commented):
Very, very excited to see this!

spec.md
```diff
@@ -350,6 +350,95 @@ circuit Foo :
...
+  Time Step      Clock       Init
+  -------------- ----------- -----------
+  1 (optional)   undefined   undefined
+  2              0           1
+  3              1           1
+  4 (optional)   1           1 or 0
+  5 (optional)   0           1 or 0
+  6              0           0
+  7              1           0
+  8              0           0
```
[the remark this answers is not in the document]
fabianschuiki: The optional stuff ensures that the simulator has a tiny bit of wiggle room in how it generates the waves. For example, the optional (1) allows for these two to be valid:

wavedrom

The optional (4) and (5) allow the init signal to drop anytime between the first and second rising edge:

wavedrom (2)

I think I can collapse this to the following, for simplicity:

wavedrom (3)

spec.md
fabianschuiki: I think we want this to be (1). This is what's already in the BNF syntax further down in the document, and what's implemented on the CIRCT side. I can add a few lines on how this works for the `formal` and `simulation` parts.

fabianschuiki: I've added a separate "Test Parameters" section that elaborates for both `formal` and `simulation` what the syntax of the parameters is.

spec.md
```diff
@@ -350,6 +375,95 @@ circuit Foo :
...
+    collectCoverage = 1
+```
+
+### Ports
+
+The target module must have exactly the following four ports:
+
+  Name      Type        Direction
+  --------- ----------- -----------
+  clock     Clock       input
```
seldridge, line 426: Nit: FIRRTL syntax would be good here:

```suggestion
  clock     `Clock`.{firrtl}       input
```

Ditto for the other ports.

spec.md
```diff
@@ -4305,19 +4420,25 @@ decl_extmodule =
 
 decl_layer =
   "layer" , id , string , ":" , [ info ] , newline , indent ,
-  { decl_layer , newline } ,
+    { decl_layer , newline } ,
```
seldridge, line 4423: Thanks for this. Always feel free to land these changes directly.

seldridge (approved):
LGTM

PR #293, opened 2025-04-03
[minor] Introduce fprintf

uenoku:
This PR proposes `fprtinf` operation that output strings to a file instead of default file descriptor. https://github.com/llvm/circt/pull/8346 is PR on CIRCT


seldridge (approved):
LGTM! Please add an entry in `revision-history.yaml`.

PR #318, opened 2025-08-12
Prefer tee -a over >> so format diff is discoverable.

dtzSiFive:
This has burned folks multiple times.

Posting it more prominently would be better, but at least this way the diff isn't buried in an unreachable-from-PR location, and will be visible when clicking through the failed job.

This helps occasional and frequent contributors alike.


seldridge:
Nice improvement.

It is probably even more actionable to print a message which says, "run `make format`".

seldridge:
> It is probably even more actionable to print a message which says, "run make format".

Err... it already says this. 🙃 

The only change I can see is that it is not useful to show the ` ``` diff ` in the console output, but is necessary for the GitHub Summary.

dtzSiFive:
> It is probably even more actionable to print a message which says, "run `make format`".

`make format` is unusable in most environments, it relies on very specific versions that are unlikely to be what folks have on hand (and specific latex dependencies).

I have a nix flake (probably out-of-date with what CI uses -- whatever ubuntu has on that image?) to grab these, not sure what the suggested story is for random contributors?

Just running `git diff` to the terminal would probably be better (or digging up something better than the step summary which if you search seems known to be problematic in the way we're finding it).

.github/workflows/continuous-integration-ci.yml
```diff
@@ -38,10 +38,10 @@ jobs:
         run: |
           make format
           if ! git diff --quiet; then
-            echo 'Please run `make format` on your change to apply the following formatting fixes:' >> $GITHUB_STEP_SUMMARY
-            echo '```diff' >> $GITHUB_STEP_SUMMARY
-            git diff >> $GITHUB_STEP_SUMMARY
-            echo '```' >> $GITHUB_STEP_SUMMARY
+            echo 'Please run `make format` on your change to apply the following formatting fixes:' | tee -a $GITHUB_STEP_SUMMARY
+            echo '```diff' | tee -a $GITHUB_STEP_SUMMARY
```
seldridge, line 42: This doesn't need to be `tee`'d. Ditto for line 44.

seldridge (approved):

.github/workflows/continuous-integration-ci.yml
dtzSiFive: Ack, thanks.  Updated.

seldridge:
Isn't this a problem with basically any formatting, though? We have `clang-format` setup on CIRCT and that needs a specific version choice. Note: there is no LaTeX dependency for `make format`, only pandoc.

What I do think is making things worse is that the version of pandoc in this repo is somewhat far behind what most distributions are providing. I will bump this...

.github/workflows/continuous-integration-ci.yml
dtzSiFive: Grumble if GH doesn't have a place to do this but in a way that PR can see it.  D'oh.

dtzSiFive:
> Isn't this a problem with basically any formatting, though? We have `clang-format` setup on CIRCT and that needs a specific version choice. Note: there is no LaTeX dependency for `make format`, only pandoc.

(I believe you re:not needing latex, that seems like how it should work and I didn't check was just reporting what it takes to build a pdf similar to what CI needs, good correction thanks!)

clang-format is comparatively stable across versions, and is widely available in multiple versions.  But absolutely, this should be specified and pinned in a project's CI to avoid this sort of mismatch.  I keep use a clang-format pinned (and have my lsp's formatting use) a specific version over on CIRCT for this reason (preemptively, good habit to get into).

Pandoc changes quite a lot (seems to?) and is packaged inconsistently so it's not easy to get a user up to speed that way -- at least offhand.  Maybe there's a way?

Having a format check is nice, so abstracting the dependencies with nix or something might be the best we can do here while relying on pandoc for this?

dtzSiFive:
Posting the patch would help infrequent contributors (apply posted patch), but not as convenient for development.

.github/workflows/continuous-integration-ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           if ! git diff --quiet; then
             echo 'Please run `make format` on your change to apply the following formatting fixes:' >> $GITHUB_STEP_SUMMARY
```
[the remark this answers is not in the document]
dtzSiFive: Oh okay I'll restore this to `tee -a` and push a test breakage.

seldridge:
> Pandoc changes quite a lot (seems to?) and is packaged inconsistently so it's not easy to get a user up to speed that way -- at least offhand. Maybe there's a way?
> 
> Having a format check is nice, so abstracting the dependencies with nix or something might be the best we can do here while relying on pandoc for this?

It's annoyingly not stable around formatting across versions. There may be a way to lock this down with more options to the formatting.

I bumped to latest here: https://github.com/chipsalliance/firrtl-spec/pull/319 This is at least what MacPorts is providing. Also, there are binary releases for all operating systems for pandoc and pandoc-crossref:
- https://github.com/jgm/pandoc/releases/tag/3.7.0.2
- https://github.com/lierdakil/pandoc-crossref/releases/tag/v0.3.20

PR #348, opened 2026-03-20
[minor] Add property classes and objects

seldridge:
Add classes, objects (instantiations of classes), and the `Inst` type
which is used to pass around references to objects.  This is the last
major piece of "Objective" FIRRTL (OFIRRTL) which adds support for
instantiable metadata.

The ABI for OFIRRTL is to-be-defined, though it has a semi-standard ABI
due to how `firtool` compiles this.

AI-assisted-by: Augment (Sonnet 4.5)

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -226,6 +226,84 @@ Foo #(
 endmodule
 ```
 
+## Classes
+
+Classes are object-oriented constructs that encapsulate property-typed data and can be instantiated to create objects.
+Unlike modules, which represent hardware, classes represent metadata and design intent that can be used for verification, debugging, or other non-hardware purposes.
+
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -1058,6 +1186,28 @@ circuit Example:
   ;; snippetend
 ```
 
+### Inst Type
+
+The `Inst` type represents a class instance property.
+It is parameterized by the name of a class, and represents an instance (object) of that class.
```
mikeurbach: the `instance (object)` seems a little LLM-esque. should we just say "represents an object of that class"?

mikeurbach (approved):
Looks good to me, just some minor nitpicks. Thanks for sending this.

PR #352, opened 2026-04-01
Disallow truncation in connects

darthscsi:
Automatic truncations are unsafe and have caused latent bugs in real usage.  Just eliminate them.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1535,8 +1535,9 @@ The flow of all other expressions are source, including mux and cast.
 
 The type equivalence relation is used to determine whether a connection between two components is legal.
 
-An unsigned integer type is always equivalent to another unsigned integer type regardless of bit width, and is not equivalent to any other type.
-Similarly, a signed integer type is always equivalent to another signed integer type regardless of bit width, and is not equivalent to any other type.
+An unsigned integer type is equivalent to another unsigned integer type as long as the bit width is the same or lower, and is not equivalent to any other type.
+Similarly, a signed integer type is always equivalent to another signed integer type as long as the bit width is the same or lower, and is not equivalent to any other type.
+A connection cannot cause a truncation.
 
 Clock types are equivalent to clock types, and are not equivalent to any other type.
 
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -1535,8 +1535,9 @@ The flow of all other expressions are source, including mux and cast.
 
 The type equivalence relation is used to determine whether a connection between two components is legal.
 
-An unsigned integer type is always equivalent to another unsigned integer type regardless of bit width, and is not equivalent to any other type.
-Similarly, a signed integer type is always equivalent to another signed integer type regardless of bit width, and is not equivalent to any other type.
+An unsigned integer type is equivalent to another unsigned integer type as long as the bit width is the same or lower, and is not equivalent to any other type.
+Similarly, a signed integer type is always equivalent to another signed integer type as long as the bit width is the same or lower, and is not equivalent to any other type.
+A connection cannot cause a truncation.
```
seldridge, line 1540: This makes type equivalence _not_ commutative:

```
type_eq(UInt<1>, UInt<2>) == false
type_eq(UInt<2>, UInt<1>) == true
type_eq(UInt<1>, UInt<1>) == true
```

Consider explicitly mentioning this.

I did a quick search through the spec for other uses of "equivalent".  The only other use is for the description of a `mux` where:

> The types of the two input expressions are equivalent.

This needs the original type equivalence definition to allow for the following to compile:

```
FIRRTL version 7.0.0
circuit Foo:
  public module Foo:
    input sel: UInt<1>
    input a: UInt<1>
    input b: UInt<2>
    output out: UInt<2>

    connect out, mux(sel, a, b)
    connect out, mux(sel, b, a)
    connect out, mux(sel, a, a)
```

I'm not sure how to best handle this and I don't like the divergence.  Two ideas:

1. Keep the original type equivalence definition, but add an additional qualification in the connect algorithm:

```
1.   The types of the left-hand side and right-hand side must by type equivalent (see [@sec:type-equivalence]).

2.   If the left-hand side and right-hand side are unsigned integer or signed integer types and both have explicit bit widths, then the width of the right-hand side must not be larger than the width of the left-hand side.
```

2. Keep this definition here, but add a qualification for the `mux`.

